PR #96, opened 2019-11-12
Migrate the project to go modules

krnowak:
This PR migrates the project to the official go module system. It is now not required anymore to put the project in GOPATH.


krnowak:
Hmph, it looks like I updated the wrong CI file. Should the `.travis.yml` file be dropped then?

joaquimrocha:
I will take a look tomorrow.

Makefile
```diff
@@ -1,3 +1,6 @@
+GO111MODULE=on
+export GO111MODULE
```
joaquimrocha: nitpick: I assume this would work if we had it only in one line (`export GO11MODULE=on`).

.github/workflows/backend.yml
```diff
@@ -29,22 +29,12 @@ jobs:
...
       run: |
-        cd $RUNNER_WORKSPACE
-        NEBRASKA_REPO=$GITHUB_WORKSPACE
-        export GOPATH=${RUNNER_WORKSPACE}/go
-        NEBRASKA_GO=$GOPATH/src/github.com/$GITHUB_REPOSITORY
-        mkdir -p $GOPATH/bin
-        curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-        cd $NEBRASKA_GO
-        $GOPATH/bin/dep ensure
+        cd $GITHUB_WORKSPACE
```
joaquimrocha: :+1: :+1: :+1:

PR #100, opened 2019-11-14
Pin the version of bindata used for generating files

krnowak:
I don't know what version of go-bindata was used to generate the files, so I pinned it to the latest commit.


go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/rubenv/sql-migrate v0.0.0-20180217203553-081fe17d19ff
 	github.com/satori/go.uuid v1.2.0
 	github.com/stretchr/testify v1.4.0
+	github.com/twitter/go-bindata v0.0.0-20180208183117-2fa2cba09795 // indirect
```
joaquimrocha: I don't think this version is being updated anymore.
This one looks like the real stuff:
https://github.com/go-bindata/go-bindata
In which case we should use the latest tag: https://github.com/go-bindata/go-bindata/tree/v3.1.2

krnowak: I looked at it more and realized that this PR was broken. I built the generator as `bindata`, while go generate was calling `go-bindata` that accidentally was somewhere in `PATH`. Heh.

Anyway, there is also https://github.com/kevinburke/go-bindata for example. It's also maintained, I think. I used the version of go-bindata as discussed here: https://github.com/kinvolk/nebraska/blob/168be841ac3e49d73610319774dfff86fc8c2994/pkg/api/api.go#L16-L22

Whether we want to switch to a different implementation of go-bindata or a different solution of generating files, I think it should be a part of a different PR. Here, I'm pinning it for consistency.

joaquimrocha: I don't mind whether we use the one I suggested or this one. It's just important to not keep one that hasn't been updated for over one year. It's of course okay to have another PR for that, but since we're pinning a version that's outdated, I thought it would belong here too.

krnowak: Alright, just updated the PR to use github.com/kevinburke/go-bindata.

joaquimrocha: Thanks. I will take a look.

PR #101, opened 2019-11-15
backend: Fix the tests, update the building process and run tests in CI

krnowak:
When working on the package updates/replacing, I noticed two things:

- tests fail
- CI does not run tests

Tests failed for me when I tried to run them against the docker image built from `Dockerfile.postgres`. First reason was that postgres config caused the postgres process to reject the connection. Second reason was that some migration SQL added some data to some tables, so when sample data in tests was added to the database, there were public key constraint violations.

I didn't want to change `Dockerfile.postgres`, so I added another one, for testing. It's a bit simpler than the original.

When tests succeeded on my computer, I wanted to run them in CI. In the meantime I updated the way we build and lint the project. We used a similar way of doing it in opentelemetry-go project, so that is where I got the inspiration from.

So when building and testing the backend, running `make backend` and `make check-with-container` might be a way of doing it.


Makefile
```diff
@@ -34,6 +57,16 @@ backend: tools/go-bindata tools/golangci-lint
 	go mod tidy
 	go build -o bin/nebraska ./cmd/nebraska
 
+.PHONY: test-clean-work-tree
+test-clean-work-tree:
+	@if ! git diff --quiet; then \
```
joaquimrocha: You could specify here the paths you want to check for the difference (i.e. use only the backend paths / excluding the tools). This way you wouldn't have to care about what's in the `tools` folder and thus not need to have a `.gitignore` there (which is a bit annoying when changing stuff in those folders intentionally).

Makefile
```diff
@@ -34,6 +42,9 @@ tools:
 tools/go-bindata: go.mod go.sum
 	go build -o tools/go-bindata github.com/kevinburke/go-bindata/go-bindata
 
+tools/golangci-lint: go.mod go.sum
+	go build -o ./tools/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
```
joaquimrocha: I don't understand why we're compiling and shipping these external tools. I consider these as dependencies and thus would expect these to be installed in any environment/script that's testing Nebraska.

.gitignore
```diff
@@ -6,3 +6,6 @@ node_modules
 frontend/built/js/main.js
 frontend/built/css/styles.css
 frontend/yarn-error.log
+
+/tools/go-bindata
+/tools/golangci-lint
```
joaquimrocha: See my comment in the file that tests for the tree cleanliness as we shouldn't need this (or even building the tools AFAIU).

docker/wait.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
joaquimrocha: I'd also prefer a more detailed name for this file. Something like `wait_for_postgres.sh`

docker/postgres-test-docker-entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+
+    su-exec postgres pg_ctl -w start
+
+    su-exec postgres psql -c "CREATE DATABASE nebraska_tests;"
+    su-exec postgres psql -c "ALTER DATABASE nebraska_tests SET TIMEZONE = 'UTC';"
+
+    su-exec postgres pg_ctl -m fast -w stop
+fi
+
+touch "${NEBTMP}/done"
```
joaquimrocha: Maybe use a more scoped file name here? `postgres_init_done`?

Makefile
krnowak: We compile the tools, because we are going to use them during the build process. We are not shipping them. Unless I misunderstood what do you mean by "shipping".

I'll discuss the use of the tools in other comment.

docker/wait.sh
krnowak: Right, will do.

docker/postgres-test-docker-entrypoint.sh
krnowak: Right, sounds better.

krnowak:
> Thanks for this great PR. It'll improve the quality a lot.
> I left a few comments in the patches. I also wonder if it's a good thing to always lint on `make backend`, which may slow the process a bit, or have just a githook for it (and the CI checking any PR ultimately in case the hooks are not used). The hook could be added automatically on `make [backend]`. But this means that only commits would be a bit slowed by the lint.
> 
> If the lint happens faster than I remember, then it's fine to keep it as is. We can also add this PR and change it to use a git commit hook later if it makes sense.

I've changed Makefile to time all the commands that are run during `make backend`, and I have made some changes in `pkg/{api,omaha,syncer}` and in `cmd/nebraska`. Got times like:

```
# go generate ./...
real 0.11
user 0.14
sys 0.06

# go build ./...
real 1.90
user 5.32
sys 0.52

# go test ./... with skipping
real 1.61
user 4.74
sys 0.60

# lint
real 1.27
user 3.38
sys 0.27

# go mod tidy
real 0.14
user 0.26
sys 0.10

# building nebraska binary
real 0.96
user 1.46
sys 0.23
```

Running `make backend` immediately again I got:

```
# go generate ./...
real 0.10
user 0.13
sys 0.06

# go build ./...
real 1.45
user 4.36
sys 0.51

# go test ./... with skipping
real 1.16
user 3.12
sys 0.49

# lint
real 0.44
user 0.95
sys 0.20

# go mod tidy
real 0.13
user 0.26
sys 0.08

# building nebraska binary
real 0.20
user 0.47
sys 0.11
```

Go build takes 1.5 to 2 seconds, go test something like that too, go linter 0.5 to 1.5 seconds. So in total something between 3.5 seconds to 6.5 seconds for all the steps. I think I can stand it. ;)

Another reason for running a linter is that I wouldn't want to wait for CI to point out some lint mistakes, nor would I want to remember a second command to run the linter. Also, we are running it with the `--fix` option, so the linter is able to fix some easy issues for us (like formatting).

Git hooks are an option too, but not sure if it's so pressing right now.

You haven't raised that point, but I'll add it - we also run go-bindata everytime just to make sure that the generated file is in sync with the sql files. And we can easily see it also without waiting for CI to point it out.

Makefile
krnowak: Limiting the check to the backend parts is a good idea, I'll try to cook something up (that's basically `git diff --quiet -- go.mod go.sum tools/tools.go pkg cmd updaters`, but I'll check).

But I'd keep the entries in the .gitignore, because those tools will appear there when we run `make backend` and a) I expect them to be there and b) I don't think they are important enough for `git status` to bother me about them.

krnowak:
Updated. Changed some stuff in Makefile to make sure we clean up docker container when the tests fail.

krnowak:
Ugh, ubuntu and its stupid dash shell. Updated the PR to just use bash.

joaquimrocha:
Approving and merging the PR. Hopefully we can revisit the thing about not just installing (not compiling) the external tools in another PR.

PR #103, opened 2019-11-18
Update some packages

krnowak:
The updated packages are mostly the "quick wins" ones. What's left is to update is web framework, database ORM and logging. The last two are kinda coupled since the ORM we use currently is using the current logger used in nebraska.

### DIFF
--- a/pkg/api/packages.go
+++ b/pkg/api/packages.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"time"
 
-	"gopkg.in/fatih/set.v0"
 	"gopkg.in/mgutz/dat.v1"
 	runner "gopkg.in/mgutz/dat.v1/sqlx-runner"
 )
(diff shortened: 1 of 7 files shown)


pkg/api/packages.go
```diff
@@ -240,18 +239,22 @@ func (api *API) updatePackageBlacklistedChannels(tx *runner.Tx, pkg *Package) er
 		return err
 	}
 
-	newChannelsBlacklist := set.NewNonTS()
+	e := struct{}{}
```
joaquimrocha: nitpick: I hope this is standard procedure in Go. As a non-expert in Go, I don't love the non-constant/single letter var just to assign as an empty value to use map as a set.
If the name/non-constness is biz as usual, carry on :smile_cat:

pkg/omaha/omaha.go
```diff
@@ -6,7 +6,7 @@ import (
 	"io"
 	"strconv"
 
-	omahaSpec "github.com/aquam8/go-omaha/omaha"
+	omahaSpec "github.com/coreos/go-omaha/omaha"
```
joaquimrocha: I am very glad to seeing this change. Thanks!

pkg/omaha/omaha.go
```diff
@@ -67,72 +67,64 @@ func (h *Handler) Handle(rawReq io.Reader, respWriter io.Writer, ip string) erro
...
 
-	return h.crAPI.RegisterEvent(machineID, appID, group, eventType, eventResult, event.PreviousVersion, event.ErrorCode)
+func (h *Handler) getStatusMessage(crErr error) omahaSpec.AppStatus {
+	return omahaSpec.AppStatus(h.getStatusMessageStr(crErr))
 }
 
-func (h *Handler) getStatusMessage(crErr error) string {
+// TODO(krnowak): This seems to return a bunch of custom errors. Not
+// sure if we should try to match it to the standard or extra
+// AppStatus constants.
```
joaquimrocha: Yeah, I'd try to match as much as possible to the standard ones.

pkg/api/packages.go
krnowak: First of all, there isn't anything like **standard** procedure or whatever in Go. There are **idiomatic** things. **Idiomatic**. I think I hate this word, by the way. ;)

So anyway, not sure if it is a standard thing. Probably not. Wanted to save some typing, but that's a silly reason. Will fix it.

pkg/omaha/omaha.go
krnowak: Thing is that there isn't many of the standard constants:

https://github.com/coreos/go-omaha/blob/master/omaha/codes.go#L147-L159

So, for now I'll leave it as is and file an issue about it. I will need to check the omaha specification to see what app statuses they specify.

PR #109, opened 2019-12-05
backend: Migrate to a maintained uuid package

krnowak:
github.com/satori/go.uuid haven't seen code changes for over a year
now. This normally wouldn't be a problem if not for the fact that the
library apparently has some issues with generation of the random
UUIDs. Also the top issue in the repository has a title "PSA: This
repo is dead".

Alternatives were github.com/gofrs/uuid or github.com/google/uuid. The
former seems to be a continuation of the satori's repo. Picked the
latter, since I wasn't sure about the longevity of the former.

We still require the repository indirectly, though. When we replace
the ORM we have now with something else, then the old UUID package
will likely go away.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/gobuffalo/packr v1.30.1 // indirect
 	github.com/golangci/golangci-lint v1.21.0
 	github.com/google/go-github/v28 v28.1.1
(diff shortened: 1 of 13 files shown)


krnowak:
Updated to fix conflicts.

PR #115, opened 2019-12-20
Add architecture information to the UI

joaquimrocha:
This PR adds the new architecture information of packages and channels to the UI.

Screenshot from 2019-12-20 22-15-50
Screenshot from 2019-12-20 22-15-29
Screenshot from 2019-12-20 22-15-21
Screenshot from 2019-12-20 22-03-41

### DIFF
--- a/frontend/src/js/constants/helpers.js
+++ b/frontend/src/js/constants/helpers.js
@@ -5,6 +5,15 @@ import orange from '@material-ui/core/colors/orange';
 import red from '@material-ui/core/colors/red';
 import moment from 'moment';
 
+// Indexes/keys for the architectures need to match the ones in
+// pkg/api/arches.go.
+export const ARCHES = {
+  0: 'ALL',
+  1: 'AMD64',
+  2: 'ARM64',
+  3: 'X86',
+};
+
 const colors = makeColors();
 
 function makeColors() {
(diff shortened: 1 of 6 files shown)


frontend/src/js/constants/helpers.js
```diff
@@ -5,6 +5,13 @@ import orange from '@material-ui/core/colors/orange';
 import red from '@material-ui/core/colors/red';
 import moment from 'moment';
 
+export const ARCHES = {
```
krnowak: Maybe a comment that it matches (or needs to match) values in `pkg/api/arches.go`?

krnowak:
Maybe in a future iteration we could add some colored labels for arches, so it will be easier to quickly distinguish between ARM64 and AMD64 channels/packages.

joaquimrocha:
> Maybe in a future iteration we could add some colored labels for arches, so it will be easier to quickly distinguish between ARM64 and AMD64 channels/packages.

Sure. That section needs a UX review anyway.

frontend/src/js/constants/helpers.js
joaquimrocha: I will do that.

joaquimrocha:
Merging after having added the comment @krnowak suggested.

PR #116, opened 2020-01-03
Refactor authentication a bit and add a noop authenticator 

krnowak:
This PR introduces some breaking changes - it renames the flags that were used by github authentication (e.g. `-client-id` is now `-gh-client-id`). Same with environment variables (e.g. `NEBRASKA_OAUTH_CLIENT_ID` is now `NEBRASKA_GITHUB_OAUTH_CLIENT_ID`).

If the breaking changes are undesired, then my plan would be to use "github.com/spf13/pflag" instead of the "flag" package from standard library. Then add the old flags back and mark them as deprecated (the standard "flag" package doesn't have the deprecation functionality).

This PR basically moves the github authentication code into a separate object (github authenticator), refactors some code in nebraska, adds a noop authenticator (both are implementations of a single interface) and tells nebraska to instatiate either authenticator depending on a value of the `-auth-mode` flag.

There are some differences between old github authentication and new one - mostly is about locking some maps as they could be read/written in parallel. And did some gin-specific fixes.

I updated the authorization docs too, to reflect the changes in this PR.


docs/authorization.md
```diff
@@ -1,12 +1,27 @@
...
+
+- Start the database: `docker run --privileged -d -p
+  127.0.0.1:5432:5432 nebraska/postgres`
+
+- Start the Nebraska backend:
+
+  - `LOGXI=* nebraska -auth-mode noop -http-static-dir $PWD/frontend/built
+    -http-log`
+
+- In the browser, access `http://localhost:8000`
```
joaquimrocha: :+1:

docs/authorization.md
```diff
@@ -1,12 +1,27 @@
...
-users through the GitHub Oauth flow and create a persistent session.
+Nebraska uses either a noop authentication, or OAuth or Bearer tokens
+to authenticate and authorize users. Currently, only GitHub is
+supported as an OAuth authentication backend.  [GitHub personal access
+tokens](https://github.com/settings/tokens) can be used as bearer
+token. If no bearer token is sent, Nebraska will authorize users
+through the GitHub Oauth flow and create a persistent session.
 
-# Deploying Nebraska for testing on local computer
+# Deploying Nebraska for testing on local computer (noop authentication)
```
joaquimrocha: Nitpick: I'd use `...for testing/development` .

krnowak: Done.

cmd/nebraska/auth/auth.go
```diff
@@ -0,0 +1,30 @@
...
+	log "github.com/mgutz/logxi/v1"
+
+	"github.com/kinvolk/nebraska/cmd/nebraska/ginhelpers"
+)
+
+var (
+	logger = log.New("auth")
+)
+
+// Authenticator interface provides a way to authenticate a user
```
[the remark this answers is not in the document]
krnowak: Fixed.

krnowak:
> looks good on a cursory look. I didn't test it :)

Don't worry, I did. :)

PR #119, opened 2020-01-21
backend: Fix nil pointer dereference in syncer

krnowak:
It maybe possible that the app response will not contain the update
check element, so after decoding the update check field in the struct
may be nil. If that happens, treat it as no update available.


krnowak:
Updated the PR to also add an update check response in case of some internal failure.

Issue #122, opened 2020-01-23
Fix instances backend pagination


joaquimrocha:
The API method that fetches the instances from the DB supports pagination but it's incompatible with the needed use-case for two main reasons:

1. There is no way of getting the total count of the instances when they have been filtered (by version and/or status). So we cannot display e.g. "200/1000" in the instance view unless we retrieve all the instances before (which defeats the purpose of pagination).

2. The pagination is happening [before](https://github.com/kinvolk/nebraska/blob/65c6f69db6e6df87bb2cc848439e489893726c17/pkg/api/instances.go#L269) the filters are applied, which means that if we happen to request a page that has no instances matching the filters, nothing is returned.

In order to fix this and make it useful, we should return the maximum number of instances + the number of filtered instances + the actual filtered and paginated instances in the result.

ashu8912:
@joaquimrocha this is fixed and can be closed now

joaquimrocha:
Thanks @ashu8912 .

PR #124, opened 2020-01-23
Show correct number of instances in instances list view

joaquimrocha:
Due to #122 the instances list view was displaying just up to 500 instances.
This patch fixes that until we have the backend pagination working correctly.

### DIFF
--- a/frontend/src/js/components/Instances/List.js
+++ b/frontend/src/js/components/Instances/List.js
@@ -169,7 +169,10 @@ function ListView(props) {
     const now = moment();
     if (now.diff(lastCheck, 'seconds', true) > CHECKS_TIMEOUT) {
       setLastCheck(now);
-      instancesStore.getInstances(application.id, group.id, null);
+      instancesStore.getInstances(application.id, group.id, null,
+        {
+          perpage: group.instances_stats.total
+        });
     }
 
     return function cleanup() {
--- a/frontend/src/js/components/Instances/List.js
+++ b/frontend/src/js/components/Instances/List.js
@@ -82,7 +82,7 @@ function InstanceFilter(props) {
           >
             <MenuItem key="" value="">Show All</MenuItem>
             {
-              versions.map(({version}) => {
+              (versions || []).map(({version}) => {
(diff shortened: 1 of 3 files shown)


frontend/src/js/components/Instances/List.js
```diff
@@ -206,7 +219,7 @@ function ListView(props) {
                 versions={group.version_breakdown}
                 onFiltersChanged={onFiltersChanged}
                 filter={filters}
-                disabled={instances.length == 0}
+                disabled={!instances || instances.length === 0}
```
krnowak: Why `===`?

frontend/src/js/components/Instances/List.js
```diff
@@ -94,8 +97,9 @@ function ListView(props) {
   let {application, group} = props;
   const [page, setPage] = React.useState(0);
   const [rowsPerPage, setRowsPerPage] = React.useState(10);
-  const [instances, setInstances] = React.useState([]);
+  const [instances, setInstances] = React.useState(null);
```
krnowak: I don't quite understand why we use `null` instead of `[]` when we have all those length checks already. Is there something that behaves differently if it gets `null` instead of an empty array?

frontend/src/js/components/Instances/List.js
joaquimrocha: It's strict equality and some lints warn when comparing basic types like ints (we don't have such a lint in Nebraska yet but it's in our plans).

frontend/src/js/components/Instances/List.js
joaquimrocha: Yes. The difference is that we were showing the loading status (a progress circle) when the list was empty since that was how it was defined until the list of instances is finally fetched.
However, if nothing is returned, the instances list would remain empty and the loading state would be shown forever.

I thought the commit message was enough to understand this since this pattern is used in other places of the UI, but now I will update it with this explanation.

This should happen very often, but it's one of those things that make a product feel more polished when it does happen.

joaquimrocha:
Merging, since I addressed the questions.

PR #125, opened 2020-01-23
backend: Ignore instances with bracketed IDs in more places

krnowak:
This is to avoid skews in statistics or metrics resulting from taking
fake instances into account.


pkg/api/instances.go
```diff
@@ -266,7 +271,7 @@ func (api *API) instancesQuery(p InstancesQueryParams) *dat.SelectDocBuilder {
 		From("instance_application").
 		Where("application_id = $1 AND group_id = $2", p.ApplicationID, p.GroupID).
 		Where(fmt.Sprintf("last_check_for_updates > now() at time zone 'utc' - interval '%s'", validityInterval)).
-		Where(`instance_id NOT SIMILAR TO '\{[a-fA-F0-9-]{36}\}'`).
+		Where(ignoreFakeInstanceCondition("instance_id")).
```
joaquimrocha: :+1:

PR #127, opened 2020-01-27
auth/github: support GitHub Enterprise

iaguis:
If the flag `-gh-enterprise-url` is specified, use that as endpoint for
GitHub authentication following the GitHub Enterprise conventions.

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -138,6 +138,14 @@ generated on github (`https://github.com/settings/tokens`).
 
 Personal access token requires just one scope: `read:org`.
 
+# GitHub Enterprise
+
+If you want to authenticate against GitHub Enterprise, set `-gh-enterprise-url`
+to the URL of your GitHub Enterprise instance, for example
+`https://github.myorganization.net`.
+
+After that, follow the same procedures as with GitHub.
+
 # Troubleshooting:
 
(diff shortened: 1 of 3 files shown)


docs/authorization.md
```diff
@@ -138,6 +138,14 @@ generated on github (`https://github.com/settings/tokens`).
 
 Personal access token requires just one scope: `read:org`.
 
+# GitHub Enterprise
+
+If you want to authenticate against GitHub Enterprise, set `-gh-enterprise-url`
```
joaquimrocha: nitpick: I would probably call this setting `-gh-endpoint-url` instead of enterprise.

iaguis: I'm on the fence on this.

I agree that `-gh-endpoint-url` is nicer but when the user sets it to a URL we assume it's an enterprise instance and we create a [GitHub enterprise client](https://github.com/kinvolk/nebraska/pull/127/files#diff-c3445bdc939c13da5ed6e83f91de4039R321), which has a slightly different behavior. So for example it won't work if you set the CLI flag to the public endpoint.

So I think I'd lean towards keeping it `-gh-enterprise-url`. Does that make sense?

joaquimrocha: Yup. That's fine.

PR #128, opened 2020-01-28
frontend: Make the access management URL configurable

joaquimrocha:
The access management was hardcoded since we were supposed to always
manage the access in that Github page. However, when using Github
Enterprise, that URL will be different.

This patch thus allows to specify the URL in the .env configuration.


joaquimrocha:
This approach was a quick way to build the frontend with the right URL, but a better idea is to have it coming from the backend (as its already set there). See https://github.com/kinvolk/nebraska/issues/130 .

Hence closing this PR.

PR #129, opened 2020-01-29
Disable cgo and enable goproxy in compilation

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@gmail.com>

While compiling Nebraska using docker-compose or even ``make container-nebraska`` a user might face errors because:

* golangci-lint is compiled outside of the container, and then used inside an Alpine container which doesn't have glibc. The same apply to other tools

* Network environments with performance problems might face issues while compiling, so using GOPROXY might bring performance improvement in compilation

To reproduce the behavior pre PR, please clone the master repo, and then run ``make container-nebraska`` or try using the docker-compose file :)

### DIFF
--- a/Dockerfile.nebraska
+++ b/Dockerfile.nebraska
@@ -1,6 +1,7 @@
 FROM alpine:3.10 as nebraska-build
 
-ENV GOPATH=/go
+ENV GOPATH=/go \
+    GOPROXY=https://proxy.golang.org
 
 RUN apk update && \
 	apk add git go nodejs npm ca-certificates make musl-dev bash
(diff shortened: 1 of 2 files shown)


joaquimrocha:
Hi @rikatz , I apologize for the time it's taking. Somehow this got under my radar after the initial review tagging!
It's late here now, but I promise tomorrow we will look into it.

rikatz:
Hi folks, sorry for taking so long. I've been traveling for business in the past 3 weeks and couldn't take a look into this.

As I've seen you already reached a better solution, so coming back home I'll give another try and see if now everything is fine.

Thank you!

joaquimrocha:
For sure! Thanks @rikatz .

PR #131, opened 2020-02-27
Fix access management URL

### LINKED ISSUE #130: Add right access management link for Github Enterprise
When authenticating against Github Enterprise, the access management link is not correct (points to the regular Github).

We should fix this, and since we already have the information in the backend about which Github flavor are we using, we should access that from the frontend (as opposed to add an independent option to the frontend).

Probably the right way to do this is to have an API endpoint that will return a "general configuration" map. This configuration will contain the authentication backend base URL, so we can use it for fixing this very issue, and any other useful data from the backend (to be used now or in the future).

joaquimrocha:
This PR allows us to set the access management URL according to the auth backend being used, i.e. fixes #130 and makes it more future-proof when we add more authentication backends.

### DIFF
--- a/cmd/nebraska/controller_test.go
+++ b/cmd/nebraska/controller_test.go
@@ -5,6 +5,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/kinvolk/nebraska/cmd/nebraska/auth"
 )
 
 func TestGetRequestIP(t *testing.T) {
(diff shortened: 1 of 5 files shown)


cmd/nebraska/controller_test.go
```diff
@@ -27,3 +29,49 @@ func TestGetRequestIP(t *testing.T) {
...
+func TestClientConfig(t *testing.T) {
+	// Test the client configuration with the "noop" auth backend
+	noopAuthConfig := &auth.NoopAuthConfig{}
+	conf := &controllerConfig{
+		noopAuthConfig: noopAuthConfig,
+	}
+
+	ctl, err := newController(conf)
+
+	assert.Nil(t, err, "Couldn't create controller")
```
krnowak: I think there is an `assert.NoError` function too. It's better for error checking.

cmd/nebraska/controller_test.go
```diff
@@ -27,3 +29,49 @@ func TestGetRequestIP(t *testing.T) {
...
+
+	// Test the client configuration with the Github auth backend
+	ghAuthConfig := &auth.GithubAuthConfig{}
+	conf = &controllerConfig{
+		githubAuthConfig: ghAuthConfig,
+	}
+
+	ctl, err = newController(conf)
+
+	assert.Nil(t, err, "Couldn't create controller")
```
krnowak: Same here.

cmd/nebraska/controller_test.go
```diff
@@ -27,3 +29,49 @@ func TestGetRequestIP(t *testing.T) {
...
+	ghAuthConfig = &auth.GithubAuthConfig{
+		EnterpriseURL: phonyURL,
+	}
+	conf = &controllerConfig{
+		githubAuthConfig: ghAuthConfig,
+	}
+
+	ctl, err = newController(conf)
+
+	assert.Nil(t, err, "Couldn't create controller")
```
krnowak: Same here.

joaquimrocha:
Updated the branch with the `NoError` assert.

PR #138, opened 2020-03-04
Pagination fix for activity list

### LINKED ISSUE #133: Fix pagination in the activity list
The activity list's pagination is getting a scrollbar which harms the UX.
I think the cause is that the width of that area is not enough to contain the pagination itself when its page information is more than just 2 digits (but I am not 100% sure).

Screenshot from 2020-02-28 17-02-30

We should fix it, even if we need to expand the whole area a bit, or make the font-size for the pagination smaller.

ashu8912:
resolves #133

### DIFF
--- a/frontend/src/js/components/Activity/Container.react.js
+++ b/frontend/src/js/components/Activity/Container.react.js
@@ -1,6 +1,7 @@
 import Box from '@material-ui/core/Box';
 import Grid from '@material-ui/core/Grid';
 import Paper from '@material-ui/core/Paper';
+import { makeStyles } from '@material-ui/core/styles';
 import TablePagination from '@material-ui/core/TablePagination';
 import moment from 'moment';
 import React from "react";
@@ -11,7 +12,20 @@ import ListHeader from '../Common/ListHeader';
 import Loader from '../Common/Loader';
 import List from "./List.react";
 
+const useStyles = makeStyles({
+  toolbar: {
+    padding: 0,
+  },
+  caption: {
+    fontSize: '.85em',
+  },
+  select: {
+    fontSize: '.85em',
+  }
+});
+
 function Container(props) {
+  const classes = useStyles();
   const [activity, setActivity] = React.useState(getActivityEntries());
   const rowsOptions = [5, 10, 50];
   const [page, setPage] = React.useState(0);
@@ -99,6 +113,7 @@ function Container(props) {
             </Grid>
             <Grid item>
               <TablePagination
+                classes={classes}
                 rowsPerPageOptions={rowsOptions}
                 component="div"
                 count={activity.length}


frontend/src/js/components/Main.react.js
```diff
@@ -27,19 +30,34 @@ const nebraskaTheme = createMuiTheme({
...
+  overrides:{
+    MuiToolbar:{
+      gutters:{
+        [defaultTheme.breakpoints.up('sm')]: {
+          paddingLeft: '8px',
+          paddingRight: '8px',
+        }
+      }
+    }
+  }
```
joaquimrocha: This will change the whole theme.
I think this problem is local enough that we should just override the styling for the pagination used directly (if needed, we can create a styled component, but I think the only place this issue refers to is the activities pagination).

frontend/src/js/components/Main.react.js
```diff
@@ -27,19 +30,34 @@ const nebraskaTheme = createMuiTheme({
...
+    padding:0
+  }
+})
 export default function Main() {
   React.useEffect(() => {
     document.title = process.env.PROJECT_NAME;
   },
   []);
-
+ const classes=useStyles();
```
joaquimrocha: I prefer these declarations up in the first lines of the function declaration (as it's commonly used this way throughout the project).

frontend/src/js/components/Activity/Container.react.js
```diff
@@ -11,6 +12,14 @@ import ListHeader from '../Common/ListHeader';
 import Loader from '../Common/Loader';
 import List from "./List.react";
 
+const useStyles = makeStyles(theme => ({
+  root: {
+    overflow:'hidden',
```
joaquimrocha: Is this necessary with the font reduction? I'd prefer to find a solution that didn't turn off the scrollbars (they may be needed for accessibility).
Maybe reducing the inner margins/padding together with the font changes is enough?

joaquimrocha: Please notice that usually a space is used between the colon and the values.

frontend/src/js/components/Activity/Container.react.js
```diff
@@ -11,6 +12,14 @@ import ListHeader from '../Common/ListHeader';
...
+    fontSize:"0.685rem"
```
joaquimrocha: The project has been using single quotes for Js strings when possible.

frontend/src/js/components/Activity/Container.react.js
```diff
@@ -11,6 +12,14 @@ import ListHeader from '../Common/ListHeader';
 import Loader from '../Common/Loader';
 import List from "./List.react";
 
+const useStyles = makeStyles(theme => ({
```
joaquimrocha: Nitpick: If you're not using values from the `theme`, you don't need to include it (i.e. you can pass an object directly instead of a function). (I know other places do use this too, so this is a nitpick to avoid having to clean it up when we set a lint)

frontend/src/js/components/Main.react.js
ashu8912: @joaquimrocha overriding it in a theme will be good as what i see is every pagination in our project should have this much padding for gutters higher gutter padding value to paginations make the entire pagination move horizontally on clicks.
This fixes that problem also if we would want such a behaviour later in some pagination in future we can always override settings for that pagination.
But right now setting the gutter padding value in theme is a good idea as most of our pagination will be needing it.

ashu8912: ok as discussed shifting override of gutter padding to Activity List will be a better approach so going with that

joaquimrocha:
I added a simpler approach which doesn't touch the main theme.
Merging now to make the release (as the impact is minimal anyway).

PR #139, opened 2020-03-04
Frontend : Fixed visual clutter for package select

### LINKED ISSUE #132: Restyle list of packages select
When selecting a package from the channel edition, the list of packages shows the creation date at the same level as the package's name which clutters things visually.

Screenshot from 2020-02-28 10-28-06

Instead, we should use the list item's secondary text for displaying the date.

ashu8912:
The package name and createdAt field seem to arise in the same line which lead to bad user experience
made the createdAt field the secondary text for each package Item.

fixes #132


ashu8912:
aah i think my prettier configuration added some extra beautify changes.
Hey @joaquimrocha i was thinking of adding a pretiierrc and eslint config to the project so that every contributor should use the same configuration.

joaquimrocha:
I fixed the alignment of the code, and rewrote the commit message to be consistent with the convention used.

PR #143, opened 2020-03-05
docs: Update/fix the recommended way of running the database

joaquimrocha:
The documentation about running a database for development was still
using the container built by CoreRoller (and with the name updated to
Nebraska by mistake).

Instead of relying on a dedicated container for the database, we should
just provide the instructions to quickly set it up for Nebraska. So
these changes replace the mentioned legacy instructions this way.

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -7,12 +7,28 @@ tokens](https://github.com/settings/tokens) can be used as bearer
 token. If no bearer token is sent, Nebraska will authorize users
 through the GitHub Oauth flow and create a persistent session.
 
+# Preparing the database for Nebraska
+
+Nebraska uses the `PostgreSQL` database, and expects the used database to
+be set to the UTC timezone.
+
+For a quick setup of `PostgreSQL` for Nebraska's development, you can use
+the `postgres` container as follows:
+
+- Start `Postgres`:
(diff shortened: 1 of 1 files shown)


docs/authorization.md
```diff
@@ -7,12 +7,28 @@ tokens](https://github.com/settings/tokens) can be used as bearer
...
+
 # Deploying Nebraska for testing on local computer (noop authentication)
 
 - Go to the nebraska project directory and run `make all`
 
-- Start the database: `docker run --privileged -d -p
-  127.0.0.1:5432:5432 nebraska/postgres`
+- Start the database (see the section above if you need a quick setup).
 
 - Start the Nebraska backend:
```
schu: Here we should probably mention that `NEBRASKA_DB_URL` needs to be set for it to work:

```
NEBRASKA_DB_URL=postgres://postgres:nebraska@localhost:5432/nebraska?sslmode=disable LOGXI=* nebraska ...
```

joaquimrocha: It works here without that if I tested it correctly.

joaquimrocha:
> Looks good to me modulo comment.
> 
> I would rather put that paragraph into the README (or a new e.g. `docs/developer.md` document), but we can do that later.

Yup, let's do that later.

PR #147, opened 2020-03-05
Fix architecture sticky behaviour in channel list

### LINKED ISSUE #145: Don't show the architecture overlapping the channels list
The architecture is oveerlapping the channels' list when scrolling up.
Screenshot from 2020-03-05 18-04-51

We should not make that sticky.

ashu8912:
fixes #145

### DIFF
--- a/frontend/src/js/components/Channels/List.react.js
+++ b/frontend/src/js/components/Channels/List.react.js
@@ -33,7 +33,7 @@ function ChannelList(props) {
       {Object.entries(getChannelsPerArch()).map(([arch, channels]) =>
         <MuiList
           key={arch}
-          subheader={<ListSubheader>{ARCHES[arch]}</ListSubheader>}
+          subheader={<ListSubheader disableSticky >{ARCHES[arch]}</ListSubheader>}
           dense
         >
           {channels.map(channel =>


frontend/src/js/components/Channels/List.react.js
```diff
@@ -33,7 +33,7 @@ function ChannelList(props) {
       {Object.entries(getChannelsPerArch()).map(([arch, channels]) =>
         <MuiList
           key={arch}
-          subheader={<ListSubheader>{ARCHES[arch]}</ListSubheader>}
+          subheader={<ListSubheader disableSticky={true}>{ARCHES[arch]}</ListSubheader>}
```
joaquimrocha: `Nitpick:` AFAIU in JSX, if you provide a property without assigning it a value, it will be set to `true`.
This means you should have to set `={true}`, meaning it will be more readable.

ashu8912: yes @joaquimrocha you are right keeping it just the prop name is more readable ,i have made commit that removes the ={true}

ashu8912:
Screenshot 2020-03-06 at 10 59 17 PM

joaquimrocha:
Merged to master. Thanks.

PR #152, opened 2020-03-09
Added / to packages URL so the packages are downloaded properly when …

Lplazier:
…Nebraska is configured to host and sync packages

controller.go Syncer config is missing a slash at the end "/flatcar" of the packages URL creating the following URL when packages are loaded

https://nebraskaserver/flatcarflatcar-amd64-2345.3.0.gz

This results in the file not being found.

I0304 08:05:11.067962  3574 action_processor.cc:82] ActionProcessor::ActionComplete: finished OmahaRequestAction, starting DownloadAction
I0304 08:05:11.068303  3574 install_plan.cc:53] InstallPlan: , new_update, url: https://nebraska server/flatcarflatcar-amd64->
I0304 08:05:11.068821  3574 update_attempter.cc:408] Download status: active
I0304 08:05:11.069169  3574 multi_range_http_fetcher.cc:30] starting first transfer
I0304 08:05:11.069728  3574 multi_range_http_fetcher.cc:58] starting transfer of range 0+?
I0304 08:05:11.070104  3574 libcurl_http_fetcher.cc:48] Starting/Resuming transfer
I0304 08:05:11.070497  3574 libcurl_http_fetcher.cc:152] Setting up curl options for HTTP
I0304 08:05:11.070989  3574 libcurl_http_fetcher.cc:427] Setting up timeout source: 1 seconds.
E0304 08:05:12.006531  3574 libcurl_http_fetcher.cc:243] Unable to get http response code: Empty reply from server
I0304 08:05:12.007259  3574 libcurl_http_fetcher.cc:274] Transfer resulted in an error (0), 0 bytes downloaded
I0304 08:05:12.007503  3574 multi_range_http_fetcher.cc:151] Received transfer complete.
I0304 08:05:12.007741  3574 multi_range_http_fetcher.cc:108] TransferEnded w/ code 0
I0304 08:05:12.007957  3574 multi_range_http_fetcher.cc:142] Done w/ all transfers
I0304 08:05:12.013857  3574 update_attempter.cc:408] Download status: inactive

When the slash is added to the package URL via the frontend, the packages are downloaded by the system and the update. 
Added the slash at the the end "/flatcar" of the packages URL

https://nebraska server/flatcar/flatcar-amd64-2345.3.0.gz

### DIFF
--- a/cmd/nebraska/controller.go
+++ b/cmd/nebraska/controller.go
@@ -59,7 +59,7 @@ func newController(conf *controllerConfig) (*controller, error) {
 			API:          conf.api,
 			HostPackages: conf.hostFlatcarPackages,
 			PackagesPath: conf.flatcarPackagesPath,
-			PackagesURL:  conf.nebraskaURL + "/flatcar",
+			PackagesURL:  conf.nebraskaURL + "/flatcar/",
 		}
 		syncer, err := syncer.New(syncerConf)
 		if err != nil {


joaquimrocha:
@Lplazier , Thanks a lot for your report and PR.
It seems like the actual problem is that the `update_engine` will create the package URL from the update reponse from omaha (in the form of base URL + file name) by concatenating the parts directly, instead of joining them using a slash when needed.

So your solution may fix the symptoms and I we should include it (I will add a couple of comments in a sec), but I have opened https://github.com/flatcar-linux/update_engine/issues/3 for making sure this doesn't happen in the future.

Nebraska's syncer is also joining the URLs directly, so we should address that (we can do it on top of your PR, no problem).

cmd/nebraska/controller.go
```diff
@@ -59,7 +59,7 @@ func newController(conf *controllerConfig) (*controller, error) {
 			API:          conf.api,
 			HostPackages: conf.hostFlatcarPackages,
 			PackagesPath: conf.flatcarPackagesPath,
-			PackagesURL:  conf.nebraskaURL + "/flatcar",
+			PackagesURL:  conf.nebraskaURL + "/flatcar/",
```
joaquimrocha: This fixes the issue quickly, but a safer thing to do would be to programatically add the `/` if missing. Otherwise this code can be changed in the future without a developer noticing their breaking things.
As I mentioned in the PR dedicated comment, the real fix should be done at the "consumer" level when getting a response with the URL + file name, in order to join the URL correctly instead of simply concatenating it.

However, this PR (with the changes I am proposing) is good to make sure all versions of the update_server (prior to when https://github.com/flatcar-linux/update_engine/issues/3 gets fixed) will keep working with it.

joaquimrocha:
PR #159 includes an improved version of this PR. Thanks @Lplazier .

Issue #150, opened 2020-03-07
Allow to see instance's history when the list is filtered


joaquimrocha:
When in the instances view, if one filters the instances (e.g. to see only the ones with errors), it is impossible to open their history (clicking the IP).

Steps to reproduce:
1. Go to a group's instance view and filter the instances by status (e.g. choose the Error).
2. Click the IP address, to open the instance's quick history view.
3. Verify that the quick history is not opened, instead, the filter is reset (I think this is the cause of why the history is not shown).

Acceptance criteria:
1. Follow the steps above until the second one.
2. Verify the history view is opened and the filter is not reset.

ashu8912:
@joaquimrocha this was merged in master and thus this issue can be closed

joaquimrocha:
Fixed in version 0.2.5.

PR #153, opened 2020-03-10
Fix instance history view when filter status changed

ashu8912:


ashu8912:
fixes #150

ashu8912:
ezgif com-video-to-gif

frontend/src/js/components/Instances/List.js
```diff
@@ -101,7 +101,7 @@ function ListView(props) {
   const [filteredInstances, setFilteredInstances] = React.useState([]);
   const [lastCheck, setLastCheck] = React.useState(moment([0, 0])); // Long long time ago.
   const [filters, setFilters] = React.useState({status: '', version: ''});
-
+
```
joaquimrocha: Remove this.

frontend/src/js/components/Instances/List.js
```diff
@@ -118,7 +118,7 @@ function ListView(props) {
   function applyFilters(_filters={}) {
     let newFilters = _filters || {status: '', version: ''};
     setFilters(newFilters);
-
+
```
joaquimrocha: Remove this.

frontend/src/js/components/Instances/List.js
```diff
@@ -141,12 +141,14 @@ function ListView(props) {
...
-      setInstances(cachedInstances);
+    if (filteredInstances.length == 0) {
       setFilteredInstances(cachedInstances);
+    }
+    if (!instances || instances.length == 0) {
+      setInstances(cachedInstances);
       return;
     }
-
+
```
joaquimrocha: Remove this.

frontend/src/js/components/Instances/List.js
```diff
@@ -156,11 +158,7 @@ function ListView(props) {
         setInstances(cachedInstances);
       }
     }
-
-    if (filteredInstances.length == 0) {
-      setFilteredInstances(cachedInstances);
-    }
-  }
+}
```
joaquimrocha: This bracket is not correctly aligned.

frontend/src/js/components/Instances/List.js
```diff
@@ -141,12 +141,14 @@ function ListView(props) {
 
   function onChangeInstances() {
     let cachedInstances = instancesStore.getCachedInstances(application.id, group.id) || [];
-    if (!instances || instances.length == 0 || filteredInstances.length == 0) {
-      setInstances(cachedInstances);
+    if (filteredInstances.length == 0) {
       setFilteredInstances(cachedInstances);
+    }
```
joaquimrocha: The problem seems to have to do with filtering the instances with the wrong filter. So I am surprised that this switch fixes the issue for all cases. Does this still work correctly if you go to other pages in the table?

I'd like a better explanation of what the exact cause of the issue was in order to understand how this fix is enough. I will give anther look in the morning in case I am just too tired to see exactly what's going on.

ashu8912: yes you are right @joaquimrocha the above switch doesn't seem to solve the problem and moreover the above code change doesn't have any effect to how the stituation was before.I tested the frontend with the changes and it seemed to work but a strange thing is i am not able to reproduce this bug now.

frontend/src/js/components/Instances/List.js
ashu8912: As this commit was updated the formatting errors are no longer there

frontend/src/js/components/Instances/List.js
```diff
@@ -179,7 +179,7 @@ function ListView(props) {
       instancesStore.removeChangeListener(onChangeInstances);
     };
   },
-  [lastCheck, instances]);
+  [lastCheck, instances,filteredInstances]);
```
joaquimrocha:
```suggestion
  [lastCheck, instances, filteredInstances]);
```

frontend/src/js/components/Instances/List.js
```diff
@@ -156,11 +156,7 @@ function ListView(props) {
         setInstances(cachedInstances);
       }
     }
-
-    if (filteredInstances.length == 0) {
-      setFilteredInstances(cachedInstances);
-    }
-  }
+}
```
joaquimrocha: This is still not correctly aligned.
And this commit title is `frontend: Fix Remove ..`, it should be `frontend: Remove ...`.
Also, the context for the title is the whole frontend project, and it would be more helpful to mention at least what component it is talking about. So I guess something like `frontend: Remove unreachable block in the Instances/List component` .

When reading a commit message, usually a developer needs some more context than what's writting below. You should say something like: `This patch removes an if block from the function XYZ of the ZYX component, as it cannot be reached due to prior logic in the function.`
```
The if statement is an unreachable code segment as a similar condition
is present in codebase above this codebase with a return.
```

```
This patch removes that code segment
                                    ^ missing period (but not needed if you go with a similar message that already says this)
```

ashu8912: Thankyou @joaquimrocha i have made these changes.

joaquimrocha:
The 1st commit still had a confusing message. I have fixed it. Pushing after CI finishes.

PR #155, opened 2020-03-12
Enhance Dev Environment(make it similar to create-react-app)

ashu8912:


joaquimrocha:
As discussed with @ashu8912 , I think we should try a solution that involves create-react-app, rather than introducing complex webpack confs.

Issue #149, opened 2020-03-06
Add Js linter & commit hooks


joaquimrocha:
We should add a Js linter (locally when building, and in the CI), and also a commit hook (I have a half finished one for checking the commit messages).

Details TBD.

ashu8912:
yes also a precommit-hook will be helpful that runs linter and prettier config before each commit

ashu8912:
@joaquimrocha whats the status for the commit hook

joaquimrocha:
We don't have any specific hooks but we do have the linter, so let's close this and maybe look into the hook later.

PR #162, opened 2020-03-19
Add linter(eslint)

ashu8912:


ashu8912:
fixes #149

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+    "parser": "babel-eslint",
+    "plugins": [
+      "react"
+    ],
+    "rules": {
+      "space-in-parens": [ 0, "always" ],
+      "template-curly-spacing": [ 2, "always" ],
+      "array-bracket-spacing": [ 2, "always" ],
+      "object-curly-spacing": [ 2, "always" ],
+      "computed-property-spacing": [ 2, "always" ],
```
joaquimrocha: I would have to test how these work against what we have. I think it's better to have these turned off/set to never.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "no-multiple-empty-lines": [ 2, { "max": 1, "maxEOF": 0, "maxBOF": 0 } ],
```
joaquimrocha: I want a newline at the end of files (so Github and other tools don't warn us). Do we need `maxEOF` to be 1?

frontend/.eslintrc
joaquimrocha: I mean it should be allowed, but not enforced at the moment. And we can change it later.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "quotes": [ 1, "single", "avoid-escape" ],
```
joaquimrocha: `avoid-escape` seems to be deprecated in favor or `avoidEscape`: https://eslint.org/docs/rules/quotes#top
We also want `allowTemplateLiterals`.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "no-use-before-define": [ 2, { "functions": false } ],
```
joaquimrocha: I don't want to enforce this for now (nor to have a warning).

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "semi": [0, "never"],
```
joaquimrocha: I want us to have semicolons.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "prefer-const": 1,
```
joaquimrocha: Cool.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "react/prefer-es6-class": 0,
+      "react/jsx-filename-extension": 0,
+      "react/jsx-curly-spacing": [ 2, "always" ],
```
joaquimrocha: I'm on the fence with this as I guess we'll have to change a lot of code. But as a rule I like it.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "react/jsx-indent": [ 2, 4 ],
```
joaquimrocha: Let's use 2-spaces as we've been doing.
Same for the Js code (maybe you need to add the rule).

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "class-methods-use-this": [ 1 ],
```
joaquimrocha: Not really important IMO. So maybe let's not bother.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "no-case-declarations": [ 1 ],
```
joaquimrocha: :+1:

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "no-return-assign": [ 1 ],
```
joaquimrocha: Let's go with error and always here.

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "no-param-reassign": [ 1 ],
```
joaquimrocha: :+1:

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "no-shadow": [ 1 ],
```
joaquimrocha: :+1:

frontend/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+      "no-underscore-dangle" : [0, "always"]
```
joaquimrocha: :+1:

frontend/package.json
```diff
@@ -44,17 +44,21 @@
     "@iconify/react": "^1.1.1",
     "@iconify/tools": "^1.3.9",
     "babel-loader": "^8.0.6",
+    "customize-cra": "^0.9.1",
```
joaquimrocha: Why is this needed?

PR #165, opened 2020-03-24
Fix eslint setup

ashu8912:


frontend/.env
```diff
@@ -1,3 +1,4 @@
 REACT_APP_PROJECT_NAME=Nebraska
 REACT_APP_PRIMARY_COLOR=#09bac8
 REACT_APP_APPBAR_BG=
+EXTEND_ESLINT=true
```
joaquimrocha: Remove this, since we no longer need it, right?

frontend/.eslintrc
```diff
@@ -5,6 +5,7 @@
     "simple-import-sort"
   ],
   "rules": {
+    "eqeqeq": [2,"always"],
```
joaquimrocha: This should go in a separate commit, as it's not part of fixing the lint.

frontend/package.json
```diff
@@ -43,25 +43,25 @@
...
+    "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint 'src/**/*.{js,jsx}'",
     "lint:fix": "eslint --fix 'src/**/*.{js,jsx}'"
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": "react-app",
+    "rules": {
+    }
```
joaquimrocha: This is new, do we need this?

PR #167, opened 2020-04-01
Added time interval for Charts

ashu8912:


pkg/api/groups.go
```diff
@@ -307,21 +307,29 @@ func (api *API) groupInstancesStatusQuery() string {
 		InstanceStatusDownloaded, InstanceStatusDownloading, InstanceStatusOnHold, validityInterval, ignoreFakeInstanceCondition("instance_id"))
 }
 
-func (api *API) GetGroupVersionCountTimeline(groupID string) (map[time.Time](VersionCountMap), error) {
+func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (map[time.Time](VersionCountMap), error) {
 	var timelineEntry []VersionCountTimelineEntry
+	var interval string
+	switch duration {
```
joaquimrocha, line 313: I think the duration variable needs to be sanitized. i.e. it should be validated and matched against the acceptable  options.
I'd also not use the format of PostgresSQL (with spacing and all), but something like `Xd` -> X days.

pkg/api/groups.go
```diff
@@ -307,21 +307,29 @@ func (api *API) groupInstancesStatusQuery() string {
...
 		 instance_versions AS (SELECT instance_id, created_ts, version, status FROM instance_status_history WHERE instance_id IN (SELECT instance_id FROM recent_instances) AND status = 4 UNION (SELECT * FROM recent_instances) ORDER BY created_ts DESC)
-	SELECT ts, (CASE WHEN version IS NULL THEN '' ELSE version END), sum(CASE WHEN version IS NOT null THEN 1 ELSE 0 END) total FROM (SELECT * FROM time_series LEFT JOIN LATERAL(SELECT distinct ON (instance_id) instance_Id, version, created_ts FROM instance_versions WHERE %[3]s AND created_ts <= time_series.ts ORDER BY instance_Id, created_ts DESC) _ ON true) AS _
+	SELECT ts, (CASE WHEN version IS NULL THEN '' ELSE version END), sum(CASE WHEN version IS NOT null THEN 1 ELSE 0 END) total FROM (SELECT * FROM time_series LEFT JOIN LATERAL(SELECT distinct ON (instance_id) instance_Id, version, created_ts FROM instance_versions WHERE %[4]s AND created_ts <= time_series.ts ORDER BY instance_Id, created_ts DESC) _ ON true) AS _
 	GROUP BY 1,2
 	ORDER BY ts DESC;
-	`, validityInterval, ignoreFakeInstanceCondition("instance_id"),
+	`, duration, interval, ignoreFakeInstanceCondition("instance_id"),
```
joaquimrocha, line 331: We should use whatever is matched from the query's params.

frontend/src/js/api/API.js
```diff
@@ -80,12 +80,12 @@ class API {
     return API.doRequest('PUT', url, JSON.stringify(groupData));
   }
 
-  static getGroupVersionCountTimeline(applicationID, groupID) {
-    return API.getJSON(BASE_URL + '/apps/' + applicationID + '/groups/' + groupID + '/version_timeline');
+  static getGroupVersionCountTimeline(applicationID, groupID,duration) {
```
joaquimrocha, line 83: A space is needed after the comma. I am surprised our linter didn't catch this.
Same for the usage below (where applicable).

frontend/src/js/components/Common/TimeIntervalLinks.js
```diff
@@ -0,0 +1,32 @@
...
+import timeIntervals from '../../constants/timeInterval';
+
+const useStyles = makeStyles(theme => ({
+  title:{
+    fontSize:'1rem'
+  }
+}));
+
+export default function TimeIntervalLinks(props){
+  const [activeLink,setActiveLink]=React.useState(timeIntervals[0]);
```
joaquimrocha, line 12: Hmm, the `=` needs to be surrounded by spaces.

frontend/src/js/components/Common/TimeIntervalLinks.js
```diff
@@ -0,0 +1,32 @@
...
+import React from 'react';
+import timeIntervals from '../../constants/timeInterval';
+
+const useStyles = makeStyles(theme => ({
+  title:{
+    fontSize:'1rem'
+  }
+}));
+
+export default function TimeIntervalLinks(props){
```
joaquimrocha, line 11: The `{` needs a space before it.

frontend/src/js/constants/timeInterval.js
```diff
@@ -0,0 +1,3 @@
+const timeIntervals=['30 days','7 days','1 day','1 hour'];
```
joaquimrocha, line 1: `=` should be surrounded by spaces.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -8,17 +8,19 @@ import React from 'react';
 import { Area, AreaChart, CartesianGrid, Tooltip, XAxis, YAxis } from 'recharts';
 import semver from 'semver';
 import _ from 'underscore';
-import { cleanSemverVersion, getMinuteDifference,makeColorsForVersions, makeLocaleTime } from '../../constants/helpers';
-import { applicationsStore, instancesStore } from '../../stores/Stores';
+import { cleanSemverVersion, getMinuteDifference, makeColorsForVersions, makeLocaleTime } from '../../constants/helpers';
+import { applicationsStore, groupChartStore,instancesStore } from '../../stores/Stores';
```
joaquimrocha, line 12: More formatting issues on the `,` here.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -63,17 +61,16 @@ function TimelineChart(props) {
...
     // The midnight tick just gets the date, not the hours (since they're zero)
-    ticks[midnightTick] = makeLocaleTime(midnightDay, {showTime: false});
-
+    ticks[midnightTick] = makeLocaleTime(midnightDay, { showTime: false });
     return ticks;
   }
 
   function TimelineTooltip(props) {
-    const {label, data} = props;
+    const { label, data } = props;
```
joaquimrocha, line 73: This shouldn't be enforcible IIRC.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -86,47 +83,47 @@ function TimelineChart(props) {
...
-        onClick={({activeLabel}) => {props.onSelect(activeLabel);}}
-      >
+      width={width}
+      height={height}
+      data={props.data}
+      margin={{
+        top: 10, right: 30, left: 0, bottom: 0,
+      }}
+      onClick={({ activeLabel }) => { props.onSelect(activeLabel); }}
+    >
```
joaquimrocha, line 95: All these don't seem to be related to the PR.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -86,47 +83,47 @@ function TimelineChart(props) {
...
-        />
+        dataKey="index"
+        type="number"
+        interval={0}
+        domain={[0, 'dataMax']}
+        ticks={Object.keys(getTickValues(4))}
+        tickFormatter={index => {
+          return ticks[index];
+        }}
+      />
```
joaquimrocha, line 107: Ditto.

cmd/nebraska/controller.go
```diff
@@ -351,8 +351,12 @@ func (ctl *controller) getGroups(c *gin.Context) {
 
 func (ctl *controller) getGroupVersionCountTimeline(c *gin.Context) {
 	groupID := c.Params.ByName("group_id")
-
-	versionCountTimeline, err := ctl.api.GetGroupVersionCountTimeline(groupID)
+	durationMap := c.Request.URL.Query()
+	var duration string
+	if _, ok := durationMap["duration"]; ok {
+		duration = durationMap["duration"][0]
+	}
```
krnowak, line 358: This could be shortened to `duration := c.Query("duration")`.

cmd/nebraska/controller.go
```diff
@@ -368,8 +372,12 @@ func (ctl *controller) getGroupVersionCountTimeline(c *gin.Context) {
 
 func (ctl *controller) getGroupStatusCountTimeline(c *gin.Context) {
 	groupID := c.Params.ByName("group_id")
-
-	statusCountTimeline, err := ctl.api.GetGroupStatusCountTimeline(groupID)
+	durationMap := c.Request.URL.Query()
+	var duration string
+	if _, ok := durationMap["duration"]; ok {
+		duration = durationMap["duration"][0]
+	}
```
krnowak, line 379: Same here about shortening.

pkg/api/groups.go
```diff
@@ -307,21 +307,29 @@ func (api *API) groupInstancesStatusQuery() string {
...
+func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (map[time.Time](VersionCountMap), error) {
 	var timelineEntry []VersionCountTimelineEntry
+	var interval string
+	switch duration {
+	case "30 days":
+		interval = "24 hour"
+	case "7 days":
+		interval = "8 hour"
+	default:
+		interval = "1 hour"
```
krnowak, line 319: I think that here normally we should either do something like `log.Warnf("invalid duration %s, falling back to 1 hour interval", duration)` or, preferably since we apparently do no logging at all in `pkg/api`, `return fmt.Errorf("invalid duration %s", duration)`.

pkg/api/groups.go
krnowak: If we go with a finite set of valid options, that you could use an enumeration here (like we do with architecture in `arch.go` and then translate the enumeration into pair of postgres strings:

```go
type Duration int

const (
	OneDay Duration = iota
	SevenDays
	ThirtyDays
)
```

then have something like:

```go
// durationToPostgresTimings returns a postgres duration and interval for
// passed duration enumeration value
func durationToPostgresTimings(duration Duration) (string, string, error) {
	switch (duration) {
	case ThirtyDays:
		return "30 days", "24 hour", nil
	case SevenDays:
		return …
	case OneDay:
		return …
	default:
		return "", "", fmt.Errorf("invalid duration enumeration value %d", duration)
	}
}
```

so you could use it later as `duration, interval, err := durationToPostgresTimings(durationParam)`

pkg/api/groups.go
```diff
@@ -375,7 +383,7 @@ func (api *API) GetGroupStatusCountTimeline(groupID string) (map[time.Time](map[
 		LEFT JOIN LATERAL(SELECT * FROM instance_status_history WHERE group_id=$1 AND %[2]s AND created_ts BETWEEN time_series.ts - INTERVAL '1 hour' + INTERVAL '1 sec' AND time_series.ts) _ ON TRUE) AS _
 	GROUP BY 1,2,3
 	ORDER BY ts DESC;
-	`, validityInterval, ignoreFakeInstanceCondition("instance_id"))
+	`, duration, ignoreFakeInstanceCondition("instance_id"))
```
krnowak, line 386: You could do similar stuff as in `GetGroupVersionCountTimeline`, but ignore the interval: `duration, _, err := durationToPostgresTimings(durationParam)`.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -269,9 +266,39 @@ export function VersionCountTimeline(props) {
 
   // Make the timeline data again when needed.
   React.useEffect(() => {
+    function getVersionTimeline(group) {
+      // Check if we should update the timeline or it's too early.
+      const lastUpdate = new Date(timeline.lastUpdate);
+      const cachedVersionChartData = groupChartStore.getVersionChartData(duration);
+      if (duration !== '1 hour' && cachedVersionChartData){
```
joaquimrocha: I don't think we should do this check here. Instead, we should just not cache the values for 1 hour, in the store you created.
This way the code gets much simpler, less error prone, and we can also decide to cache the values for 1 hour if needed, etc.

cmd/nebraska/controller.go
```diff
@@ -356,11 +356,7 @@ func (ctl *controller) getGroups(c *gin.Context) {
 
 func (ctl *controller) getGroupVersionCountTimeline(c *gin.Context) {
 	groupID := c.Params.ByName("group_id")
-	durationMap := c.Request.URL.Query()
-	var duration string
-	if _, ok := durationMap["duration"]; ok {
-		duration = durationMap["duration"][0]
-	}
+	duration := c.Query("duration")
```
joaquimrocha, line 364: This whole changes should be squashed with the original patches.

pkg/api/groups.go
```diff
@@ -341,7 +341,7 @@ func durationToPostgresTimings(duration Duration) (string, string, error) {
 	case OneDay:
 		return "1days", "1 hour", nil
 	case OneHour:
-		return "1hour", "5 minute", nil
+		return "1hour", "15 minute", nil
```
joaquimrocha: Belongs to the original patch that introduced the change.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -266,6 +275,7 @@ export function VersionCountTimeline(props) {
 
   // Make the timeline data again when needed.
   React.useEffect(() => {
+    let canceled = false;
```
joaquimrocha, line 286: The cancellation is nice but doesn't belong to this patch. You can add them in their own separate patch.

cmd/nebraska/controller.go
```diff
@@ -186,7 +187,7 @@ func (ctl *controller) updateApp(c *gin.Context) {
 		return
 	}
 
-	app, err = ctl.api.GetApp(app.ID)
+	app, err = ctl.api.GetApp(app.ID, validityInterval)
```
joaquimrocha, line 190: AFAIU: This is no longer needed because we now no longer return stats/versions in the groups, right?

frontend/src/js/api/API.js
```diff
@@ -13,8 +13,12 @@ class API {
...
   }
 
-  static getApplication(applicationID) {
-    return API.getJSON(BASE_URL + '/apps/' + applicationID);
+  static getApplication(applicationID, duration) {
+    if (!duration){
+      return API.getJSON(BASE_URL + '/apps/' + applicationID);
+
+    }
+    return API.getJSON(BASE_URL + '/apps/' + applicationID + `?duration=${duration}`);
```
joaquimrocha, line 21: As mentioned, this should no longer be needed.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -70,7 +70,15 @@ function TimelineChart(props) {
 
         const tick = getMinuteDifference(tickDate, startTs) * dimension / lengthMinutes;
         // Show only the time.
-        ticks[tick] = makeLocaleTime(tickDate, dateFormat);
+        if (lengthMinutes === 30 * 24 * 60){
```
joaquimrocha, line 73: I see you're adding the month ticks here. That's okay. Just please use the `DAY` constant as I suggested before.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -343,8 +352,8 @@ export function VersionCountTimeline(props) {
               </React.Fragment>
               :
               <Typography>
-                Showing for the last 24 hours
-                (click the chart to choose a different time point).
+                Showing for the last {duration.displayValue}
+                (click the interval to choose a different time point).
```
joaquimrocha, line 356: Use `click the chart` as it was before.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
     const app = applicationsStore.getCachedApplication(props.appID);
 
     if (!app) {
       applicationsStore.getApplication(props.appID);
-      return;
     }
-
-    if (app !== application) {
+    if (app !== application)
+    {
```
joaquimrocha, line 50: Doesn't belong here.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
       setApplication(app);
     }
 
     const groupFound = app ? _.findWhere(app.groups, {id: props.groupID}) : null;
     if (groupFound !== group) {
       setGroup(groupFound);
     }
+
```
joaquimrocha, line 58: Doesn't belong here.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
   }
+  function getDataForUpdateProgressChart(){
```
joaquimrocha: Add a space between `)` and `{`.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
+
+    const progressChartData = groupChartStore
+      .getUpdateProgressChartData(updateProgressChartDuration.queryValue);
+    if (!progressChartData) {
+      setLoadingUpdateProgressChart(true);
+      applicationsStore
+        .getApplicationWithDurationFilter(props.appID, updateProgressChartDuration.queryValue)
+        .then((application=>{
```
joaquimrocha: I think we have a rule for using a space around the `=>`.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
+    } else {
+      console.log(progressChartData);
+      setLoadingUpdateProgressChart(false);
+      setUpdateProgressChartData(progressChartData);
+    }
+  }
   function updateGroup() {
     props.handleUpdateGroup(props.groupId, props.appID);
   }
+  function updateProgressChart(duration){
```
joaquimrocha, line 63: Add a space between ) and {.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
+      applicationsStore
+        .getApplicationWithDurationFilter(props.appID, updateProgressChartDuration.queryValue)
+        .then((application=>{
+          setLoadingUpdateProgressChart(false);
+          const groupFound = _.findWhere(application.groups, {id: props.groupID});
 
+          setUpdateProgressChartData(groupFound.instances_stats);
+        }));
+    } else {
+      console.log(progressChartData);
```
joaquimrocha: Don't log this.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
+  function updateProgressChart(duration){
+    setUpdateProgressChartDuration(duration);
+  }
   function updateVersionChart(duration){
     setVersionChartSelectedDuration(duration);
   }
   function updateStatusChart(duration){
     setStatusChartDuration(duration);
   }
+  React.useEffect(()=>{
```
joaquimrocha: Spaces around the `=>` needed.

frontend/src/js/stores/ApplicationsStore.js
```diff
@@ -54,7 +54,9 @@ class ApplicationsStore extends Store {
         }
       });
   }
-
+  getApplicationWithDurationFilter(applicationID, duration){
```
joaquimrocha, line 57: Add a space between ) and {.

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -22,5 +23,15 @@ export default class GroupChartsStore{
   getStatusChartData(key){
     return this.StatusBreakdownChartData[key];
   }
+  setUpdateProgressChartData(key, data){
```
joaquimrocha: Add a space between ) and {.

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -22,5 +23,15 @@ export default class GroupChartsStore{
...
+    if (this.UpdateProgressChartData.hasOwnProperty(key))
+    {
+      return;
+    }
+    this.UpdateProgressChartData[key] = data;
+  }
+  getUpdateProgressChartData(key){
```
joaquimrocha: Add a space between ) and {.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,46 +29,75 @@ const useStyles = makeStyles({
...
+    getDataForUpdateProgressChart();
+  }, [updateProgressChartDuration]);
   React.useEffect(() => {
+
```
joaquimrocha, line 72: No new line needed.

pkg/api/groups.go
```diff
@@ -334,11 +334,11 @@ func (api *API) GetGroupInstancesStats(groupID string) (*InstancesStatusStats, e
 func durationToPostgresTimings(duration Duration) (string, string, error) {
 	switch duration {
 	case ThirtyDays:
-		return "30 days", "24 hour", nil
+		return "30 days", "6 days", nil
```
joaquimrocha, line 345: Let's try 24 hours now that we have a faster DB.
Also these changes should be squashed to their introductory patches.

frontend/package.json
```diff
@@ -61,8 +61,7 @@
   },
   "eslintConfig": {
     "extends": "react-app",
-    "rules": {
-    }
+    "rules": {}
```
joaquimrocha, line 64: Doesn't belong here.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -298,7 +298,7 @@ export function VersionCountTimeline(props) {
       } else {
         setTimelineChartData({data:[], keys:[], colors:[]});
         applicationsStore
-          .getGroupVersionCountTimeline(group.application_id, group.id, 
+          .getGroupVersionCountTimeline(group.application_id, group.id,
```
joaquimrocha, line 301: Doesn't belong here.

frontend/src/js/components/Instances/List.js
```diff
@@ -161,7 +166,10 @@ function ListView(props) {
   }
 
   React.useEffect(() => {
-    fetchInstances(filters, page, rowsPerPage);
+    const queryParams = new URLSearchParams(window.location.search);
+    const duration = queryParams.get('duration');
+    fetchInstances(filters, page, rowsPerPage, duration);
+
```
joaquimrocha, line 171: No newline needed here.

frontend/src/js/components/Instances/List.js
```diff
@@ -100,8 +100,9 @@ function ListView(props) {
   const [instancesObj, setInstancesObj] = React.useState({instances: [], total: 0});
   const [instanceFetchLoading, setInstanceFetchLoading] = React.useState(false);
   const [filters, setFilters] = React.useState({status: '', version: ''});
+  const [totalInstancesWithoutFilter, setTotalInstancesWithoutFilter] = React.useState(0);
+  function fetchInstances(filters, page, perPage, duration){
 
-  function fetchInstances(filters, page, perPage){
```
joaquimrocha: Add a newline before the function declaration.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -214,7 +214,9 @@ function ItemExtended(props) {
                     {!loadingUpdateProgressChart &&
                     <Link
                       className={classes.link}
-                      to={{pathname: `/apps/${props.appID}/groups/${props.groupID}/instances`}}
+                      to={{pathname: `/apps/${props.appID}/groups/${props.groupID}/instances`,
+                           search:`duration=${updateProgressChartDuration.queryValue}`
```
joaquimrocha: Add a space between `:` and `

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -90,7 +90,7 @@ function ItemExtended(props) {
   function updateStatusChart(duration){
     setStatusChartDuration(duration);
   }
-  React.useEffect(()=>{
+  React.useEffect(() => {
```
joaquimrocha: This doesn't belong here.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -66,7 +66,7 @@ function ItemExtended(props) {
       setLoadingUpdateProgressChart(true);
       applicationsStore
         .getApplicationWithDurationFilter(props.appID, updateProgressChartDuration.queryValue)
-        .then((application=>{
+        .then((application => {
```
joaquimrocha: This doesn't belong here.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -526,7 +526,7 @@ export function StatusCountTimeline(props) {
       } else {
         setTimelineChartData({data:[], keys:[], colors:[]});
         applicationsStore
-          .getGroupStatusCountTimeline(group.application_id, group.id, 
+          .getGroupStatusCountTimeline(group.application_id, group.id,
```
joaquimrocha, line 529: This doesn't belong here.

frontend/src/js/components/Instances/List.js
```diff
@@ -226,9 +234,11 @@ function ListView(props) {
             alignItems="stretch"
           >
             <Grid item md>
+              {!instanceFetchLoading &&
```
joaquimrocha: This doesn't seem to belong here.

frontend/src/js/components/Instances/Item.react.js
```diff
@@ -67,7 +67,7 @@ function Item(props) {
   if (!_.isEmpty(props.lastVersionChannel)) {
     if (version === props.lastVersionChannel) {
       versionStyle = 'success';
-    } else if (semver.gt(version, props.lastVersionChannel)) {
+    } else if (semver.valid(version) && semver.gt(version, props.lastVersionChannel)) {
```
joaquimrocha, line 70: This seems to be the only change that's related to the commit's description. Please check whether the others belong to different patches.

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -0,0 +1,26 @@
+export default class GroupChartsStore{
+  constructor(){
+    this.VersionBreakdownChartData = {};
+    this.StatusBreakdownChartData = {};
```
joaquimrocha, line 4: Don't Capitalize these variables (i.e. they should start with lower-case).

cmd/nebraska/controller.go
ashu8912: done

pkg/api/groups.go
ashu8912: done

frontend/src/js/components/Groups/Charts.js
```diff
@@ -9,16 +9,18 @@ import { Area, AreaChart, CartesianGrid, Tooltip, XAxis, YAxis } from 'recharts'
...
 import { cleanSemverVersion, getInstanceStatus, getMinuteDifference, makeColorsForVersions, makeLocaleTime, useGroupVersionBreakdown } from '../../constants/helpers';
-import { applicationsStore } from '../../stores/Stores';
+import { applicationsStore, groupChartStore } from '../../stores/Stores';
 import Loader from '../Common/Loader';
 import SimpleTable from '../Common/SimpleTable';
 import makeStatusDefs from '../Instances/StatusDefs';
 
 function TimelineChart(props) {
-  const {width = 500, height = 400, interpolation = 'monotone'} = props;
+  const { width = 500, height = 400, interpolation = 'monotone' } = props;
```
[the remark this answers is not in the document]
ashu8912: @joaquimrocha as the new eslint rules were added in the master so I ran lint over each commit and this is the result of that I did this because I wanted every commit to be in sync with current master

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+  setStatusChartData(key, data){
+    if (key === '1h' || this.statusBreakdownChartData.hasOwnProperty(key)){
+      return;
+    }
+    this.statusBreakdownChartData[key] = data;
+  }
+  getStatusChartData(key){
+    return this.statusBreakdownChartData[key];
+  }
```
joaquimrocha, line 24: Add a space between `)` and `{` for methods declarations.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -462,9 +473,36 @@ export function StatusCountTimeline(props) {
 
   // Make the timeline data again when needed.
   React.useEffect(() => {
+    function getStatusTimeline(group) {
+      const statusChartData = groupChartStore.getStatusChartData(duration);
```
joaquimrocha, line 494: I realized now that the way this store is actually a problem because it's assuming there will only be one group. So if you change to another group, the data will still be there. It needs to store the data per group.

I am also thinking that, since you're creating a store, we could move the logic for fetching the new data also into it, i.e. the caller never knows if the data is cached or needs to be refetched because that'd be the store's responsibility.
This means that the data getter should be passed a callback instead of returning the data directly; and another advantage is that we don't have to update the store from the caller (i.e. we don't need a `setter` in the store).

pkg/api/groups.go
joaquimrocha: Do we need these changes now that our DB queries are faster?
(If we do, they should be added to the original patch because this is fixing something introduced by your PR)

pkg/api/groups.go
```diff
@@ -323,13 +325,12 @@ func (api *API) GetGroupInstancesStats(groupID string) (*InstancesStatusStats, e
...
 	WHERE group_id=$1 AND last_check_for_updates > now() at time zone 'utc' - interval '%s' AND %s`,
 		InstanceStatusError, InstanceStatusUpdateGranted, InstanceStatusComplete, InstanceStatusInstalled,
-		InstanceStatusDownloaded, InstanceStatusDownloading, InstanceStatusOnHold, validityInterval, ignoreFakeInstanceCondition("instance_id"))
+		InstanceStatusDownloaded, InstanceStatusDownloading, InstanceStatusOnHold, durationString, ignoreFakeInstanceCondition("instance_id"))
 
 	err := api.dbR.SQL(query, groupID).QueryStruct(&instancesStats)
 	if err != nil {
 		return nil, err
 	}
-
```
joaquimrocha, line 317: Doesn't belong here.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -31,19 +31,19 @@ function TimelineChart(props) {
...
-      }
-
-      return ticks;
-    }
     if (lengthMinutes === 7 * DAY) {
       tickCount = 7;
       dateFormat = {useDate:true, showTime:false};
     }
+    if (lengthMinutes === 60){
+      for (let i = 0;i < 4;i++){
```
joaquimrocha, line 39: Add space after `;`.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -31,19 +31,19 @@ function TimelineChart(props) {
...
+        const minuteValue = lengthMinutes / 4 * i;
+        startTs.setMinutes(new Date(props.data[0].timestamp).getMinutes() + minuteValue);
+        ticks[i] = makeLocaleTime(startTs, {useDate:false});
+
```
joaquimrocha, line 43: Remove this newline.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -266,13 +266,13 @@ export function VersionCountTimeline(props) {
...
-            });
-            makeChartData(group, versionCountTimeline || []);
-            setSelectedEntry(-1);
+              groupChartStore.setVersionChartData(duration, versionCountTimeline);
+              setTimeline({
+                timeline: versionCountTimeline,
+                lastUpdate: lastUpdate.toUTCString(),
+              });
+              makeChartData(group, versionCountTimeline || []);
+              setSelectedEntry(-1);
```
joaquimrocha: Is this needed? Doesn't seem to belong here.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -486,21 +487,24 @@ export function StatusCountTimeline(props) {
...
-            groupChartStore.setStatusChartData(duration, statusCountTimeline);
-            setTimeline({
-              timeline: statusCountTimeline,
-              lastUpdate: new Date().toUTCString(),
-            });
-
-            makeChartData(statusCountTimeline || []);
-            setSelectedEntry(-1);
+            groupChartStore.setStatusChartData(duration.queryValue, statusCountTimeline);
+            if (!canceled){
```
joaquimrocha, line 509: Add a space between `)` and `{`.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -266,20 +267,23 @@ export function VersionCountTimeline(props) {
...
-              setSelectedEntry(-1);
+            groupChartStore.setVersionChartData(duration, versionCountTimeline);
+            if (!canceled)
+            { setTimeline({
+              timeline: versionCountTimeline,
+              lastUpdate: lastUpdate.toUTCString(),
+            });
+            makeChartData(group, versionCountTimeline || []);
+            setSelectedEntry(-1);
+            }
```
joaquimrocha: Definitely doesn't belong here.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -473,6 +473,7 @@ export function StatusCountTimeline(props) {
 
   // Make the timeline data again when needed.
   React.useEffect(() => {
+    let canceled = false;
```
joaquimrocha, line 492: The cancellable logic here seems to belong to the next patch.

frontend/src/js/components/Groups/ItemExtended.react.js
joaquimrocha: Still needs this ^

frontend/src/js/components/Instances/List.js
joaquimrocha: Please remove the newline ^

frontend/src/js/components/Instances/List.js
joaquimrocha: Still doesn't seem to belong here.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -489,8 +501,10 @@ export function StatusCountTimeline(props) {
         setSelectedEntry(-1);
       } else {
         setTimelineChartData({data: [], keys: [], colors: []});
-        applicationsStore.getGroupStatusCountTimeline(group.application_id, group.id, duration)
-          .done(statusCountTimeline => {
+        applicationsStore
+          .getGroupStatusCountTimeline(group.application_id, group.id,
+           duration.queryValue)
+          .then(statusCountTimeline => {
```
joaquimrocha, line 507: The done/fail changes seem to belong to the patch that introduced them (in your PR).

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,21 +29,24 @@ const useStyles = makeStyles({
...
   const [group, setGroup] = React.useState(null);
   const [instancesStats, setInstancesStats] = React.useState({});
+  const [updateProgressChartDuration, setUpdateProgressChartDuration] =
+    React.useState(defaultTimeInterval);
   const [versionChartSelectedDuration, setVersionChartSelectedDuration] =
     React.useState(defaultTimeInterval);
   const [statusChartDuration, setStatusChartDuration] =
     React.useState(defaultTimeInterval);
   const classes = useStyles();
+
```
joaquimrocha: Doesn't belong here.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,21 +29,24 @@ const useStyles = makeStyles({
...
   const [statusChartDuration, setStatusChartDuration] =
     React.useState(defaultTimeInterval);
   const classes = useStyles();
+  
   function onChange() {
     const app = applicationsStore.getCachedApplication(props.appID);
 
     if (!app) {
       applicationsStore.getApplication(props.appID);
-      return;
```
joaquimrocha, line 38: Why do you remove this?

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -29,21 +29,24 @@ const useStyles = makeStyles({
...
     }
-
```
joaquimrocha, line 40: Doesn't belong here.

frontend/src/js/api/API.js
```diff
@@ -70,8 +70,8 @@ class API {
     return API.getJSON(`${BASE_URL}/apps/${applicationID}/groups/${groupID}/status_timeline?duration=${duration}`);
   }
 
-  static getGroupInstancesStats(applicationID, groupID) {
-    return API.getJSON(BASE_URL + '/apps/' + applicationID + '/groups/' + groupID + '/instances_stats');
+  static getGroupInstancesStats(applicationID, groupID, duration) {
+    return API.getJSON(BASE_URL + '/apps/' + applicationID + '/groups/' + groupID + '/instances_stats' + `?duration=${duration}`);
```
joaquimrocha, line 74: Same comment about replacing the whole line to be consistent.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -40,7 +39,7 @@ function ItemExtended(props) {
   const [statusChartDuration, setStatusChartDuration] =
     React.useState(defaultTimeInterval);
   const classes = useStyles();
-  
+
```
joaquimrocha, line 42: Doesn't belong here.

pkg/api/groups.go
ashu8912: @joaquimrocha but this interval and duration belong to this or only will it be fine to add this patch to the original patch which has nothing to do with durations

ashu8912: @joaquimrocha sorry i misunderstood you earlier 
these are actually the values we had earlier and the chart load time is decent with these old values as well. So i think we can stick with these values.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -29,15 +31,11 @@ function TimelineChart(props) {
...
-        ticks[index] = makeLocaleTime(props.data[index].timestamp, {useDate: false});
-      }
-
-      return ticks;
+    if (lengthMinutes === 7 * 24 * 60)
+    {
+      tickCount = 7;
+      dateFormat = {useDate:true, showTime:false};
     }
-
```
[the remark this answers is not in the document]
ashu8912: @joaquimrocha which line are you referring to I don't see a removal here

frontend/src/js/components/Groups/ItemExtended.react.js
ashu8912: @joaquimrocha same which line or is it fixed?

frontend/src/js/components/Groups/Charts.js
```diff
@@ -269,9 +252,38 @@ export function VersionCountTimeline(props) {
 
   // Make the timeline data again when needed.
   React.useEffect(() => {
+    async function getVersionTimeline(group) {
+      // Check if we should update the timeline or it's too early.
+      const lastUpdate = new Date(timeline.lastUpdate);
+      const cachedVersionChartData = groupChartStore.
```
joaquimrocha, line 266: Shouldn't the `.` be in the next line? I am surprised the linter let this pass.

frontend/src/js/components/Groups/ItemExtended.react.js
joaquimrocha: It's probably fixed.

frontend/src/js/components/Groups/ItemExtended.react.js
```diff
@@ -52,6 +58,12 @@ function ItemExtended(props) {
     props.handleUpdateGroup(props.groupId, props.appID);
   }
 
+  function updateVersionChart(duration) {
+    setVersionChartSelectedDuration(duration);
+  }
+  function updateStatusChart(duration) {
+    setStatusChartDuration(duration);
+  }
```
joaquimrocha: Why is this needed? Doesn't seem to do much. Where is it used?

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -0,0 +1,69 @@
...
+export default class GroupChartsStore{
+  constructor(){
+    this.versionBreakdownChartData = {};
+    this.statusBreakdownChartData = {};
+  }
+
+  async getGroupVersionCountTimeline(appID, groupID, duration) {
+    let versionCountTimeline;
+    try {
+      versionCountTimeline = await applicationsStore.
```
joaquimrocha, line 11: Add the `.` to the next line.

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -0,0 +1,69 @@
...
+      return;
+    }
+    if (!this.statusBreakdownChartData[groupID]) {
+      this.statusBreakdownChartData[groupID] = {};
+    }
+
+    this.statusBreakdownChartData[groupID][duration] = data;
+  }
+
+  getStatusChartData(groupID, key) {
```
joaquimrocha, line 63: I think `key` here means duration, right?

frontend/src/js/components/Groups/Charts.js
```diff
@@ -269,9 +252,38 @@ export function VersionCountTimeline(props) {
...
+          setTimeline({
+            timeline: versionCountTimeline,
+            lastUpdate: lastUpdate.toUTCString(),
+          });
+          makeChartData(group, versionCountTimeline || []);
+          setSelectedEntry(-1);
+        } catch (error) {
+          console.error(error);
+        }
+
```
joaquimrocha: You said you wanted to keep the setter from outside the groupStore but I don't understand why. What's the advantage/need to use this outside of the groupStore? I think it would be much better to keep it abstracted in the group (I listed the reasons before).

frontend/src/js/components/Groups/Charts.js
```diff
@@ -31,19 +31,18 @@ function TimelineChart(props) {
...
-    }
     if (lengthMinutes === 7 * DAY) {
       tickCount = 7;
       dateFormat = {useDate: true, showTime: false};
     }
+    if (lengthMinutes === 60) {
+      for (let i = 0; i < 4; i++){
+        const minuteValue = lengthMinutes / 4 * i;
+        startTs.setMinutes(new Date(props.data[0].timestamp).getMinutes() + minuteValue);
+        ticks[i] = makeLocaleTime(startTs, {useDate:false});
```
joaquimrocha, line 42: Missing a space after the `:`.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -268,19 +269,22 @@ export function VersionCountTimeline(props) {
         try {
           const versionCountTimeline = await groupChartStore.
             getGroupVersionCountTimeline(group.application_id, group.id, duration.queryValue);
-          setTimeline({
+          if (!canceled)
+          { setTimeline({
```
joaquimrocha, line 281: We've been using
```
if (!canceled) {
   ...
}
```

Please keep it consistent.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -69,7 +69,15 @@ function TimelineChart(props) {
 
         const tick = getMinuteDifference(tickDate, startTs) * dimension / lengthMinutes;
         // Show only the time.
-        ticks[tick] = makeLocaleTime(tickDate, dateFormat);
+        if (lengthMinutes === 30 * DAY){
+          if (tickDate.getMonth() - startTs.getMonth() === 1){
+            ticks[tick] = makeLocaleTime(tickDate, {showTime:false, dateFormat:{month:'short', day:'numeric'}});
```
joaquimrocha, line 74: Add a space after `:`.

frontend/src/js/components/Instances/List.js
```diff
@@ -96,12 +96,10 @@ function ListView(props) {
...
   const versionBreakdown = useGroupVersionBreakdown(group);
   const [rowsPerPage, setRowsPerPage] = React.useState(10);
-  const [totalInstances, setTotalInstances] = React.useState(-1);
-  const [instancesObj, setInstancesObj] = React.useState({instances: [], total: 0});
+  const [instancesObj, setInstancesObj] = React.useState({instances: [], total: -1});
   const [instanceFetchLoading, setInstanceFetchLoading] = React.useState(false);
   const [filters, setFilters] = React.useState({status: '', version: ''});
-  const [totalInstancesWithoutFilter, setTotalInstancesWithoutFilter] = React.useState(0);
-  
+  const [totalInstancesWithoutFilter, setTotalInstancesWithoutFilter] = React.useState(-1);
```
joaquimrocha, line 102: This and the surrounding changes seem to be related to the 2nd commit before this one.

frontend/src/js/components/Groups/ItemExtended.react.js
ashu8912: @joaquimrocha this is the function that is triggered when somebody clicks on some other duration

frontend/src/js/components/Instances/List.js
ashu8912: fixed

pkg/api/groups.go
```diff
@@ -8,6 +8,21 @@ import (
...
+type Duration int
+
+const (
+	OneHour Duration = iota
+	OneDay
+	SevenDays
+	ThirtyDays
+)
+
+var durationMap = map[string]Duration{
```
krnowak, line 20: This is a general comment, related to this line, but not specifically about this line.

There seems to be a bit of confusion in the code about the string duration parameters. We use the `string` type to pass a duration we got from the request and we use the same type for the passing the postgres timings too. Some functions expect to get a duration string, but receives postgres timing. Or other way around. This is because one type (`string`) expresses two or three different things (duration parameter, postgres duration, postgres interval), so we don't have any type safety here.

I'd like to suggest creating distinct types for duration parameter we get from the request (`durationParam`) and for postgres timings (`postgresDuration` and `postgresInterval`). The exported functions in this package that receive a duration string parameter should still take it as a `string` type, but in the code, they should cast it to `durationParam`. The unexported functions that receive a duration string should start taking either `durationParam` or `postgresDuration` or `postgresInterval`. That way we give the compiler an ability to report a type error to us, when we mix up the strings. Please see the code I wrote on go playground which does it: https://play.golang.org/p/MgZqLN0LfpL

pkg/api/instances.go
```diff
@@ -176,22 +176,27 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+			return InstancesWithTotal{}, err
+		}
 	}
-	if err != nil {
-		return InstancesWithTotal{}, err
+	if len(duration) <= 0 {
+		duration = "1d"
 	}
-	err = api.getFilterInstancesQuery("COUNT (*)", p).QueryStruct(&totalCount)
+	err = api.getFilterInstancesQuery("COUNT (*)", p, duration).QueryStruct(&totalCount)
```
krnowak, line 199: `getFilterInstancesQuery` puts the passed duration straight into the SQL query, so it expects a postgres duration string. And you pass a string we got from the request, which likely is not a valid postgres duration string.

pkg/api/instances.go
```diff
@@ -302,14 +307,13 @@ func (api *API) getFilterInstancesQuery(selectPart string, p InstancesQueryParam
 
 // instancesQuery returns a SelectDocBuilder prepared to return all instances
 // that match the criteria provided in InstancesQueryParams.
-func (api *API) instancesQuery(p InstancesQueryParams) *dat.SelectDocBuilder {
-	instancesSubquery := api.getFilterInstancesQuery("instance_id", p)
-
+func (api *API) instancesQuery(p InstancesQueryParams, duration string) *dat.SelectDocBuilder {
+	dbDuration, _, _ := durationToPostgresTimings(durationMap[duration])
```
krnowak, line 311: Silent conversion going on here.

pkg/api/instances.go
```diff
@@ -176,22 +176,27 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
 	var totalCount uint64
 	var err error
 	p.Page, p.PerPage = validatePaginationParams(p.Page, p.PerPage)
-	err = api.instancesQuery(p).Paginate(p.Page, p.PerPage).QueryStructs(&instances)
-	if err == sql.ErrNoRows {
-		return InstancesWithTotal{
-			TotalInstances: 0,
-			Instances:      nil,
-		}, nil
+	if len(duration) > 0 {
```
krnowak, line 184:
```suggestion
	if duration != "" {
```

krnowak: Also, if duration is an empty string, then we only get a total count of instances, but no information about the instances. Is it expected? Is it allowed for the duration string to be empty?

pkg/api/updates_test.go
```diff
@@ -202,7 +202,7 @@ func TestGetUpdatePackage_RolloutStats(t *testing.T) {
 
 	group, _ := a.GetGroup(tGroup.ID)
 	assert.True(t, group.RolloutInProgress)
-	stats, _ := a.GetGroupInstancesStats(group.ID)
+	stats, _ := a.GetGroupInstancesStats(group.ID, validityInterval)
```
krnowak, line 205: Passing postgres duration, while the function expects the "1d" format.

pkg/api/updates_test.go
```diff
@@ -211,7 +211,7 @@ func TestGetUpdatePackage_RolloutStats(t *testing.T) {
 
 	group, _ = a.GetGroup(tGroup.ID)
 	assert.True(t, group.RolloutInProgress)
-	stats, _ = a.GetGroupInstancesStats(group.ID)
+	stats, _ = a.GetGroupInstancesStats(group.ID, validityInterval)
```
krnowak, line 214: Same.

pkg/api/updates_test.go
```diff
@@ -221,7 +221,7 @@ func TestGetUpdatePackage_RolloutStats(t *testing.T) {
...
```
krnowak, line 224: Same.

pkg/api/updates_test.go
```diff
@@ -230,7 +230,7 @@ func TestGetUpdatePackage_RolloutStats(t *testing.T) {
...
```
krnowak, line 233: Same.

pkg/api/updates_test.go
```diff
@@ -239,7 +239,7 @@ func TestGetUpdatePackage_RolloutStats(t *testing.T) {
 
 	group, _ = a.GetGroup(tGroup.ID)
 	assert.False(t, group.RolloutInProgress)
-	stats, _ = a.GetGroupInstancesStats(group.ID)
+	stats, _ = a.GetGroupInstancesStats(group.ID, validityInterval)
```
krnowak, line 242: Same.

frontend/src/js/constants/timeInterval.js
```diff
@@ -0,0 +1,10 @@
+const timeIntervals = [
+  {displayValue: '30 days', queryValue: '30d'},
+  {displayValue: '7 days', queryValue: '7d'},
+  {displayValue: '1 day', queryValue:'1d'},
+  {displayValue:'1 hour', queryValue:'1h'}
+];
+
+export const defaultTimeInterval = timeIntervals
+  .find((interval) => interval.queryValue === '1d');
+export default timeIntervals;
```
joaquimrocha, line 10: I still think this would be better in `constants/helpers.js`

frontend/src/js/components/Groups/ItemExtended.react.js
joaquimrocha: I meant that this function is just calling another function. So you could even pass an arrow function `(duration) => setStatusChartDuration(duration)`.
It's a nitpick though. You don't have to change it.

frontend/src/js/components/Instances/Item.react.js
```diff
@@ -77,8 +77,9 @@ function Item(props) {
         versionStyle = 'danger';
     }
   }
-
-  const instancePath = `/apps/${appID}/groups/${groupID}/instances/${instanceID}`;
+  const duration = new URLSearchParams(window.location.search);
+  console.log(duration.toString());
```
joaquimrocha, line 81: Remove this log.

ashu8912:
Thanks @joaquimrocha  and @krnowak for the reviews i have considered all the changes mentioned

frontend/src/js/constants/helpers.js
```diff
@@ -16,6 +16,15 @@ export const ARCHES = {
 };
 
 const colors = makeColors();
+export const timeIntervals = [
+  {displayValue: '30 days', queryValue: '30d'},
+  {displayValue: '7 days', queryValue: '7d'},
+  {displayValue: '1 day', queryValue:'1d'},
+  {displayValue:'1 hour', queryValue:'1h'}
```
joaquimrocha, line 23: Missing a space after some colons.

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -0,0 +1,75 @@
+import { applicationsStore } from './Stores';
+export default class GroupChartsStore{
+  constructor(){
```
joaquimrocha, line 3: Missing space between `)` and `{`.

frontend/src/js/stores/GroupChartsStore.js
```diff
@@ -0,0 +1,75 @@
...
+
+  async getGroupVersionCountTimeline(appID, groupID, duration) {
+    let versionCountTimeline;
+    versionCountTimeline = this.getVersionChartData(groupID, duration);
+    if (!versionCountTimeline) {
+      try {
+        versionCountTimeline = await applicationsStore.
+          getGroupVersionCountTimeline(appID, groupID, duration);
+        this.setVersionChartData(groupID, duration, versionCountTimeline);
+      } catch (error){
```
joaquimrocha, line 16: Missing space between `)` and `{`.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -69,7 +69,15 @@ function TimelineChart(props) {
 
         const tick = getMinuteDifference(tickDate, startTs) * dimension / lengthMinutes;
         // Show only the time.
-        ticks[tick] = makeLocaleTime(tickDate, dateFormat);
+        if (lengthMinutes === 30 * DAY){
```
joaquimrocha, line 72: Missing space between `)` and `{`.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -69,7 +69,15 @@ function TimelineChart(props) {
...
+          if (tickDate.getMonth() - startTs.getMonth() === 1){
```
joaquimrocha, line 73: Missing space between `)` and `{`.

frontend/src/js/components/Groups/Charts.js
```diff
@@ -31,19 +31,18 @@ function TimelineChart(props) {
...
-      }
-
-      return ticks;
-    }
     if (lengthMinutes === 7 * DAY) {
       tickCount = 7;
       dateFormat = {useDate: true, showTime: false};
     }
+    if (lengthMinutes === 60) {
+      for (let i = 0; i < 4; i++){
```
joaquimrocha, line 39: Missing space between `)` and `{`.

frontend/src/js/components/Layouts/InstanceLayout.js
```diff
@@ -66,7 +67,7 @@ export default function InstanceLayout(props) {
             label: groupName
           },
           {
-            path: `/apps/${appID}/groups/${groupID}/instances`,
+            path: `/apps/${appID}/groups/${groupID}/instances?${duration}`,
```
joaquimrocha, line 70: Is this path correct? didn't you forget the parameter name?

pkg/api/instances.go
```diff
@@ -176,22 +176,38 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
-func (api *API) GetInstances(p InstancesQueryParams) (InstancesWithTotal, error) {
+func (api *API) GetInstances(p InstancesQueryParams, duration string) (InstancesWithTotal, error) {
 	var instances []*Instance
 	var totalCount uint64
 	var err error
+
 	p.Page, p.PerPage = validatePaginationParams(p.Page, p.PerPage)
-	err = api.instancesQuery(p).Paginate(p.Page, p.PerPage).QueryStructs(&instances)
-	if err == sql.ErrNoRows {
+	if duration == "" {
```
krnowak, line 185: There seems to be a change in how the function works now, right? Empty duration now means "retrieve how many instances were updates in some default duration (validityInterval), but do not get the details of the updated instances"?. Is this a useful thing to have? Does the frontend send requests for getting instances without the duration parameter?

pkg/api/instances.go
```diff
@@ -154,7 +154,7 @@ func (api *API) GetInstance(instanceID, appID string) (*Instance, error) {
 	err := api.dbR.
 		SelectDoc("*").
 		From("instance").
-		One("application", api.instanceAppQuery(appID)).
+		One("application", api.instanceAppQuery(appID, validityInterval)).
```
krnowak, line 157: Could you check if defining `validityInterval` as a `postgresDuration` type would work?

krnowak:
Also, there's a conflict to resolve.

pkg/api/instances.go
ashu8912: Yeah @krnowak this is a required change because in application page view on frontend we need the total no of instances for an application without needing the instances. coz fetching instances also in that view is useless and makes page slow to load

krnowak: Hm, I think that this should likely become a separate entry in API (like `GetInstancesTotalCount` or something), but certainly not in this PR. Thanks for the explanation.

pkg/api/instances.go
ashu8912: yeah it works

krnowak: Cool.

pkg/api/instances.go
```diff
@@ -202,6 +214,31 @@ func (api *API) GetInstances(p InstancesQueryParams) (InstancesWithTotal, error)
 	return result, nil
 }
 
+func (api *API) GetInstancesCount(p InstancesQueryParams, duration string) (InstancesWithTotal, error) {
```
krnowak, line 217: Hm, do we want to have a separate type for reporting the instances count? The `instances` field on `InstancesWithTotal` is always `nil` in this function,.

pkg/api/instances.go
joaquimrocha: Yeah, let's keep that change for later.

krnowak: Too late.

pkg/api/instances.go
ashu8912: fixed with specifying the return type to be just uint64

pkg/api/instances.go
ashu8912: ;-;

PR #168, opened 2020-04-01
Improve linter

joaquimrocha:
The linter was missing a few rules. This new file should be also shared by the dashboard.


joaquimrocha:
(The lint fails because we changed it)

ashu8912:
ok Joaquim the eslint rules look fine to me

PR #173, opened 2020-04-07
Fix pagination and filter query

ashu8912:


pkg/api/instances.go
```diff
@@ -171,12 +178,13 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
 }
 
 // GetInstances returns all instances that match with the provided criteria.
-func (api *API) GetInstances(p InstancesQueryParams) ([]*Instance, error) {
+func (api *API) GetInstances(p InstancesQueryParams) (InstanceWithFilter, error, error) {
```
joaquimrocha, line 181: Hmmm, the function should only return one error. Think of errors here as exceptions, once there is one which is not recoverable, you just return it. So the error returned in this case will either be for getting the count, or getting the instances, but we don't need two return params for that.

pkg/api/instances.go
```diff
@@ -171,12 +178,13 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
 	var instances []*Instance
-
-	err := api.instancesQuery(p).QueryStructs(&instances)
-
-	return instances, err
+	var totalCount uint64
+	err := api.instancesQuery(p).Paginate(p.Page, p.PerPage).QueryStructs(&instances)
```
joaquimrocha, line 184: If there is an error here. You can just early return (no purpose in getting an error for the instances but still getting their count afterwards).

pkg/api/instances.go
```diff
@@ -171,12 +178,13 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	countErr := api.getFilterInstanceCount(p).QueryStruct(&totalCount)
```
joaquimrocha, line 185: You should reuse the same `err` variable as this line should only occur (after you address my last comment) if it didn't early return. To do that you probably have to declare `var err error` above, like you declared the `instances` and `totalCount`.

pkg/api/instances.go
```diff
@@ -171,12 +178,13 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	result := InstanceWithFilter{TotalInstances: totalCount, Instances: instances}
```
joaquimrocha, line 186: nitpick: AFAIU ,you can omit the names for the members here, if you are assigning them in the right order (and it's only two now, so it's easy to get right).

pkg/api/instances.go
```diff
@@ -260,6 +268,22 @@ func (api *API) instanceAppQuery(appID string) *dat.SelectDocBuilder {
...
+		Where("application_id = $1 AND group_id = $2", p.ApplicationID, p.GroupID).
+		Where(fmt.Sprintf("last_check_for_updates > now() at time zone 'utc' - interval '%s'", validityInterval)).
+		Where(ignoreFakeInstanceCondition("instance_id"))
+
+	if p.Status != 0 {
+		instancesSubquery.Where("status = $1", p.Status)
+	}
+	if p.Version != "" {
+		instancesSubquery.Where("version = $1", p.Version)
+	}
```
joaquimrocha, line 284: Since this is the same query as getting the instances, I think you can just use the previous `SelectBuilder` (let's call it `instancesSelect`) when getting the instances before you paginate them. And you just use `Select("COUNT (*)").instancesSelect`.

This should be possible. If it's not (:-1:) then it should be abstracted into a helper function so we don't repeat the same code in two places (which likely results in maintenance problems down the road).

pkg/api/instances.go
```diff
@@ -260,6 +268,22 @@ func (api *API) instanceAppQuery(appID string) *dat.SelectDocBuilder {
...
+	return instancesSubquery
```
joaquimrocha, line 285: Once you figure out how to reuse the `getInstances` return, this whole function shouldn't be needed. But in case you keep it for some reason, maybe you can return the integer value already instead of requiring one extra step from the caller.

pkg/api/instances_test.go
```diff
@@ -161,15 +161,15 @@ func TestGetInstancesFiltered(t *testing.T) {
...
 	assert.NoError(t, err)
 	expectedIDs := map[string]struct{}{
 		instanceID2: {},
 		instanceID3: {},
 		instanceID4: {},
 	}
-	if assert.Equal(t, 3, len(instances)) {
-		for _, instance := range instances {
+	if assert.Equal(t, 3, len(result.Instances)) {
+		for _, instance := range result.Instances {
```
joaquimrocha, line 172: (comment for the whole section)

It would be very nice if you could add a test to check the pagination and the total.

pkg/api/instances.go
krnowak: This can return `sql.ErrNoRows` error, which likely should be just handled by setting the `Instances` member to nil. Otherwise yeah, if other error happens, return early. Also, no need for two `error` return values.

pkg/api/instances.go
krnowak: I'd prefer:
```go
return InstanceWithFilter{
	TotalInstances: totalCount,
	Instances: instances,
}, nil
```

pkg/api/instances.go
```diff
@@ -260,6 +268,22 @@ func (api *API) instanceAppQuery(appID string) *dat.SelectDocBuilder {
 func ignoreFakeInstanceCondition(instanceIDField string) string {
 	return fmt.Sprintf(`(%[1]s IS NULL OR %[1]s NOT SIMILAR TO '\{[a-fA-F0-9-]{36}\}')`, instanceIDField)
 }
+func (api *API) getFilterInstanceCount(p InstancesQueryParams) *dat.SelectBuilder {
+	instancesSubquery := api.dbR.
+		Select("COUNT (*)").
+		From("instance_application").
+		Where("application_id = $1 AND group_id = $2", p.ApplicationID, p.GroupID).
+		Where(fmt.Sprintf("last_check_for_updates > now() at time zone 'utc' - interval '%s'", validityInterval)).
```
krnowak, line 276:
```suggestion
		Where("last_check_for_updates > now() at time zone 'utc' - interval $1", validityInterval).
```

pkg/api/instances.go
krnowak: You can't do `Select("COUNT (*)").instanceSelect`. But you could have a helper like:
```go
func (api *API) someHelper(selectPart string) *dat.SelectBuilder {
	return api.dbR.
		Select(selectPart).
		From("instance_application").
		…
}
```

So in the `instancesQuery` function you could call `someHelper("instance_id")` and in `getFilterInstanceCount` - `someHelper("COUNT (*)")`.

pkg/api/instances.go
```diff
@@ -171,12 +176,22 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
 }
 
 // GetInstances returns all instances that match with the provided criteria.
-func (api *API) GetInstances(p InstancesQueryParams) ([]*Instance, error) {
+func (api *API) GetInstances(p InstancesQueryParams) (InstanceWithTotal, error) {
```
joaquimrocha, line 179: Maybe we should return a pointer here. @krnowak what do you think?

pkg/api/instances.go
joaquimrocha: @ashu8912 You should accept Krzesimir's suggestion and squash it with your commit.

pkg/api/instances.go
```diff
@@ -261,18 +276,32 @@ func ignoreFakeInstanceCondition(instanceIDField string) string {
 	return fmt.Sprintf(`(%[1]s IS NULL OR %[1]s NOT SIMILAR TO '\{[a-fA-F0-9-]{36}\}')`, instanceIDField)
 }
 
+func (api *API) getFilterInstancesQuery(selectPart string, appID string, groupID string) *dat.SelectBuilder {
```
joaquimrocha, line 279: If you take the `p InstancesQueryParams` here, you skip having one arg per used parameter, and you can do the filtering here instead of repeating the filtering logic in the two places that call this function.

cmd/nebraska/controller.go
```diff
@@ -636,14 +636,12 @@ func (ctl *controller) getInstances(c *gin.Context) {
 	p.Page, _ = strconv.ParseUint(c.Query("page"), 10, 64)
 	p.PerPage, _ = strconv.ParseUint(c.Query("perpage"), 10, 64)
 
-	instances, err := ctl.api.GetInstances(p)
+	result, err := ctl.api.GetInstances(p)
 	switch err {
-	case nil:
-		if err := json.NewEncoder(c.Writer).Encode(instances); err != nil {
+	case nil, sql.ErrNoRows:
```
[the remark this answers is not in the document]
krnowak: This matches if `err` is either `nil` or is `sql.ErrNoRows`. The select functions tend to return `sql.ErrNoRows` if select returned zero rows.

Maybe I misunderstood your question.

pkg/api/instances.go
krnowak: Doesn't matter that much - the copy of the returned value is shallow, so it's just 16 bytes.

pkg/api/instances.go
```diff
@@ -171,12 +176,22 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
-	return instances, err
+	var totalCount uint64
+	err := api.instancesQuery(p).Paginate(p.Page, p.PerPage).QueryStructs(&instances)
+	if err != nil && err != sql.ErrNoRows {
+		return InstanceWithTotal{}, err
+	}
+	countErr := api.getFilterInstanceCount(p).QueryStruct(&totalCount)
+	if countErr != nil {
+		return InstanceWithTotal{}, countErr
+	}
```
krnowak, line 189:
```suggestion
	err = api.getFilterInstanceCount(p).QueryStruct(&totalCount)
	if err != nil {
		return InstanceWithTotal{}, err
	}
```

pkg/api/instances.go
```diff
@@ -171,12 +176,22 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	result := InstanceWithTotal{
+		TotalInstances: totalCount,
+		Instances:      instances,
+	}
+	return result, err
```
krnowak, line 194:
```suggestion
	return result, nil
```

cmd/nebraska/controller.go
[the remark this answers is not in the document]
krnowak: Also, probably would be nicer if the `GetInstances` function didn't return `sql.ErrNoRows` at all. Then this code could become:

```go
result, err := ctl.api.GetInstances(p)
if err == nil {
	if err := json.NewEncoder(c.Writer).Encode(result); err != nil {
	…
} else {
	logger.Error("getInstances - getting instances", "error", err.Error(), "params", p)
	…
}
```

pkg/api/instances.go
krnowak: Oh, yeah, that was my mistake in my suggestion about the `someHelper` function - the helper should certainly have `selectPart string, p InstanceQueryParams` parameters and do this `if p.Status != 0` and `if p.Version != ""` stuff too. Sorry about that.

ashu8912:
@krnowak and @joaquimrocha I have made the changes please review.

pkg/api/instances_test.go
```diff
@@ -107,28 +107,29 @@ func TestGetInstances(t *testing.T) {
...
 
 	_, _ = a.GetUpdatePackage(tInstance.ID, "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
 	_ = a.RegisterEvent(tInstance.ID, tApp.ID, tGroup.ID, EventUpdateComplete, ResultSuccessReboot, "", "")
 
-	instances, err = a.GetInstances(InstancesQueryParams{ApplicationID: tApp.ID, GroupID: tGroup.ID, Status: InstanceStatusComplete, Page: 1, PerPage: 10})
+	result, err = a.GetInstances(InstancesQueryParams{ApplicationID: tApp.ID, GroupID: tGroup.ID, Status: InstanceStatusComplete, Page: 1, PerPage: 10})
 	assert.NoError(t, err)
-	assert.Equal(t, 1, len(instances))
+	assert.Equal(t, 1, len(result.Instances))
+	assert.Equal(t, (int)(result.TotalInstances), len(result.Instances))
```
joaquimrocha, line 132: This test is not 100% correct. It should test whether the result returned is <= the perPage paramter.
And you should include more pagination tests with maximumized/minimized values, i.e. page = 1000 , perPage = 0, perPage = 1 (assuming we have more instances than 1), etc.

frontend/src/js/api/API.js
```diff
@@ -161,9 +160,12 @@ class API {
 
   static getJSON(url) {
     PubSub.publish(MAIN_PROGRESS_BAR, 'add');
-
     return fetch(url)
-      .then((response) => response.json())
+      .then((response) => {
+        if (!response.ok){
```
joaquimrocha, line 165: Add a space between `) {`. :thinking: why hasn't the linter caught this?

frontend/src/js/api/API.js
```diff
@@ -161,9 +160,12 @@ class API {
 
   static getJSON(url) {
     PubSub.publish(MAIN_PROGRESS_BAR, 'add');
-
```
joaquimrocha, line 164: This is not part of this commit.

frontend/src/js/api/API.js
```diff
@@ -122,7 +122,6 @@ class API {
     if (!_.isEmpty(queryOptions)) {
       url += '?' + queryString.stringify(queryOptions);
     }
-
```
joaquimrocha, line 125: This is not part of this commit.
Also please realize that you have added a period to the commit's title (remove it please).

frontend/src/js/api/API.js
```diff
@@ -183,7 +185,12 @@ class API {
       };
     }
     return fetch(url, fetchConfigObject)
-      .then((response) => response.json())
+      .then((response) => {
+        if (!response.ok){
```
joaquimrocha, line 189: Same comment as before.

frontend/src/js/components/Instances/StatusDefs.js
```diff
@@ -13,41 +13,49 @@ function makeStatusDefs(theme) {
       label: 'Complete',
       color: theme.palette.success.main,
       icon: clipboardCheck,
+      queryValue: '4'
```
joaquimrocha, line 16: (Commenting here since github doesn't let me comment on the commits' message bodies)
Please rewite the message carefully as it is a bit confused and has capitalized `Thus` after a comma.
The title should probably also mention that the status defs belong to the instance, otherwise it's a bit generic.

frontend/src/js/components/Instances/List.js
```diff
@@ -24,16 +24,15 @@ const CHECKS_TIMEOUT = 60; // secs
 
 function InstanceFilter(props) {
   const statusDefs = makeStatusDefs(useTheme());
-  const {onFiltersChanged, versions} = props;
+  const { onFiltersChanged, versions } = props;
```
joaquimrocha, line 27: This fix doesn't belong to the commit and is not necessary.

frontend/src/js/components/Instances/List.js
```diff
@@ -24,16 +24,15 @@ const CHECKS_TIMEOUT = 60; // secs
...
 
   function changeFilter(filterName, filterValue) {
     if (filterValue === props.filter[filterName]) {
       return;
     }
 
-    const filter = props.filter;
+    const filter = {...props.filter};
     filter[filterName] = filterValue;
-
```
joaquimrocha, line 36: This doesn't belong to the commit.

frontend/src/js/components/Instances/List.js
```diff
@@ -46,7 +45,7 @@ function InstanceFilter(props) {
         >
           <InputLabel htmlFor="select-status" shrink>Filter Status</InputLabel>
           <Select
-            onChange={event => changeFilter('status', event.target.value) }
+            onChange={event => changeFilter('status', event.target.value)}
```
joaquimrocha, line 48: Another change that doesn't belong to the commit.

frontend/src/js/components/Instances/List.js
```diff
@@ -71,7 +70,7 @@ function InstanceFilter(props) {
         >
           <InputLabel htmlFor="select-versions" shrink>Filter Version</InputLabel>
           <Select
-            onChange={event => changeFilter('version', event.target.value) }
+            onChange={event => changeFilter('version', event.target.value)}
```
joaquimrocha, line 73: Ditto.

frontend/src/js/components/Instances/List.js
```diff
@@ -81,7 +80,7 @@ function InstanceFilter(props) {
           >
             <MenuItem key="" value="">Show All</MenuItem>
             {
-              (versions || []).map(({version}) => {
+              (versions || []).map(({ version }) => {
```
joaquimrocha, line 83: Ditto.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
+  const { application, group } = props;
   const [page, setPage] = React.useState(0);
   const [rowsPerPage, setRowsPerPage] = React.useState(10);
   const [instances, setInstances] = React.useState(null);
-  const [filteredInstances, setFilteredInstances] = React.useState([]);
+  const [instanceFetchLoading, setInstanceFetchLoading] = React.useState(false);
+  const [totalInstances, setTotalInstances] = React.useState(0);
   const [lastCheck, setLastCheck] = React.useState(new Date(0, 0)); // Long long time ago.
-  const [filters, setFilters] = React.useState({status: '', version: ''});
+  const [filters, setFilters] = React.useState({ status: '', version: '' });
```
joaquimrocha, line 103: Doesn't belong here.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
 }
 
 function ListView(props) {
-  const {application, group} = props;
+  const statusDefs = makeStatusDefs(useTheme());
+  const { application, group } = props;
```
joaquimrocha, line 96: Doesn't belong here.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
+  const [totalInstances, setTotalInstances] = React.useState(0);
   const [lastCheck, setLastCheck] = React.useState(new Date(0, 0)); // Long long time ago.
-  const [filters, setFilters] = React.useState({status: '', version: ''});
+  const [filters, setFilters] = React.useState({ status: '', version: '' });
 
-  function handleChangePage(event, newPage) {
+  function fetchInstances(filters, page, perpage){
+    setInstanceFetchLoading(true);
+    const fetchFilters = {...filters};
+    if (filters.status === ''){
```
joaquimrocha, line 108: Add a space between `)` and `{`.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
   const [instances, setInstances] = React.useState(null);
-  const [filteredInstances, setFilteredInstances] = React.useState([]);
+  const [instanceFetchLoading, setInstanceFetchLoading] = React.useState(false);
+  const [totalInstances, setTotalInstances] = React.useState(0);
   const [lastCheck, setLastCheck] = React.useState(new Date(0, 0)); // Long long time ago.
-  const [filters, setFilters] = React.useState({status: '', version: ''});
+  const [filters, setFilters] = React.useState({ status: '', version: '' });
 
-  function handleChangePage(event, newPage) {
+  function fetchInstances(filters, page, perpage){
```
joaquimrocha, line 105: It'd be nice if you used `perPage` here.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
+      }).then((result) => {
+      setInstanceFetchLoading(false);
+      const massagedInstances = result.instances.map((instance) => {
+        instance.statusInfo = instancesStore.getInstanceStatus(instance.application.status,
+            instance.application.version);
+        return instance;
+      });
+      setInstances(massagedInstances);
+      setTotalInstances(result.total_instance);
+      instancesStore.setInstances(result.instances);
```
joaquimrocha, line 128: Why are you setting the instances to the store here? This is not how the store is supposed to be used AFAIU.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
+    await fetchInstances(filters, newPage, rowsPerPage);
     setPage(newPage);
   }
 
-  function handleChangeRowsPerPage(event) {
+  async function handleChangeRowsPerPage(event) {
+    await fetchInstances(filters, 0, +event.target.value);
     setRowsPerPage(+event.target.value);
-    setPage(0);
+    setPage(1);
```
joaquimrocha, line 142: Why do we go to the 2nd page?

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
 
-  function applyFilters(_filters = {}) {
-    const newFilters = _filters || {status: '', version: ''};
+  async function applyFilters(_filters = {}) {
+    let newFilters;
+    if (Object.keys(_filters).length !== 0) {
+      newFilters = _filters;
+    } else {
+      newFilters = { status: '', version: '' };
+    }
```
joaquimrocha, line 155: You can do:
`const newFilters = Object.keys(_filters).length !== 0 ? _filters : {status: '', version: ''};`

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
-        });
+      fetchInstances(filters, page, rowsPerPage);
     }
-
-    return function cleanup() {
-      instancesStore.removeChangeListener(onChangeInstances);
-    };
   },
-  [lastCheck, instances, filteredInstances]);
+    [lastCheck, instances]);
```
joaquimrocha, line 173: I am not sure the `lastCheck` makes sense now as it is a single timestamp but now we fetch a "set" of instances (so the timestamp has no longer to do with setting the whole instances).

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
-    if (!instances || instances.length === 0)
-      return group.instances_stats.total;
-    if (filteredInstances.length === instances.length)
-      return filteredInstances.length;
-    return `${filteredInstances.length}/${instances.length}`;
+    if (totalInstances === 0){
+      return 0;
+    }
+    if (totalInstances < rowsPerPage || (rowsPerPage) * (page + 1) > totalInstances){
+      return `${totalInstances}/${totalInstances}`;
```
joaquimrocha, line 180: The idea of this count is to tell the user how many instances are being filtered out of the total. i.e. if there are 10 instances with version 0.0.1 out of a total of 1000, then it should show 10/1000.
But your totalInstances here is the total number of filtered instances (10 in the example above).

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
+    }
+    return `${(rowsPerPage) * (page + 1)}/${totalInstances}`;
```
joaquimrocha, line 182: I don't think you will get the right count with this.
If you are on page 1 of 10 pages (full), and you have 100 instances filtered, out of a total unfiltered of 1000, then you should get 100/1000, but this will give you 20/1000.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,103 +92,101 @@ function InstanceFilter(props) {
...
   }
 
   function isFiltered() {
     return filters.status || filters.version;
   }
-
   return (
+
```
joaquimrocha, line 189: This doesn't belong here.

frontend/src/js/stores/InstancesStore.js
```diff
@@ -10,51 +10,20 @@ class InstancesStore extends Store {
...
-          application[groupID] = [];
-          this.emitChange();
-        }
-      });
+  }
+  setInstances(instances){
+    this.instances = instances;
+  }
+  getCachedInstances(){
+    return this.instances;
```
joaquimrocha, line 22: This is no longer a store in the sense of caching things as it was.
We need to rethink this approach, or at least the implications of this in code that has been tested before.

ashu8912:
@krnowak I have added code to the backend please review

pkg/api/instances.go
```diff
@@ -171,12 +176,24 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
-
-	return instances, err
+	var totalCount uint64
+	var err error
+	p.Page, p.PerPage = validatePaginationParams(p.Page, p.PerPage)
+	err = api.instancesQuery(p).Paginate(p.Page, p.PerPage).QueryStructs(&instances)
+	if err != nil && err != sql.ErrNoRows {
+		return InstanceWithTotal{}, err
+	}
+	err = api.getFilterInstancesQuery("COUNT (*)", p).QueryStruct(&totalCount)
```
joaquimrocha, line 186: If the previous `err` was `ErrNoRows` then we can skip this query.

pkg/api/instances.go
```diff
@@ -54,6 +55,10 @@ type Instance struct {
 	CreatedTs   time.Time           `db:"created_ts" json:"created_ts"`
 	Application InstanceApplication `db:"application" json:"application,omitempty"`
 }
+type InstanceWithTotal struct {
+	TotalInstances uint64      `json:"total_instance"`
```
joaquimrocha, line 59: Please assign the JSON key of `total`.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,19 +90,51 @@ function InstanceFilter(props) {
...
+      const massagedInstances = result.instances.map((instance) => {
+        instance.statusInfo = instancesStore.getInstanceStatus(instance.application.status,
+            instance.application.version);
+        return instance;
+      });
+      setInstances(massagedInstances);
+      setTotalInstances(result.total_instance);
+      instancesStore.setInstances(result.instances);
+    })
+      .catch(()=>{
```
joaquimrocha, line 133: Add spaces around `=>` please.

frontend/src/js/components/Instances/List.js
```diff
@@ -114,75 +143,25 @@ function ListView(props) {
...
-  [lastCheck, instances, filteredInstances]);
+  []);
 
   function getInstanceCount() {
-    if (!instances || instances.length === 0)
-      return group.instances_stats.total;
-    if (filteredInstances.length === instances.length)
-      return filteredInstances.length;
-    return `${filteredInstances.length}/${instances.length}`;
+    return `${totalInstances}/${group.instances_stats.total}`;
```
joaquimrocha, line 170: I would like this label to be just the `group.instances_status.total` when no filter is applied. It was more or less what the previous code did.

frontend/src/js/components/Instances/List.js
```diff
@@ -233,21 +211,18 @@ function ListView(props) {
             </Grid>
           }
           <Grid item md={12}>
-            { instances ?
-              (filteredInstances.length > 0 ?
+            {!instanceFetchLoading ?
+              (instances && instances.length > 0 ?
```
joaquimrocha, line 221: nitpick: if you now have `instanceFetchLoading`, then you can assign the default value of `[]` to the `instances` status. Which avoids having to test for its validity.

frontend/src/js/components/Instances/List.js
```diff
@@ -93,19 +90,51 @@ function InstanceFilter(props) {
...
+        setInstanceFetchLoading(false);
+      });
+  }
+  async function handleChangePage(event, newPage) {
+    await fetchInstances(filters, newPage, rowsPerPage);
     setPage(newPage);
   }
 
-  function handleChangeRowsPerPage(event) {
+  async function handleChangeRowsPerPage(event) {
```
joaquimrocha, line 142: You are repeating the logic we do in the `useEffect` twice here.
I think a more organized way of handling the fetch would be to just change the `page` / `rowsPerPage`, and have then trigger the `useEffect` (adding the mentioned statuses as dependencies for the `useEffect`). i.e. changing the page or the number of rows would trigger a rebuild + fetch but just in one place.

cmd/nebraska/controller.go
```diff
@@ -608,7 +608,6 @@ func (ctl *controller) getInstanceStatusHistory(c *gin.Context) {
 	groupID := c.Params.ByName("group_id")
 	instanceID := c.Params.ByName("instance_id")
 	limit, _ := strconv.ParseUint(c.Query("limit"), 10, 64)
-
```
joaquimrocha, line 611: This doesn't belong here.

pkg/api/instances.go
```diff
@@ -148,16 +148,14 @@ func (api *API) RegisterInstance(instanceID, instanceIP, instanceVersion, appID,
 }
 
 // GetInstance returns the instance identified by the id provided.
-func (api *API) GetInstance(instanceID, appID string) (*Instance, error) {
+func (api *API) GetInstance(instanceID string, appID string) (*Instance, error) {
 	var instance Instance
-
```
joaquimrocha, line 148: Does not belong here.

pkg/api/instances.go
```diff
@@ -148,16 +148,14 @@ func (api *API) RegisterInstance(instanceID, instanceIP, instanceVersion, appID,
...
 	err := api.dbR.
 		SelectDoc("*").
 		From("instance").
 		One("application", api.instanceAppQuery(appID)).
 		Where("id = $1", instanceID).
 		QueryStruct(&instance)
-
```
joaquimrocha, line 155: Does not belong here.

frontend/src/js/components/Instances/Details.js
```diff
@@ -145,45 +147,41 @@ function StatusRow(props) {
...
-        {props.events.map((entry, i) =>
-          <StatusRow key={i} entry={entry} />
-        )
-        }
-      </TableBody>
-    </Table>
+  return props.events.length === 0 ? (
+    <Empty>
+      No events to report for this instance yet.
+    </Empty>) : (
```
joaquimrocha, line 153: nitpick: Can we have the ` : ` in the next line?

frontend/src/js/components/Instances/Details.js
```diff
@@ -145,45 +147,41 @@ function StatusRow(props) {
...
+      <Table>
+        <TableHead>
+          <TableRow>
+            <TableCell>Status</TableCell>
+            <TableCell>Version</TableCell>
+            <TableCell>Time{props.events && props.events.status}</TableCell>
```
joaquimrocha, line 159: I know this was already in the code before this patch, but should this be here?

frontend/src/js/components/Instances/Details.js
```diff
@@ -145,45 +147,41 @@ function StatusRow(props) {
...
-    };
+    API.getInstanceStatusHistory(application.id, group.id, instance.id).then((statusHistory)=>{
+      setEventHistory(statusHistory);
+    })
+      .catch(()=>{
+        setEventHistory([]);
+      });
   },
   [instance]);
-
```
joaquimrocha, line 186: Doesn't belong here.

frontend/src/js/components/Instances/Item.react.js
```diff
@@ -41,19 +42,17 @@ function Item(props) {
 
   function fetchStatusHistoryFromStore() {
     const selected = props.selected;
-
```
joaquimrocha, line 44: Doesn't belong here.

frontend/src/js/components/Instances/List.js
```diff
@@ -108,21 +109,26 @@ function ListView(props) {
...
-      setTotalInstances(result.total_instance);
-      instancesStore.setInstances(result.instances);
+      if (result.instances) {
+        const massagedInstances = result.instances.map((instance) => {
+          instance.statusInfo = instancesStore.getInstanceStatus(instance.application.status,
+              instance.application.version);
+          return instance;
+        });
+        setInstances(massagedInstances);
+        setTotalInstances(result.total_instance);
```
joaquimrocha, line 126: Maybe it makes sense to have the `total` and `instances` in the same status together (i.e. use `instances` as `{ total: ..., instances: ... }`) when you make the changes to use `useEffect` for the fetching only.

frontend/src/js/components/Instances/List.js
```diff
@@ -108,21 +109,26 @@ function ListView(props) {
...
+      } else {
+        setInstances([]);
+        setTotalInstances(result.total_instance);
+      }
+
```
joaquimrocha, line 131: Remove this extra newline.

frontend/src/js/components/Instances/Table.js
```diff
@@ -42,6 +43,7 @@ function Table(props) {
           <Item
             key={'instance_' + i}
             instance={instance}
+            instances={instances}
```
joaquimrocha, line 46: Why is this needed?

frontend/src/js/components/Layouts/InstanceLayout.js
```diff
@@ -14,30 +15,43 @@ export default function InstanceLayout(props) {
   const [instance, setInstance] = React.useState(null);
 
   function onChange() {
-    instancesStore.getInstances(appID, groupID, instanceID);
+    API.getInstance(appID, groupID, instanceID).then((instance)=>{
```
joaquimrocha, line 18: Surround `=>` by one space in each side.

frontend/src/js/components/Layouts/InstanceLayout.js
```diff
@@ -14,30 +15,43 @@ export default function InstanceLayout(props) {
...
+    });
     const apps = applicationsStore.getCachedApplications() || [];
     const app = apps.find(({id}) => id === appID);
     if (app !== application) {
       setApplication(app);
       setGroup(getGroupFromApplication(app));
     }
 
-    onChangeInstances();
+    //onChangeInstances();
```
joaquimrocha, line 28: Don't add commented out code.

frontend/src/js/components/Layouts/InstanceLayout.js
```diff
@@ -14,30 +15,43 @@ export default function InstanceLayout(props) {
...
-  React.useEffect(() => {
-    applicationsStore.addChangeListener(onChange);
-    instancesStore.addChangeListener(onChangeInstances);
+  // function onChangeInstances() {
+  //   const instances = instancesStore.getCachedInstances(appID, groupID) || [];
+  //   const inst = instances.find(({id}) =>id === instanceID);
+  //   if (inst !== instance) {
+  //     setInstance(inst);
+  //   }
+  // }
```
joaquimrocha, line 41: Don't add commented out code.

frontend/src/js/components/Layouts/InstanceLayout.js
```diff
@@ -14,30 +15,43 @@ export default function InstanceLayout(props) {
...
 
-    applicationsStore.getApplication(appID);
+  React.useEffect(() => {
+    if (applicationsStore.getCachedApplications()) {
+      onChange();
+    } else {
+      applicationsStore.addChangeListener(onChange);
+      //instancesStore.addChangeListener(onChangeInstances);
```
joaquimrocha, line 48: Don't add commented out code.

frontend/src/js/components/Layouts/InstanceLayout.js
```diff
@@ -14,30 +15,43 @@ export default function InstanceLayout(props) {
...
 
+      applicationsStore.getApplication(appID);
+    }
     return function cleanup() {
       applicationsStore.removeChangeListener(onChange);
-      instancesStore.removeChangeListener(onChangeInstances);
+      //instancesStore.removeChangeListener(onChangeInstances);
```
joaquimrocha, line 54: Don't add commented out code.

frontend/src/js/stores/ApplicationsStore.js
```diff
@@ -51,6 +51,8 @@ class ApplicationsStore extends Store {
             this.applications.unshift(applicationItem);
           }
           this.emitChange();
+        } else {
+          return application;
```
joaquimrocha, line 55: I am not sure this is how this method should behave.

pkg/api/instances.go
```diff
@@ -171,12 +176,30 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	err = api.instancesQuery(p).Paginate(p.Page, p.PerPage).QueryStructs(&instances)
+	if err != nil && err != sql.ErrNoRows {
+		return InstanceWithTotal{}, err
+	}
+	if err == sql.ErrNoRows {
+		return InstanceWithTotal{
+			TotalInstances: 0,
+			Instances:      nil,
+		}, nil
+	}
```
joaquimrocha, line 193: Nitpick: If you move this block before the `if` above, then you don't have to check again for whether `err` is `sql.ErrNoRows`.

frontend/src/js/components/Instances/List.js
```diff
@@ -115,74 +142,26 @@ function ListView(props) {
...
-        // @todo: Allow to manually refresh list from UI.
-        setInstances(cachedInstances);
-      }
-    }
-  }
-
   React.useEffect(() => {
-    instancesStore.addChangeListener(onChangeInstances);
     // @todo: This check avoids multiple unnecessary fetches, but we should
     // use a smarter refresh in the background that updates the list when needed.
```
joaquimrocha, line 157: This comment referred to something that you removed, so please remove the comment.

frontend/src/js/components/Instances/List.js
```diff
@@ -94,9 +95,8 @@ function ListView(props) {
   const {application, group} = props;
   const [page, setPage] = React.useState(0);
   const [rowsPerPage, setRowsPerPage] = React.useState(10);
-  const [instances, setInstances] = React.useState([]);
+  const [instancesObj, setInstancesObj] = React.useState({instances:[], total:0});
```
joaquimrocha, line 98: There should be a space between `key:` and the value, i.e. `total: 0`.

joaquimrocha: Also, I don't think these changes (using the `instancesObj`) belong to this commit, right?

frontend/src/js/components/Instances/List.js
```diff
@@ -108,21 +108,23 @@ function ListView(props) {
...
-      instancesStore.setInstances(result.instances);
+      if (result.instances) {
+        const massagedInstances = result.instances.map((instance) => {
+          instance.statusInfo = instancesStore.getInstanceStatus(instance.application.status,
+              instance.application.version);
+          return instance;
+        });
+        setInstancesObj({instances:massagedInstances, total:result.total});
+      } else {
+        setInstancesObj({instances:[], total:result.total});
```
joaquimrocha, line 126: Add a space between the keys and the values, as mentioned before.

frontend/src/js/stores/InstancesStore.js
```diff
@@ -6,39 +6,8 @@ class InstancesStore extends Store {
 
   constructor() {
     super();
-    this.instances = {};
```
joaquimrocha, line 9: (Comment for the entire patch)
If you leave the instances store just for this method, then it means we'll have a class just for providing a method that can even be static (`getInstanceStatus`).
If this is method is called in just one module, move it there. Otherwise move it to a helper/lib module like `constants/helpers`.

frontend/src/js/components/Layouts/InstanceLayout.js
```diff
@@ -29,14 +29,6 @@ export default function InstanceLayout(props) {
...
   }
 
-  function onChangeInstances() {
-    const instances = instancesStore.getCachedInstances(appID, groupID) || [];
-    const inst = instances.find(({id}) =>id === instanceID);
-    if (inst !== instance) {
-      setInstance(inst);
-    }
-  }
-
```
joaquimrocha, line 39: You remove this function but it's still being called in `useEffect`.

frontend/src/js/stores/InstancesStore.js
```diff
@@ -10,51 +10,20 @@ class InstancesStore extends Store {
...
-  getCachedInstances(applicationID, groupID) {
-    let cachedInstances = null;
-    if (this.instances.hasOwnProperty(applicationID)) {
-      if (this.instances[applicationID].hasOwnProperty(groupID)) {
-        cachedInstances = this.instances[applicationID][groupID];
-      }
-    }
-    return cachedInstances;
-  }
+  getInstances(applicationID, groupID, queryOptions = {}) {
```
joaquimrocha: I thought I had mentioned that there is no point in adding these functions here to later remove them in the same PR.

frontend/src/js/components/Instances/Details.js
```diff
@@ -210,15 +210,15 @@ function DetailsView(props) {
                 </Grid>
                 <Grid item xs={12}>
                   <CardFeatureLabel>Status:</CardFeatureLabel>&nbsp;
-                  <StatusLabel status={instance.statusInfo} />
+                  <StatusLabel status={instancesStore
+                    .getInstanceStatus(instance.application.status)}
```
joaquimrocha: AFAIU you already have this in `statusInfo` if you come from the list view, right?

frontend/src/js/components/Instances/Details.js
```diff
@@ -210,15 +210,15 @@ function DetailsView(props) {
...
+                  />
                 </Grid>
                 <Grid item xs={12}>
                   <CardFeatureLabel>Last Update Check:</CardFeatureLabel>&nbsp;
-                  {instance.statusInfo &&
                   <CardLabel>
-                    {makeLocaleTime(instance.statusInfo.created_ts)}
+                    {makeLocaleTime(instance.created_ts)}
```
joaquimrocha, line 219: Is this change correct? Is the timestamp the same as with `statusInfo` at this point?

frontend/src/js/constants/helpers.js
```diff
@@ -81,3 +81,83 @@ export function makeColorsForVersions(theme, versions, channel = null) {
 
   return versionColors;
 }
+export function getInstanceStatus(statusID, version) {
```
joaquimrocha, line 84: Add an extra newline to the top of this.

frontend/src/js/components/Layouts/InstanceLayout.js
ashu8912: yeah you are right i have updated this commit

joaquimrocha:
@ashu8912 I have added a new lint rule to the master and updated your PR with a fix for that. I have also fixed up some of your commit messages and tested the changes locally.

It'd still be great to add some tests for the pagination. We can talk to tomorrow about it.

pkg/api/instances.go
```diff
@@ -143,7 +148,7 @@ func (api *API) RegisterInstance(instanceID, instanceIP, instanceVersion, appID,
 }
 
 // GetInstance returns the instance identified by the id provided.
-func (api *API) GetInstance(instanceID, appID string) (*Instance, error) {
+func (api *API) GetInstance(instanceID string, appID string) (*Instance, error) {
```
krnowak, line 151: Unnecessary change.

pkg/api/instances.go
krnowak: Could you fix this one too? Thanks.
```suggestion
		Where("last_check_for_updates > now() at time zone 'utc' - interval $1", validityInterval).
```

pkg/api/instances.go
```diff
@@ -54,6 +55,10 @@ type Instance struct {
 	CreatedTs   time.Time           `db:"created_ts" json:"created_ts"`
 	Application InstanceApplication `db:"application" json:"application,omitempty"`
 }
+type InstanceWithTotal struct {
```
krnowak, line 58: @joaquimrocha suggested to use `InstancesWithTotal` which makes sense.
 
```suggestion
type InstancesWithTotal struct {
```

pkg/api/instances.go
```diff
@@ -171,12 +176,30 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
 }
 
 // GetInstances returns all instances that match with the provided criteria.
-func (api *API) GetInstances(p InstancesQueryParams) ([]*Instance, error) {
+func (api *API) GetInstances(p InstancesQueryParams) (InstanceWithTotal, error) {
```
krnowak, line 179:
```suggestion
func (api *API) GetInstances(p InstancesQueryParams) (InstancesWithTotal, error) {
```

pkg/api/instances.go
```diff
@@ -171,12 +176,30 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
-
-	err := api.instancesQuery(p).QueryStructs(&instances)
-
-	return instances, err
+	var totalCount uint64
+	var err error
+	p.Page, p.PerPage = validatePaginationParams(p.Page, p.PerPage)
+	err = api.instancesQuery(p).Paginate(p.Page, p.PerPage).QueryStructs(&instances)
+	if err == sql.ErrNoRows {
+		return InstanceWithTotal{
```
krnowak, line 186:
```suggestion
		return InstancesWithTotal{
```

pkg/api/instances.go
```diff
@@ -171,12 +176,30 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+			TotalInstances: 0,
+			Instances:      nil,
+		}, nil
+	}
+	if err != nil {
+		return InstanceWithTotal{}, err
```
krnowak, line 192:
```suggestion
		return InstancesWithTotal{}, err
```

pkg/api/instances.go
```diff
@@ -171,12 +176,30 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	}
+	err = api.getFilterInstancesQuery("COUNT (*)", p).QueryStruct(&totalCount)
+	if err != nil {
+		return InstanceWithTotal{}, err
+	}
+	result := InstanceWithTotal{
```
krnowak, line 198:
```suggestion
	result := InstancesWithTotal{
```

pkg/api/instances_test.go
krnowak: This still seems to be missing.

krnowak:
LFAD (looks fine and dandy). Thanks!

PR #178, opened 2020-04-14
Auto complete package in channel Edit Dialog

### LINKED ISSUE #46: Auto-complete list of packages when adding them to a channel
Showing the full list of packages in a select popup when adding assigning a package to a channel is a bit overwhelming and may cause performance issues depending on the number of packages, so this view should be replaced by an auto-complete popup (like the timezones' picker).

Here is the issue:
Screenshot from 2019-09-09 20-52-12

ashu8912:
fixes #46


ashu8912:
ooh this was added to the wrong branch sorry.
closing it now

PR #180, opened 2020-04-15
 Pass nebraska version via config

ashu8912:


schu:
Instead, you could pass in the exact Git version automatically during build. For example: https://github.com/kinvolk/lokomotive/blob/9bd75f062c4b7cd7c828826fffd2afcb8177a8d1/Makefile#L25

schu:
> Instead, you could pass in the exact Git version automatically during build. For example: https://github.com/kinvolk/lokomotive/blob/9bd75f062c4b7cd7c828826fffd2afcb8177a8d1/Makefile#L25

Only saw https://github.com/kinvolk/nebraska/issues/185 now.

Issue #175, opened 2020-04-13
make container-nebraska doesn't work in master


zabudskyi:
Hi,
[This path](https://github.com/kinvolk/nebraska/blob/d9cb21e3d1edf4e0e0e79a1d44f32e712559b97b/Dockerfile.nebraska#L21) is invalid now.
I built `nebraska-build` image and found out that right folder now is `build`, not `built`.
After changing this line I was able to build container but `/nebraska/static` folder didn't have `js/main.js` file, so frontend was not working.
```
ls /nebraska/static/
asset-manifest.json                                    robots.txt
favicon.png                                            service-worker.js
index.html                                             static
precache-manifest.a7af4f53058d64095abbdc2d325238f3.js
```
```
ls /nebraska/static/static/js/
2.a02f9ab0.chunk.js              2.a02f9ab0.chunk.js.map          main.f19aaee0.chunk.js.map       runtime-main.0a01877e.js.map
2.a02f9ab0.chunk.js.LICENSE.txt  main.f19aaee0.chunk.js           runtime-main.0a01877e.js
```

It works fine for commit dated Mar 16 (when 2.0.5 container was built)

joaquimrocha:
Hi @zabudskyi , we're aware of this problem and will have it fixed soon.

joaquimrocha:
This is fixed in master now.

joaquimrocha:
Thanks for reporting, @zabudskyi .

PR #181, opened 2020-04-15
Fix static serving and Docker image generation

joaquimrocha:
Since we switched to `create-react-app`, the generated directory has a different name and contents, so we need to update which files the static server and the dockerfile use.


joaquimrocha:
This fixes #175 BTW.

PR #187, opened 2020-04-18
docker: remove postgres-docker-entrypoint.sh

schu:
Addendum to e3353103c10beb083a9998e46171ce85c6eb632b - the script is not
needed any longer.


joaquimrocha:
I have simply capitalized the commit message's title (after the prefix) to be consistent with the rest.
Thanks for the patch.

PR #193, opened 2020-04-21
Make queries slimmer

joaquimrocha:
This PR decouples fetching the version breakdown + the instances stats from when getting a group's information.
Those queries were done in order to make it simpler for the UI to show the info, but they're a bit heavy so we should avoid them when they're not needed.

The PR also adds connections limits to the DB (overridable by the use of env vars).

### DIFF
--- a/frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
+++ b/frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
@@ -3,6 +3,7 @@ import { makeStyles } from '@material-ui/core/styles';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { Link as RouterLink } from 'react-router-dom';
+import API from '../../api/API';
 
 const useStyles = makeStyles(theme => ({
   groupLink: {
@@ -12,7 +13,19 @@ const useStyles = makeStyles(theme => ({
 
 function ApplicationItemGroupItem(props) {
   const classes = useStyles();
-  const instances_total = props.group.instances_stats.total ? '(' + props.group.instances_stats.total + ')' : '';
+  const {group} = props;
+  const [totalInstances, setTotalInstances] = React.useState(-1);
+
+  React.useEffect(() => {
+    // We use this function without any filter to get the total number of instances
+    // in the group.
+    API.getInstances(group.application_id, group.id)
+      .then(result => {
+        setTotalInstances(result.total);
+      })
+      .catch(err => console.error('Error loading total instances in Instances/List', err));
+  },
+  [group]);
 
   return (
     <Link
@@ -20,7 +33,7 @@ function ApplicationItemGroupItem(props) {
       to={{pathname: `/apps/${props.group.application_id}/groups/${props.group.id}`}}
       component={RouterLink}
     >
-      {props.group.name} {instances_total}
+      {props.group.name} {(totalInstances > 0) && `(${totalInstances})`}
     </Link>
   );
 }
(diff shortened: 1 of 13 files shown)


frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
```diff
@@ -12,15 +13,27 @@ const useStyles = makeStyles(theme => ({
 
 function ApplicationItemGroupItem(props) {
   const classes = useStyles();
-  const instances_total = props.group.instances_stats.total ? '(' + props.group.instances_stats.total + ')' : '';
+  const {group} = props;
+  const [totalInstances, setTotalInstances] = React.useState(-1);
```
ashu8912, line 17: I think the default totalInstance count should be 0

frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
```diff
@@ -12,15 +13,27 @@ const useStyles = makeStyles(theme => ({
...
+  [group]);
 
   return (
     <Link
       className={classes.groupLink}
       to={{pathname: `/apps/${props.group.application_id}/groups/${props.group.id}`}}
       component={RouterLink}
     >
-      {props.group.name} {instances_total}
+      {props.group.name} {(totalInstances > 0) && `(${totalInstances})`}
```
ashu8912, line 36: I think template literals are not required here

frontend/src/js/components/Groups/Item.react.js
```diff
@@ -21,13 +22,10 @@ const useStyles = makeStyles(theme => ({
 
 function Item(props) {
   const classes = useStyles();
+  const [totalInstances, setTotalInstances] = React.useState(-1);
```
ashu8912, line 26: Same default should be 0

frontend/src/js/components/Instances/List.js
```diff
@@ -95,6 +95,7 @@ function ListView(props) {
   const {application, group} = props;
   const [page, setPage] = React.useState(0);
   const [rowsPerPage, setRowsPerPage] = React.useState(10);
+  const [totalInstances, setTotalInstances] = React.useState(-1);
```
ashu8912, line 99: ^ default should be 0

frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
joaquimrocha: Nope. This is in order to give it an "unitialized" value (value yet not returned by the server), so we can decide to show something else based on that fact.

frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
joaquimrocha: What do you mean, I want the `(...)` to be shown only if the `totalInstances` are greater than 0.

frontend/src/js/components/Groups/Item.react.js
joaquimrocha: See my reply from before.

frontend/src/js/components/Instances/List.js
joaquimrocha: See my reply from before.

frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
ashu8912: got it.

frontend/src/js/components/Applications/ApplicationItemGroupItem.react.js
ashu8912: hmm this is cool but than we need to decide what to show when there is a default state coz right now seeing -1 instances in UI looks very odd

joaquimrocha: Oh! That's a different thing (and a good catch)! Let me check it out.

joaquimrocha:
Merging after fixing the `-1` in the UI that @ashu8912 pointed out.

PR #195, opened 2020-04-22
Make layouts responsive

### LINKED ISSUE #55: Make layouts responsive
The layouts are not set up to be responsive because of the extra testing required for it.
But we should planning on doing this work now that most of the core changes are getting tackled.
So let's track it here.

ashu8912:
fixes  #55


joaquimrocha:
Let's pause this until we have the new theme.

joaquimrocha:
Let's close this for now as we need to rewrite many of it.

PR #197, opened 2020-04-28
Fix unknown status filter bug

ashu8912:


joaquimrocha:
I have updated the commit's message to make it clearer.

PR #200, opened 2020-05-04
Fix channel color overriding

ashu8912:


joaquimrocha:
Merged after fixing the commit message a bit.

Issue #202, opened 2020-05-07
Don't show the duration links if there is nothing to be shown


joaquimrocha:
I have noticed that we show the duration links in the group extended view even if they don't have any instances to report:

![Screenshot from 2020-05-06 15-15-38](https://user-images.githubusercontent.com/1029635/81182082-a6df7d00-8fad-11ea-86d9-991dadb85102.png)

Ideally we disable those links if there is no duration for the user to see.
The strategy for doing this should be:

1. get the instances/count for the default duration (1 day in this case);
2. if nothing is reported, set the link's state to disabled and repeat 1 but for the longer duration after the current one.

This should allow the user to get directly into a duration that is useful.
If this requires a long time when no instances for 30d are there, then we can take a lazy approach:

1. get the instances/count for the current duration (default 1 day in this case);
2. if nothing is returned, disable the current duration link and the ones that are less than it.

This means that if the user gets to the view (1 day) and there are no instances, then 1day and 1hour durations are disabled. If they then click on 30 days and there is nothing, 30 day and less gets disabled (i.e. all the links get disabled).

ashu8912:
@joaquimrocha this can be closed now as we have the changes in master now

PR #209, opened 2020-05-12
Disable periods for which no instances available

ashu8912:


ashu8912:
fixes #202

frontend/src/js/components/Common/TimeIntervalLinks.js
```diff
@@ -10,17 +11,41 @@ const useStyles = makeStyles(() => ({
...
-            onClick={(e) => intervalChangeHandler(link)}
-          >
-            <Link underline="none" component="button" color={link.displayValue === selectedInterval.displayValue ? 'inherit' : 'primary'}>
-              <Typography className={classes.title}>
+          <Grid key={link.queryValue} item>
+            <Link underline="none"
+              component="button"
+              onClick={(e) => intervalChangeHandler(link)}
+              disabled={link.disabled}
+              color={link.disabled ? '#fafafa' : link.displayValue === selectedInterval.displayValue ? 'inherit' : 'primary'}
```
joaquimrocha, line 46: If the `disabled` property doesn't set the color already, then don't use a hardcoded color; use `textSecondary` instead.
If this doesn't work (because it's passing the color to the button component, sorry, I haven't tested), then please use the `default` value for when it's disabled, as then the disabled prop should set the color.

frontend/src/js/components/Common/TimeIntervalLinks.js
```diff
@@ -10,17 +11,41 @@ const useStyles = makeStyles(() => ({
...
+        const timeIntervalsToUpdate = [...timeIntervals];
+        for (let i = 0; i < results.length; i++) {
+          if (results[i] === 0) {
+            const timeInterval = {...timeIntervalsToUpdate[i]};
+            timeInterval.disabled = true;
+            timeIntervalsToUpdate[i] = timeInterval;
+          }
+        }
+        setTimeIntervals(timeIntervalsToUpdate);
+      });
```
joaquimrocha, line 32: How long does this take?
I wonder if making it smarter makes it more confusing, i.e.: if there is an valid instances count for a period, then the same is true for periods longer than the current one. e.g. if there are values for 1h, there have to be values for all the periods. This is what I tried to express in the original issue.

frontend/src/js/constants/helpers.js
```diff
@@ -16,14 +16,14 @@ export const ARCHES = {
...
-  {displayValue: '1 hour', queryValue: '1h'}
+export const timeIntervalsDefault = [
+  {displayValue: '30 days', queryValue: '30d', disabled: false},
+  {displayValue: '7 days', queryValue: '7d', disabled: false},
+  {displayValue: '1 day', queryValue: '1d', disabled: false},
+  {displayValue: '1 hour', queryValue: '1h', disabled: false}
 ];
 
-export const defaultTimeInterval = timeIntervals
+export const defaultTimeInterval = timeIntervalsDefault
```
joaquimrocha, line 26: Nitpick: You didn't have to change this name, as it was implicit that the intervals were default already.

frontend/src/js/components/Common/TimeIntervalLinks.js
ashu8912: @joaquimrocha this takes fraction of seconds because we query instance count only so it is very fast

joaquimrocha:
I tried the PR now and we need something more to improve the UX:
Disabling the links makes sense but now I see it brings some issues:
  1. The mouse pointer still changes to a hand pointer (suggesting there is a link but nothing happens on click);
  2. Nothing explains why the links don't work (i.e. we would have to use a tooltip for the disabled links explaning that there are no instances for that period);
  3. We go into the group's extended view by using the default period of 24h, but it may have no instances, and then once we click in a valid link, the link to where we were is disabled which is inconsistent.
  4. Out empty message still says no instances have registered yet, even if they have but for an older period.

I think we can use a simple fix for the moment (instead of touching many places):
  1. Don't disable the links, just use the `textSecondary` color for them (so they suggest there is nothing interesting there).
  2. Slightly change the empty message to:
```
No instances have registered with this group for the past PERIOD_HERE.


Instances will be shown here automatically the next time they request an update.
```

frontend/src/js/__tests__/Common/TimeIntervalLinks.spec.js
```diff
@@ -1,12 +1,12 @@
 import {render} from '@testing-library/react';
 import React from 'react';
 import TimeIntervalLinks from '../../components/Common/TimeIntervalLinks';
-import { timeIntervals } from '../../constants/helpers';
+import { timeIntervalsDefault, defaultTimeInterval } from '../../constants/helpers';
```
joaquimrocha, line 4: The linter is complaining.

frontend/src/js/__tests__/Common/TimeIntervalLinks.spec.js
```diff
@@ -1,12 +1,12 @@
...
 
 describe('TimeIntervalLinks', () => {
   it('should render correct time Interval links', () => {
-    const {getByText} = render(<TimeIntervalLinks/>);
-    timeIntervals.forEach((timeInterval) => {
+    const {getByText} = render(<TimeIntervalLinks selectedInterval={defaultTimeInterval}/>);
+    timeIntervalsDefault.forEach((timeInterval) => {
```
joaquimrocha, line 9: I am on the fence about having one commit fixing tests here. We're supposed to add commits that let the project being built + run, and I think including existing tests in this criteria is a good idea. So I would just squash this with the patch that introduced the `selectedInterval`, but it seems that patch was merged already, so it's fine here if that's the case.

joaquimrocha:
Updated the commit message. Will merge after CI succeeds.

PR #215, opened 2020-05-18
Add tests for application module

ashu8912:


joaquimrocha:
There are linter warnings for this PR. Maybe we have changed something because those warnings should be errors.

PR #217, opened 2020-05-18
Silent test console warnings

ashu8912:


ashu8912:
@joaquimrocha I have added this pr but its still open to discussion, whether we should silent console warnings for tests or not

PR #222, opened 2020-06-05
Fix charts click and duration change bug

ashu8912:


joaquimrocha:
Merged after fixing up the commit message.

PR #226, opened 2020-06-12
Add limit to request body size for /omaha and /v1/update

ashu8912:


cmd/nebraska/controller.go
```diff
@@ -821,6 +821,7 @@ func (ctl *controller) getMetrics(c *gin.Context) {
 
 func (ctl *controller) processOmahaRequest(c *gin.Context) {
 	c.Writer.Header().Set("Content-Type", "text/xml")
+	c.Request.Body = http.MaxBytesReader(c.Writer, c.Request.Body, 64*1024)
```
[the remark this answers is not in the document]
krnowak: Yep, a named constant would be nice.

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +81,23 @@ func TestClientConfig(t *testing.T) {
...
+		noopAuthConfig: noopAuthConfig,
+	}
+
+	ctl, _ := newController(conf)
+	gin.SetMode(gin.TestMode)
+
+	w := httptest.NewRecorder()
+	c, _ := gin.CreateTestContext(w)
+	c.Request = &http.Request{}
+	xmlFile, _ := os.Open("sample_request.xml")
```
joaquimrocha, line 97: I think you should create a 64k `bytes.Buffer` and then add the necessary XML strings wrapping it, so basically we could skip having a 64k file shipped in the repo.

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +83,39 @@ func TestClientConfig(t *testing.T) {
...
+	<request protocol="3.0" version="update_engine-0.4.10" updaterversion="update_engine-0.4.10" installsource="scheduler" ismachine="1">
+		<os version="Chateau" platform="CoreOS" sp="2512.2.0_x86_64"></os>
+		<app appid="e96281a6-d1af-4bde-9a0a-97b76e56dc57" version="1.2.3" track="stable" bootid="{965fb4c5-ad3e-4eb7-a4c2-ca0c0e31ec84}" oem="ami" oemversion="0.1.1-r1" alephversion="1688.5.3" machineid="INSERTINSTANCEIDHERE" lang="en-US" board="amd64-usr" hardware_class="" delta_okay="false" >
+			<ping active="1"></ping>
+			<updatecheck></updatecheck>
+			<event eventtype="3" eventresult="1"></event>
+		</app>`)
+	xmlRequest.WriteString("<!--")
+	for i := 0; i <= 64*1024; i++ {
+		xmlRequest.WriteString("gibberish")
```
joaquimrocha, line 110: This means you're testing requests with size = 9 (`gibberish`) * 64 * 1024 + the rest of the XML.

I am actually surprised that you need to send something that's slightly valid XML, as I'd assume the size check would prevent even checking the contents. If somehow the lib is checking whether the request looks like XML, maybe you just have to use `<` as the first char, and `/>` as the last.

So I'd try to make the request as:
1. Just 64 * 1024 * 'x' (or another character); if it fails then
2. Use the XML opening tag + the content (64KB minus the number of chars around for the tags) + closing tags.

Finally, the test should check that the request fails with 64KB chars, but is allowed with 64KB - 1.

cmd/nebraska/controller.go
```diff
@@ -821,8 +822,12 @@ func (ctl *controller) getMetrics(c *gin.Context) {
 
 func (ctl *controller) processOmahaRequest(c *gin.Context) {
 	c.Writer.Header().Set("Content-Type", "text/xml")
+	c.Request.Body = http.MaxBytesReader(c.Writer, c.Request.Body, UpdateMaxRequestSize)
 	if err := ctl.omahaHandler.Handle(c.Request.Body, c.Writer, getRequestIP(c.Request)); err != nil {
 		logger.Error("process omaha request", "error", err)
+		if err.Error() == fmt.Errorf("%s: %s", omaha.ErrMalformedRequest, "http: request body too large").Error() {
```
krnowak, line 828: Could you see if `if uerr := errors.Unwrap(err); uerr != nil && uerr.Error() == "http: request body too large"` would work?
```suggestion
		if uerr := errors.Unwrap(err); uerr != nil && uerr.Error() == "http: request body too large" {
```

pkg/omaha/omaha.go
```diff
@@ -69,7 +70,7 @@ func (h *Handler) Handle(rawReq io.Reader, respWriter io.Writer, ip string) erro
 
 	if err := xml.NewDecoder(rawReq).Decode(&omahaReq); err != nil {
 		logger.Warn("Handle - malformed omaha request", "error", err.Error())
-		return ErrMalformedRequest
+		return fmt.Errorf("%s: %s", ErrMalformedRequest, err.Error())
```
krnowak, line 73: Please use `%w` verb for errors. With it, `fmt.Errorf` will wrap `err`, which you can later retrieve. For more about it, see https://golang.org/doc/go1.13#error_wrapping
```suggestion
		return fmt.Errorf("%s: %w", ErrMalformedRequest, err)
```

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +83,39 @@ func TestClientConfig(t *testing.T) {
...
 	assert.NotEmpty(t, ctl.clientConfig.NebraskaVersion)
 }
+
+func TestOmahaRequestSizeLimitation(t *testing.T) {
+	noopAuthConfig := &auth.NoopAuthConfig{}
+	conf := &controllerConfig{
+		noopAuthConfig: noopAuthConfig,
+	}
+
+	ctl, _ := newController(conf)
```
krnowak, line 93:
```suggestion
	// import github.com/stretchr/testify/require
	ctl, err := newController(conf)
	require.NoError(t, err)
```

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +83,39 @@ func TestClientConfig(t *testing.T) {
...
+	gin.SetMode(gin.TestMode)
+
+	w := httptest.NewRecorder()
+	c, _ := gin.CreateTestContext(w)
+	c.Request = &http.Request{}
+	var xmlRequest strings.Builder
+	xmlRequest.WriteString(`<?xml version="1.0" encoding="UTF-8"?>
```
krnowak, line 100:
```suggestion
	xmlRequest := bytes.NewBufferString(`<?xml version="1.0" encoding="UTF-8"?>
```

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +83,39 @@ func TestClientConfig(t *testing.T) {
...
+	xmlRequest.WriteString("<!--")
+	for i := 0; i <= 64*1024; i++ {
+		xmlRequest.WriteString("x")
+	}
+	xmlRequest.WriteString("-->")
+	xmlRequest.WriteString("</request>")
+	b := []byte(xmlRequest.String())
+	requestBuffer := new(bytes.Buffer)
+	requestBuffer.Write(b)
+	c.Request.Body = ioutil.NopCloser(requestBuffer)
```
krnowak, line 117:
```suggestion
	c.Request.Body = ioutil.NopCloser(xmlRequest)
```

cmd/nebraska/controller_test.go
krnowak: We probably could check the `Content-Length` header to see if the body isn't too large. Not sure though if it's a good defence against DoS. We would need to see if the golang standard library is actually limiting the number of read bytes from the body to the number specified in the Content-Length header. If we agree that checking the Content-Length is not enough, then sending valid XML is a must, otherwise the XML decoder will bail out with an error about binary data not being a valid XML long before we even reach the size limit. Also, I'm not exactly sure how the decoder works (I haven't read the source), so to be on the cautious side, I'd sent not only any valid XML, but a valid Omaha request. This certainly complicates some maths to get message with size 64KB - 1, but it's still doable.

cmd/nebraska/controller_test.go
ashu8912: done

ashu8912:
@krnowak done all the changes thanks

cmd/nebraska/controller_test.go
krnowak: So, I didn't mean adding the comment there - the comment was about adding an import at the top of the file. The error is still ignored and my suggestion was about ensuring that the call to `newController` returns no error. `require.NoError` will fail the test immediately (so it won't continue running) if `error` is not nil.

ashu8912: aah i misread it, thanks

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +83,37 @@ func TestClientConfig(t *testing.T) {
...
+			<event eventtype="3" eventresult="1"></event>
+		</app>`)
+	xmlRequest.WriteString("<!--")
+	lenRequestXML := xmlRequest.Len()
+	//insert (64kb - (the valid xml + the bytes for xml comments)) bytes into the large xml dummy request
+	for i := 0; i <= 64*1024-(lenRequestXML+13); i++ {
+		xmlRequest.WriteString("x")
+	}
+	xmlRequest.WriteString("-->")
+	xmlRequest.WriteString("</request>")
```
krnowak, line 114: So, we wanted to have two test cases, one that checks if `processOmahaRequest` returns no error for request with 64kb body and one that check is `processOmahaRequest` returns an error for request with 64kb+1 body. You could likely do it like:

```go
type testCase struct {
	bodyLength int
	status int
}
for _, tc := range []testcase{
	{
		bodyLength: 64*1024,
		status: http.StatusOk,
	},
	{
		bodyLength: 64*1024+1,
		status: http.StatusBadRequest,
	},
} {
	w := …
	c, _ := …
	xmlRequest := bytes.NewStringBuffer(…)
	…
	for i := 0; i <= tc.bodyLength - (…) {…}
	…
	ctl.processOmahaRequest(c)
	assert.Equal(t, tc.status, w.Code)
}
```

Also, make sure your maths are correct in the `for` loop condition. You can ensure it in your test with `assert.Equal(t, tc.bodyLength, xmlRequest.Len())`.

joaquimrocha: I had told that to @ashu8912 but he said that the valid XML length case is already tested in other tests. Still I am thinking now that we should have the valid XML test in any case, as this allows us to have a complete test case for these limits.

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +83,37 @@ func TestClientConfig(t *testing.T) {
...
+		<os version="Chateau" platform="CoreOS" sp="2512.2.0_x86_64"></os>
+		<app appid="e96281a6-d1af-4bde-9a0a-97b76e56dc57" version="1.2.3" track="stable" bootid="{965fb4c5-ad3e-4eb7-a4c2-ca0c0e31ec84}" oem="ami" oemversion="0.1.1-r1" alephversion="1688.5.3" machineid="INSERTINSTANCEIDHERE" lang="en-US" board="amd64-usr" hardware_class="" delta_okay="false" >
+			<ping active="1"></ping>
+			<updatecheck></updatecheck>
+			<event eventtype="3" eventresult="1"></event>
+		</app>`)
+	xmlRequest.WriteString("<!--")
+	lenRequestXML := xmlRequest.Len()
+	//insert (64kb - (the valid xml + the bytes for xml comments)) bytes into the large xml dummy request
+	for i := 0; i <= 64*1024-(lenRequestXML+13); i++ {
```
joaquimrocha, line 110: I understand you are accounting for what comes next, but you should avoid unorthogonal code like this. Just keep a buffer/string with the `--></request>` assigned before the loop, so you get the full width (lenght of xmlReq + length of the string), then you write using that string var.
There is also no real need to keep readabale whitespace in it (i.e. just join`--></request>`).

cmd/nebraska/controller_test.go
ashu8912: @krnowak for testing at 64kb i need a legit omaha request because any valid xml type wont work we should have the app and other infos passed in request actually in the db otherwise we get a segmentation fault, also a check for request below 64kb not exactly at 64 but less than 64kb is already done in omaha_test.go

ashu8912: it was failing because i didn't initialize api.db anywhere and was accessing properties from it,   and not because of passing invalid app ids , sorry for the confusion :)

cmd/nebraska/controller_test.go
```diff
@@ -78,3 +84,59 @@ func TestClientConfig(t *testing.T) {
...
+		lenRequestXML := xmlRequest.Len()
+		//insert (64kb - (the valid xml + the bytes for xml comments)) bytes into the large xml dummy request
+		for i := 1; i <= tc.bodyLength-(lenRequestXML+tempBuffer.Len()); i++ {
+			xmlRequest.WriteString("x")
+		}
+		xmlRequest.WriteString(tempBuffer.String())
+		assert.Equal(t, tc.bodyLength, xmlRequest.Len())
+		c.Request.Body = ioutil.NopCloser(xmlRequest)
+		ctl.processOmahaRequest(c)
+		//400 code is returned if request size is too large
```
krnowak, line 139: This comment is outdated.

PR #236, opened 2020-06-29
Override logo based on config

ashu8912:


cmd/nebraska/controller.go
```diff
@@ -27,8 +31,10 @@ const (
 
 // ClientConfig represents Nebraska's configuration of interest for the client.
 type ClientConfig struct {
-	AccessManagementURL string `json:"access_management_url"`
-	NebraskaVersion     string `json:"nebraska_version"`
+	AccessManagementURL string      `json:"access_management_url"`
+	NebraskaVersion     string      `json:"nebraska_version"`
+	Logo                string      `json:"logo"`
+	AppTitle            interface{} `json:"app_title"`
```
joaquimrocha, line 37: Please call this just `Title` and `json:"title"`.

cmd/nebraska/controller.go
```diff
@@ -112,7 +118,27 @@ func NewClientConfig(conf *controllerConfig) *ClientConfig {
 		}
 	}
 	config.NebraskaVersion = version.Version
-
+	if *appConfigEnv != "" {
+		fileName := filepath.Base(*appConfigEnv)
+		configName := strings.TrimRight(fileName, filepath.Ext(fileName))
+		viper.SetConfigName(configName)
+		viper.AddConfigPath(filepath.Dir(*appConfigEnv))
+		err := viper.ReadInConfig()
```
joaquimrocha, line 126: Isn't it simpler to use https://godoc.org/github.com/spf13/viper#ReadConfig ?

cmd/nebraska/controller.go
```diff
@@ -112,7 +118,27 @@ func NewClientConfig(conf *controllerConfig) *ClientConfig {
...
+		if err != nil {
+			return nil
+		}
+		logoPath := viper.Get("logo")
+		rootPath, err := os.Getwd()
+		if err != nil {
+			return nil
+		}
+		svg, err := ioutil.ReadFile(filepath.Join(rootPath, fmt.Sprintf("%v", logoPath)))
```
joaquimrocha, line 135: AFAIU, this won't work if we pass an absolute path to the logo. So maybe you have to check whether the path starts by a `/` before joining to the CWD.

cmd/nebraska/controller.go
```diff
@@ -112,7 +118,27 @@ func NewClientConfig(conf *controllerConfig) *ClientConfig {
...
+		if err != nil {
+			return nil
+		}
+		config.Logo = string(svg)
+		config.AppTitle = viper.Get("name")
```
joaquimrocha, line 140: Let's make it consistent and call the variable for title `title` (set in the file).

frontend/src/js/components/Header.react.js
```diff
@@ -61,11 +61,14 @@ export default function Header() {
   return (
     <AppBar position='static' className={classes.header}>
       <Toolbar>
-        {projectLogo &&
+        {config && config.logo ?
+          <div dangerouslySetInnerHTML={{__html: config.logo}}/> :
```
joaquimrocha: You are adding this on top of the old `projectLogo`, so if both exist, both will be shown.
It should show the overridden logo when it's passed in the client config, otherwise it should show Nebraska's logo.

frontend/src/js/components/Header.react.js
```diff
@@ -61,11 +61,14 @@ export default function Header() {
...
           <Icon icon={projectLogo} height={45} />
         }
         <Typography variant='h6' className={classes.title}>
-          {process.env.REACT_APP_PROJECT_NAME}
+          {
+            config && config.app_title
+          }
```
joaquimrocha, line 72: Let's actually only use the whole `Typography` element if a `title` is set up in the `config`.

frontend/src/js/components/Header.react.js
```diff
@@ -62,7 +63,7 @@ export default function Header() {
     <AppBar position='static' className={classes.header}>
       <Toolbar>
         {config && config.logo ?
-          <div dangerouslySetInnerHTML={{__html: config.logo}}/> :
+          <div dangerouslySetInnerHTML={{__html: DOMPurify.sanitize(config.logo)}}/> :
```
joaquimrocha, line 66: Please squash this whole commit with the frontend one.

frontend/src/js/components/Header.react.js
ashu8912: @joaquimrocha i didn't get you? if "config is present and has a logo on it than we show the logo" otherwise we stick to the default logo. Am I missing something here?

joaquimrocha: Ah sorry. I failed to see the conditions here (used to look for the colon in the next line). All good.

frontend/src/js/components/Header.react.js
```diff
@@ -27,6 +29,9 @@ const useStyles = makeStyles(theme => ({
     marginBottom: theme.spacing(1),
     background: process.env.REACT_APP_APPBAR_BG || theme.palette.primary.main,
   },
+  svgContainer: {
+    '& svg': {maxHeight: '5rem', maxWidth: '10rem'}
```
joaquimrocha, line 33: Nitpick: All of our style rules so far are one per line.
(Not really important, but if you're touching the PR again maybe include a change)

ashu8912:
@joaquimrocha i have added the caching part as well

frontend/src/js/components/Header.react.js
```diff
@@ -49,6 +55,14 @@ export default function Header() {
     if (!config) {
       API.getConfig()
         .then(config => {
+          if (!localStorage.getItem('nebraska_config') && config.logo !== ''){
```
joaquimrocha, line 58: Since you're getting the data from the backend anyway, maybe here we could always override/update what's locally stored. So the cache is up to date when needed, and still prevents us from showing the default logo/title temporarily if they exist in local storage.

frontend/src/js/components/Header.react.js
```diff
@@ -49,24 +55,36 @@ export default function Header() {
...
+          const cacheConfig = {
+            title: config.title,
+            logo: config.logo
+          };
+          localStorage.setItem('nebraska_config', JSON.stringify(cacheConfig));
+          setCachedConfig(cacheConfig);
           setConfig(config);
-        })
+        }
+        )
```
joaquimrocha, line 66: Not needed.

cmd/nebraska/controller.go
```diff
@@ -112,7 +118,32 @@ func NewClientConfig(conf *controllerConfig) *ClientConfig {
 		}
 	}
 	config.NebraskaVersion = version.Version
-
+	if *appConfigEnv != "" {
+		viper.SetConfigType("env")
```
krnowak, line 122: I wonder why we even use viper. This all could be done by replacing the `-client-config` flag with `-logo-path` and `-title` flags. The `-logo-path` would make it clear that it needs to be a path, so either relative to the current working directory or absolute.

cmd/nebraska/controller.go
krnowak: `filepath.Join("/foo", "/bar")` will return `"/foo/bar"`. But I don't know why are we doing this path joining. It's only going to be confusing if we specify `/usr/share/vendor/logo.png`.

cmd/nebraska/controller.go
```diff
@@ -27,8 +31,10 @@ const (
 
 // ClientConfig represents Nebraska's configuration of interest for the client.
 type ClientConfig struct {
-	AccessManagementURL string `json:"access_management_url"`
-	NebraskaVersion     string `json:"nebraska_version"`
+	AccessManagementURL string      `json:"access_management_url"`
+	NebraskaVersion     string      `json:"nebraska_version"`
+	Logo                string      `json:"logo"`
```
krnowak, line 36: This seems to be contents of the logo image. These can be arbitrary bytes, so I think it either should be a `[]byte` type. I think that the assumption is that we only will set an svg file as a logo, which basically is an XML, so the contents will likely be an UTF8-encoded string. But if we set the logo to some PNG file, then I'm not sure if the serialization to json will work.

So, this either needs to be an array of bytes (`[]byte`) or it needs to be encoded to string (with base64 for example).

ashu8912: @krnowak we are only accepting svgs for logo

ashu8912:
@joaquimrocha and @krnowak changes done please review.

cmd/nebraska/controller.go
```diff
@@ -112,7 +116,20 @@ func NewClientConfig(conf *controllerConfig) *ClientConfig {
...
 	config.NebraskaVersion = version.Version
-
+	config.Title = *appTitle
+	if *appLogoPath != "" {
+		svg, err := ioutil.ReadFile(*appLogoPath)
+		if err != nil {
+			logger.Error("Reading svg from path in config", err.Error())
+			return nil
+		}
+		if !utf8.Valid(svg) {
```
joaquimrocha, line 126: I think this may be a bit misleading. That function doesn't test whether it's a valid SVG, rather whether it's a valid UTF-8 AFAIU. I am not sure if we want to be so strict as to find an SVG lib and try loading it here, but at least the message for this check should be slightly different: `Invalid format for SVG`.
I am not sure actually if this check fails when a valid SVG is passed but the encoding is in something else than UTF-8. In any case this is not worth making too complicated, so maybe just change the error message.

cmd/nebraska/nebraska.go
```diff
@@ -44,6 +44,8 @@ var (
 	ghReadOnlyTeams     = flag.String("gh-ro-teams", "", "comma-separated list of read-only GitHub teams in the org/team format")
 	ghEnterpriseURL     = flag.String("gh-enterprise-url", "", fmt.Sprintf("base URL of the enterprise instance if using GHE; can be taken from %s env var too", ghEnterpriseURLEnvName))
 	logger              = log.New("nebraska")
+	appLogoPath         = flag.String("logo-path", "", fmt.Sprintf("Client logo path"))
+	appTitle            = flag.String("app-title", "", fmt.Sprintf("Client app title"))
```
joaquimrocha, line 48: Let's name this as `client-logo` and `client-title` please.

cmd/nebraska/nebraska.go
```diff
@@ -44,6 +44,9 @@ var (
 	ghReadOnlyTeams     = flag.String("gh-ro-teams", "", "comma-separated list of read-only GitHub teams in the org/team format")
 	ghEnterpriseURL     = flag.String("gh-enterprise-url", "", fmt.Sprintf("base URL of the enterprise instance if using GHE; can be taken from %s env var too", ghEnterpriseURLEnvName))
 	logger              = log.New("nebraska")
+	appLogoPath         = flag.String("client-logo", "", fmt.Sprintf("Client logo path"))
```
joaquimrocha, line 47: You added `Client app` to the other options, maybe here it should also be like that (i.e. `Client app logo`). I think adding `path` here makes it confusing (is it the path from the client side, etc?).

frontend/src/js/components/Main.react.js
```diff
@@ -41,6 +41,13 @@ const nebraskaTheme = createMuiTheme({
   }
 });
 
+const appBarTheme = createMuiTheme({
+  palette: {
+    dark: '#000000',
+    light: '#ffffff'
```
joaquimrocha, line 47: This is not what I meant when we talked on Slack.
Material UI themes already have a light/dark type for the pallete (expand all options here and check it out):
https://material-ui.com/customization/default-theme/

So what I would do is: create a theme that overrides the original type option to be `dark`, and set any needed colors for the app bar (because maybe the dark app bar's bg by default is not fully black).
Then use either the default (light) theme we are already setting for everything, or use the dark one if the client header style option says so.

The reason for this is that you'd likely have to override the style for the icon in the top bar too (because we do show an icon when there's an enterprise URL). And with this approach the text, icons, etc. all get the right colors for a dark bg.

cmd/nebraska/controller.go
krnowak: This check was my suggestion, just to ensure that the file is text. The rest of thinking was the same as yours - I don't think we want to parse it to check if this indeed is a valid SVG, but at least we can catch the case, when someone gives a path to a PNG file.

About encodings: meh. If someone is still using CP-1250 encoding, then I'd let them continue shooting their feet. Just don't bring this mess here, please. :)

Another option is to check if the file is a valid XML (since SVG is XML). In case you wanted it, you could import `encoding/xml` and do:

```go
if err := xml.Unmarshal(svg, &struct{}{}); err != nil {
	logger.Error("invalid xml (hence invalid svg)", err)
	return nil
}
```

frontend/src/js/components/Main.react.js
ashu8912: got it, but for the appBar in light mode we have to change the background coz we don't use the main color in theme for appBar background instead we use the white color(according to the new design).
I have added a new patch please take a look :)

frontend/src/js/components/Header.react.js
```diff
@@ -26,55 +28,53 @@ const useStyles = makeStyles(theme => ({
   },
   header: {
     marginBottom: theme.spacing(1),
-    background: process.env.REACT_APP_APPBAR_BG || theme.palette.primary.contrastText,
+    backgroundColor: (config) => config && config.appBarColor !== 'dark' &&
```
joaquimrocha, line 31: You don't need to pass it a config just for this, as you can already get the `type` of palette information from the `theme` passed here. So `theme.palette.type === 'dark' ? theme.palette.common.black : theme.palette.common.white` .

frontend/src/js/components/Header.react.js
```diff
@@ -26,45 +28,44 @@ const useStyles = makeStyles(theme => ({
...
-}));
-
-export default function Header() {
-  const classes = useStyles();
-  const [config, setConfig] = React.useState(null);
-  const projectLogo = _.isEmpty(nebraskaLogo) ? null : nebraskaLogo;
-
-  const [menuAnchorEl, setMenuAnchorEl] = React.useState(null);
+    backgroundColor: theme.palette.type === 'dark' ?
+      theme.palette.common.black : theme.palette.common.black
```
joaquimrocha, line 32: -> `theme.palette.common.black : theme.palette.common.black`
Did you test these changes with the different options?

ashu8912: sorry i actually messed up while pushing, yeah this works fine for both the header style I have tested it

cmd/nebraska/nebraska.go
```diff
@@ -44,6 +44,9 @@ var (
 	ghReadOnlyTeams     = flag.String("gh-ro-teams", "", "comma-separated list of read-only GitHub teams in the org/team format")
 	ghEnterpriseURL     = flag.String("gh-enterprise-url", "", fmt.Sprintf("base URL of the enterprise instance if using GHE; can be taken from %s env var too", ghEnterpriseURLEnvName))
 	logger              = log.New("nebraska")
+	appLogoPath         = flag.String("client-logo", "", fmt.Sprintf("Client app logo"))
+	appTitle            = flag.String("client-title", "", fmt.Sprintf("Client app title"))
+	appHeaderStyle      = flag.String("client-header-style", "light", fmt.Sprintf("Client app header style"))
```
schu, line 49: From the help text, it's not immediately clear to me if the value should be a hex color or CSS class name or something else?

Same for `client-logo` - maybe tell the user that it should be the path to a svg file?

ashu8912: Thanks @schu I have updated the helper text

schu: Thanks!

PR #238, opened 2020-06-29
Fix type for InstanceStats

ashu8912:


pothos:
Thanks, it solved the problem for me.

PR #240, opened 2020-07-01
backend: Add error logging for API package

pothos:
Errors like those reported from the database library went unnoticed
    because there was no error logging in the API package.
    Log all errors with enough context to relate them to request errors
    when the original errors are turned into other errors for the client.

### DIFF
--- a/pkg/api/updates.go
+++ b/pkg/api/updates.go
@@ -57,6 +57,7 @@ var (
 func (api *API) GetUpdatePackage(instanceID, instanceIP, instanceVersion, appID, groupID string) (*Package, error) {
 	instance, err := api.RegisterInstance(instanceID, instanceIP, instanceVersion, appID, groupID)
 	if err != nil {
+		logger.Error("GetUpdatePackage - could not register instance (propagates as ErrRegisterInstanceFailed)", err)
 		return nil, ErrRegisterInstanceFailed
 	}
 	updateAlreadyGranted := false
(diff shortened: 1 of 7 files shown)


pkg/api/updates.go
```diff
@@ -57,6 +60,9 @@ var (
 func (api *API) GetUpdatePackage(instanceID, instanceIP, instanceVersion, appID, groupID string) (*Package, error) {
 	instance, err := api.RegisterInstance(instanceID, instanceIP, instanceVersion, appID, groupID)
 	if err != nil {
+		if err != nil {
```
ashu8912, line 63: I think we can remove this check as we are already checking it above

ashu8912:
I agree that we need to log the db errors and not just return our custom ErrMessage as it makes difficult to debug if something breaks.
Thanks for the pr @pothos , i have added a nitpick suggestion otherwise Looks Good From MY Side.
@krnowak @joaquimrocha

pkg/api/updates.go
pothos: Yes, thanks

pkg/api/updates.go
```diff
@@ -76,15 +80,20 @@ func (api *API) GetUpdatePackage(instanceID, instanceIP, instanceVersion, appID,
...
+		}
 		return nil, ErrNoPackageFound
 	}
 
 	for _, blacklistedChannelID := range group.Channel.Package.ChannelsBlacklist {
 		if blacklistedChannelID == group.Channel.ID {
 			if updateAlreadyGranted {
-				// TODO: Log any error
-				_ = api.updateInstanceStatus(instance.ID, appID, InstanceStatusComplete)
+				err = api.updateInstanceStatus(instance.ID, appID, InstanceStatusComplete)
```
joaquimrocha, line 93: Nitpick: Other blocks seem to define the `err` within the if evaluation block. (Notice the nitpick/no-worries prefix)

pothos: Done, also in other places.

pothos:
I've updated the PR to cover all API packages

PR #241, opened 2020-07-01
backend: Make test timing more robust and explain it

pothos:
At least on my machine the tests started to fail because of timing
    issues and it was also observed before.
    Make the timing more robust by using 500 instead of 100/200
    milliseconds because they were already reached by simple function
    calls, causing the update timeout being hit instead of just entering
    the next period interval.

### DIFF
--- a/pkg/api/updates_test.go
+++ b/pkg/api/updates_test.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/api/updates_test.go
```diff
@@ -169,12 +169,12 @@ func TestGetUpdatePackage_MaxTimedOutUpdatesLimitReached_SafeMode(t *testing.T)
...
 	assert.NoError(t, err)
 
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(20 * time.Millisecond) // ensure that period interval is over but update timeout isn't
 
 	_, err = a.GetUpdatePackage(uuid.New().String(), "10.0.0.3", "12.0.0", tApp.ID, tGroup.ID)
 	assert.Equal(t, ErrMaxConcurrentUpdatesLimitReached, err)
 
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(510 * time.Millisecond) // ensure that update timeout is over (with the previous sleep time 490 would also work)
```
joaquimrocha, line 177: > with the previous sleep time 490 would also work
What does this mean? (Why do we need this comment?)

pkg/api/updates_test.go
```diff
@@ -203,12 +203,12 @@ func TestGetUpdatePackage_ResumeUpdates(t *testing.T) {
...
```
joaquimrocha, line 211: Same comment as before.

pkg/api/updates_test.go
pothos: This is just to clarify that the required waiting time is not 510 because we waited 20 ms already. I chose a small value for the first period for faster test execution but if someone needs to increase this then it would be unnecessary to wait the full update timeout interval.

joaquimrocha: Still not clear to me. Maybe an indication that the comment should be given more context or the part about 490 should eb removed.

pothos: I replaced the hardcoded values now with constants with variables and did a calculation here instead.

pkg/api/updates_test.go
```diff
@@ -123,7 +124,11 @@ func TestGetUpdatePackage_MaxUpdatesLimitsReached(t *testing.T) {
 	defer a.Close()
 
 	maxUpdatesPerPeriod := 2
-	periodInterval := "100 milliseconds"
+	periodInterval := "500 milliseconds"
+	periodIntervalD, err := time.ParseDuration(strings.ReplaceAll(periodInterval, " milliseconds", "ms"))
```
joaquimrocha, line 128: Sorry, I find these too hacky.
I know you have to use a string for the period interval, but what I meant is that you should have an integer `100`, and then you use this value for sleeping (multiplying by the time unit), and use it for the group definition too as a string, but in this case using `fmt.Sprintf("%d milliseconds", periodInterval)` makes things cleaner than replacing & parsing durations.

pothos: That's what I tried first but Go's time duration type conversion was too annoying. But I can give it another try with a different order of additions and multiplications.

pothos: Done

pkg/api/updates_test.go
```diff
@@ -157,24 +160,27 @@ func TestGetUpdatePackage_MaxTimedOutUpdatesLimitReached_SafeMode(t *testing.T)
...
 	assert.NoError(t, err)
 
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(periodInterval + buffer) // ensure that period interval is over but update timeout isn't
 
 	_, err = a.GetUpdatePackage(uuid.New().String(), "10.0.0.3", "12.0.0", tApp.ID, tGroup.ID)
 	assert.Equal(t, ErrMaxConcurrentUpdatesLimitReached, err)
 
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(updateTimeout - periodInterval + buffer) // ensure that update timeout is over
```
joaquimrocha, line 183: Isn't this the same as having just `updateTimeout`?
Also a nitpick here: I'd prefer something like `bufferPeriod` or `extraWaitPeriod`, just because seeing `buffer` makese me think we'll use a data buffer :slightly_smiling_face: (nitpick means it's up to you to change this if you agree)

pothos: In this case yes but not if `periodInterval` is changed. Yes, I'll give buffer a better name.

PR #245, opened 2020-07-10
 Add Kinvolk logo to container

ashu8912:


Dockerfile
```diff
@@ -19,6 +19,7 @@ RUN apk update && \
 
 COPY --from=nebraska-build /go/src/github.com/kinvolk/nebraska/bin/nebraska /nebraska/
 COPY --from=nebraska-build /go/src/github.com/kinvolk/nebraska/frontend/build/ /nebraska/static/
+COPY --from=nebraska-build /go/src/github.com/kinvolk/nebraska/default_logo.svg /nebraska/assets/
```
joaquimrocha, line 22: Let's call it kinvolk_logo (since it's not the default).

joaquimrocha:
I am trying a new approach, so let's pause this for a sec.

joaquimrocha:
We won't use this approach in the end. We have an alternative that won't need us to include Kinvolk's logo in the repo.
Thanks for the investigation though, that made clear what we needed to do.

PR #249, opened 2020-07-17
Date formatting

joaquimrocha:
This PR fixes the date formatting for channels and packages.

### DIFF
--- a/frontend/src/js/components/Channels/Item.react.js
+++ b/frontend/src/js/components/Channels/Item.react.js
@@ -49,26 +49,26 @@ function Item(props) {
 
       text += `(${ARCHES[channel.arch]})`;
     }
-    const date = new Date(channel.created_ts);
-    const calendarDate = `${date.getDate()}/${date.getMonth() + 1}`;
-    const time = `${date.getHours()}:${date.getMinutes()}`;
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/Channels/Item.react.js
```diff
@@ -49,26 +49,26 @@ function Item(props) {
...
+        {date &&
+          <Box pl={2}>
+            <Box display="flex">
+              <Box>
+                <Tooltip title="Release date">
+                  <ScheduleIcon fontSize="small"/>
+                </Tooltip>
+              </Box>
+              <Box pl={1}>
+                {date.toLocaleDateString()}
```
ashu8912, line 67: what if the date we recieve from above is null than we would be calling toLocaleDateString on null which will lead to app break 
we can replace this with date && date.toLocaleDateString()

ashu8912: and also why are we accessing created_ts from package it should be from channel

joaquimrocha: This is not correct: the patch's commit message states clearly that we need to show the package date, not the channel (visually it even shows that it's the version/package release date).
And it cannot be null because I am making the whole block depend on whether it's valid (line 58).

PR #250, opened 2020-07-17
Optimize get apps and add seperate request for get packages

ashu8912:


frontend/src/js/components/Packages/List.react.js
```diff
@@ -16,6 +17,8 @@ import Item from './Item.react';
 function List(props) {
   const [application, setApplication] =
     React.useState(applicationsStore.getCachedApplication(props.appID) || null);
+  const [packages, setPackages] = React.useState([]);
+  const [packageLoading, setPackageLoading] = React.useState(false);
```
joaquimrocha, line 21: You can use just packages if you follow the approach we have been using of making it a list or null. If it's null we show the spinner, if it's empty we know we've tried to fetch and didn't get any results.
With this patch's approach, the packages will only be continuously tried to be fetched if they're empty (in the other approach it will happen only if they haven't ever).

frontend/src/js/components/Packages/List.react.js
```diff
@@ -72,14 +81,14 @@ function List(props) {
       <Paper>
 
         <Box padding="1em">
-          {application ?
-            _.isEmpty(application.packages || []) ?
+          {application && !_.isEmpty(packages) ?
+            _.isEmpty(packages) ?
```
krnowak, line 85: Is the second `_.isEmpty(packages)` ever evaluated to `true` in this context? We evaluate `_.isEmpty(packages)` only if `!_.isEmpty(packages)` is `true` and the evaluation in this case should always result in `false`, right?

pkg/api/applications.go
```diff
@@ -167,12 +167,6 @@ func (api *API) GetApp(appID string) (*Application, error) {
 	} else {
 		return nil, err
 	}
-	packages, err := api.getPackages(app.ID)
-	if err == nil || err == sql.ErrNoRows {
-		app.Packages = packages
```
joaquimrocha, line 172: If we're not using `app.Packages` anywhere anymore, then this patch should also remove the entry from the struct.

frontend/src/js/components/Packages/List.react.js
joaquimrocha: Yeah. As Krzesimir says, the 2nd `_.isEmpty` is only reached if the previous one is true, so the latter is never `true`.
In fact, now that we have the `packages` as a list/null here, we can just check it unless we're still using the `application` within the block.

pkg/api/applications.go
```diff
@@ -21,7 +21,6 @@ type Application struct {
 	TeamID      string     `db:"team_id" json:"-"`
 	Groups      []*Group   `db:"groups" json:"groups"`
 	Channels    []*Channel `db:"channels" json:"channels"`
-	Packages    []*Package `db:"packages" json:"packages"`
```
joaquimrocha, line 24: I am a bit on the fence of how to treat this whole PR. After the previous commit, the frontend won't be able to show the packages correctly (until the last commit), and after this one the frontend will give an error (until the last commit).
Of course the PR gets merged as a whole but the motivation here is that a single commit shouldn't break functionality like this or give errors.

I think the best solution, which keeps things separated but unbroken is to have the frontend change first (so the functionality is not broken), and then the backend changes (squashing the Packages removal with the optimization work) as they will bring the optimization.

PR #254, opened 2020-07-19
Add scripts for easy running of local instances of the database and nebraska

krnowak:
There was no straightforward way of running the local database and nebraska since dockerfile.postgres was removed. So I've added 3 scripts - one for setting up the database (which the build system is now using), one for running a local no-auth nebraska instance, and one that does both of the above and a proper cleanup. Now checking if some thing works (during a PR review, for example) should basically be `git checkout …; make; ./tools/run-all.sh` and going to `localhost:8000` in a browser.

I used that to review #250.

### DIFF
--- a/tools/run-local-nebraska.sh
+++ b/tools/run-local-nebraska.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# This script runs a local instance of nebraska. It expects that both
+# frontend and backend are already built, and the database in expected
+# is already running. To override the database location nebraska is
+# using, specify the NEBRASKA_DB_URL environment variable.
+#
+# Whatever parameters passed to the script are forwarded to nebraska.
+
+set -euo pipefail
+
+tools_dir="$(dirname "${0}")"
+binary="${tools_dir}/../bin/nebraska"
+static_dir="${tools_dir}/../frontend/build"
+
+LOGXI=* \
+"${binary}" \
+    -auth-mode noop \
+    -http-log \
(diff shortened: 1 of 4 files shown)


tools/run-local-nebraska.sh
```diff
@@ -0,0 +1,17 @@
...
+
+tools_dir="$(dirname "${0}")"
+binary="${tools_dir}/../bin/nebraska"
+static_dir="${tools_dir}/../frontend/build"
+
+LOGXI=* \
+"${binary}" \
+    -auth-mode noop \
+    -http-log \
+    -http-static-dir "${static_dir}"
```
joaquimrocha, line 17: Nitpick: This could append any params give to the script at the end of this, so we could e.g. easily run it with a different DB address.

krnowak: Sure, I'll add it.

But! Currently the only way to specify the different DB address is through the `NEBRASKA_DB_URL` environment variable, so you can do `NEBRASKA_DB_URL=… ./tools/run-local-nebraska.sh`. But the `setup_local_db.sh` also assumes the default `5432` port. Maybe there should be an option for a database port, that defaults to `5432`. What do you think?

joaquimrocha: Doesn't the `NEBRASKA_DB_URL` include the port?

krnowak:
Updated. Made it possible to specify a custom port for database and used DOCKER_CMD in more places.

PR #255, opened 2020-07-21
Add nebraska version in ui

### LINKED ISSUE #144: Show Nebraska version in the UI
We should be able to see Nebraska's version in the UI (or accessible through the browser anyway) as it may be useful when deploying new versions and reporting bugs.

ashu8912:
Fixes #144.

### DIFF
--- a/frontend/src/js/components/Footer.js
+++ b/frontend/src/js/components/Footer.js
@@ -1,0 +1,15 @@
+import { Box } from '@material-ui/core';
+import React from 'react';
+import API from '../api/API';
+
+function Footer() {
+  const [nebraskaVersion, setNebraskaVersion] = React.useState(null);
+  React.useEffect(() => {
+    API.getConfig().then((config) => {
+      setNebraskaVersion(config.nebraska_version);
+    });
+  });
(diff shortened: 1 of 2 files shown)


ashu8912:
resolves #144

frontend/src/js/components/Footer.js
```diff
@@ -0,0 +1,16 @@
+import { Box } from '@material-ui/core';
+import React from 'react';
+import API from '../api/API';
+
+function Footer(){
+  const [nebraskaVersion, setNebraskaVersion] = React.useState(null);
+  React.useEffect(() => {
+    API.getConfig().then((config) => {
+      const version = config.nebraska_version ? config.nebraska_version.split('-')[0] : null;
```
joaquimrocha, line 9: I am not sure we should modify it here. Why don't you show the full version? (any `-dirty` suffix is the responsibility of how it's built, not the UI)

frontend/src/js/components/Footer.js
```diff
@@ -0,0 +1,16 @@
+import { Box } from '@material-ui/core';
+import React from 'react';
+import API from '../api/API';
+
+function Footer(){
```
joaquimrocha, line 5: Also, add a space between `)` and `{`.

PR #256, opened 2020-07-21
Format the updates policy text

### LINKED ISSUE #251: Format the updates policy text correctly
We have the wrong formating for the updates policy string in the application view as the following screenshot shows.

Screenshot from 2020-07-17 22-45-05

It should be done the same way as this [line](https://github.com/kinvolk/nebraska/blob/93f3ad371e4d8ce588d45af3c8cc0657641297e8/frontend/src/js/components/Groups/ItemExtended.react.js#L193).

ashu8912:
Fixes #251.

### DIFF
--- a/frontend/src/js/components/Groups/Item.react.js
+++ b/frontend/src/js/components/Groups/Item.react.js
@@ -144,7 +144,7 @@ function Item(props) {
               <CardFeatureLabel>Rollout Policy</CardFeatureLabel>
               <Box p={1} mb={1}>
                 <CardLabel>
-                  {`Max ${props.group.policy_max_updates_per_period || 0} updates per ${props.group.policy_period_interval || 0}`}
+                  {`Max ${props.group.policy_max_updates_per_period || 0} updates per ${props.group.policy_period_interval || '0 minutes'}`}
                 </CardLabel>
               </Box>
             </Grid>
--- a/frontend/src/js/components/Groups/Item.react.js
+++ b/frontend/src/js/components/Groups/Item.react.js
@@ -144,8 +144,7 @@ function Item(props) {
               <CardFeatureLabel>Rollout Policy</CardFeatureLabel>
               <Box p={1} mb={1}>
                 <CardLabel>
-                  Max {props.group.policy_max_updates_per_period}
-                  updates per {props.group.policy_period_interval}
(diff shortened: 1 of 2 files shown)


ashu8912:
fixes #251

frontend/src/js/components/Groups/Item.react.js
```diff
@@ -144,8 +144,7 @@ function Item(props) {
               <CardFeatureLabel>Rollout Policy</CardFeatureLabel>
               <Box p={1} mb={1}>
                 <CardLabel>
-                  Max {props.group.policy_max_updates_per_period}
-                  updates per {props.group.policy_period_interval}
+                  {`Max ${props.group.policy_max_updates_per_period || 0} updates per ${props.group.policy_period_interval || 0}`}
```
krnowak, line 147: This seems to follow what we have in `ItemExtended.react.js`, but in worst case we are going to have a message like `Max 0 updates per 0`, which does not make sense. Maybe it should be something like: 
```suggestion
                  {`Max ${props.group.policy_max_updates_per_period || 0} updates per ${props.group.policy_period_interval || '0 minutes'}`}
```

joaquimrocha: Let's make this right then. Let's see what the real numbers mean in terms of behavior in the backend and do the changes accordingly. If 0 is possible in both places, let's change the message to what it means (`No updates allowed` / `As many updates at the same time as possible`).

Probably we shouldn't allow 0 if it means no updates, as for that we already have a main switch.

frontend/src/js/components/Groups/Item.react.js
```diff
@@ -144,7 +144,7 @@ function Item(props) {
               <CardFeatureLabel>Rollout Policy</CardFeatureLabel>
               <Box p={1} mb={1}>
                 <CardLabel>
-                  {`Max ${props.group.policy_max_updates_per_period || 0} updates per ${props.group.policy_period_interval || 0}`}
+                  {`Max ${props.group.policy_max_updates_per_period || 0} updates per ${props.group.policy_period_interval || '0 minutes'}`}
```
joaquimrocha, line 147: Please see my previous comment.

joaquimrocha:
I fixed the formatting that this PR was originally related to.
I will open an issue so we address the `0 updates in 0 minutes` case.

PR #259, opened 2020-07-31
Retrieve time interval queries from the past closest X time

ashu8912:


pkg/api/groups.go
```diff
@@ -485,15 +486,18 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
 	if err != nil {
 		return nil, err
 	}
+	roundForValue := nearestXTimeFactor * time.Minute
+	currentTimeRoundedToNearestXMinuteDivision := time.Now().UTC().Round(roundForValue)
```
joaquimrocha, line 490: AFAIU tihs can be rounded up, can you clarify what maximum value in the future this may get us? Because if the maximum is now + ~5 mins, than we may have to make sure we only use values in the past.

ashu8912: yes so for timestamp like 10:00 to 10:12 it will be rounded to 10:10 and for 10:13 to 10:15  it will be rounded to 10:15

joaquimrocha: Right. That means that we'll not be showing 2 minutes of data because we know nothing is shown for dates in the future.
I think we should make sure this is only rounded down.

joaquimrocha:
Let's close this one for now until we have a better caching strategy (and when it's needed).

PR #264, opened 2020-08-11
Fix react and DOM errors

ashu8912:


frontend/src/js/components/Channels/Item.react.js
```diff
@@ -50,14 +50,14 @@ function Item(props) {
     }
     const date = channel.package ? new Date(channel.package.created_ts) : null;
     return (
-      <Box display="flex" ml={1}>
```
joaquimrocha, line 54: Wouldn't this (whole) patch not be necessary if we use `disableTypography` in the `ListItemText`? (I haven't verified, just a quick suggestion I remember seeing a while ago)

PR #269, opened 2020-08-22
Show concrete instance error

ashu8912:


pkg/api/instances.go
```diff
@@ -221,6 +223,12 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
 		if err != nil {
 			return nil, err
 		}
+		if instanceStatusHistoryEntity.Status == 3 {
```
pothos, line 226: Suggestion: Try to use `InstanceStatusError` instead of `3`

Same for line 282

frontend/src/js/constants/helpers.js
```diff
@@ -204,3 +209,113 @@ export function useGroupVersionBreakdown(group) {
 
   return versionBreakdown;
 }
+
```
pothos, line 212:
```suggestion
// Keep in sync with https://github.com/flatcar-linux/update_engine/blob/flatcar-master/src/update_engine/action_processor.h#L25
```

frontend/src/js/constants/helpers.js
```diff
@@ -204,3 +209,113 @@ export function useGroupVersionBreakdown(group) {
...
+  [Math.pow(2, 31)]: 'DevModeFlag',
+  [Math.pow(2, 30)]: 'ResumedFlag',
+  [Math.pow(2, 29)]: 'TestImageFlag',
+  [Math.pow(2, 28)]: 'TestOmahaUrlFlag'
+};
+
+export function getErrorsAndFlags(errorCode){
+  const errorMessages = [];
+  const flags = [];
+  let flag = Object.keys(flagsCodes).find((flagCode) => (errorCode & flagCode) !== 0);
```
pothos, line 271: I think it would be easier to first go through all the flags and detect them, and in a second step calculate errorCodeVal having it cleaned from all flags.

pothos:
Tested it and so far it worked great, thanks!

pothos:
Submitting a fake error makes the label empty:

```
$ cat step-1
<?xml version="1.0" encoding="UTF-8"?>
<request protocol="3.0" version="update_engine-0.4.10" updaterversion="update_engine-0.4.10" installsource="scheduler" ismachine="1">
    <os version="Chateau" platform="CoreOS" sp="2512.2.0_x86_64"></os>
    <app appid="e96281a6-d1af-4bde-9a0a-97b76e56dc57" version="1.2.3" track="stable" bootid="{965fb4c5-ad3e-4eb7-a4c2-ca0c0e31ec84}" oem="ami" oemversion="0.1.1-r1" alephversion="1688.5.3" machineid="2b9e671d61774703ac7be60715220bfe" lang="en-US" board="amd64-usr" hardware_class="" delta_okay="false" >
        <updatecheck></updatecheck>
    </app>
</request>
$ cat step-2
<?xml version="1.0" encoding="UTF-8"?>
<request protocol="3.0" version="update_engine-0.4.10" updaterversion="update_engine-0.4.10" installsource="scheduler" ismachine="1">
    <os version="Chateau" platform="CoreOS" sp="2512.2.0_x86_64"></os>
    <app appid="e96281a6-d1af-4bde-9a0a-97b76e56dc57" version="1.2.3" track="stable" bootid="{965fb4c5-ad3e-4eb7-a4c2-ca0c0e31ec84}" oem="ami" oemversion="0.1.1-r1" alephversion="1688.5.3" machineid="2b9e671d61774703ac7be60715220bfe" lang="en-US" board="amd64-usr" hardware_class="" delta_okay="false" >
        <event eventtype="3" eventresult="0" errorcode="999"></event>
    </app>
</request>
$ curl -d @step-1  -H 'Content-Type: text/xml' http://localhost:8000/v1/update
$ curl -d @step-2  -H 'Content-Type: text/xml' http://localhost:8000/v1/update
```

pothos:
Using `errorcode="267435474"` makes JavaScript go crazy until I kill the tab

pkg/api/instances.go
ashu8912: aah nice catch thanks

ashu8912:
@pothos how should fake error cases be dealt, I was an assumption that we will always be getting a valid error code

ashu8912:
fixed the getErrorsAndFlags function it returns no errors if a wrong error code is passed

ashu8912:
now in case of a wrong error_code we fallback to error message saying error updating

frontend/src/js/constants/helpers.js
```diff
@@ -204,3 +213,89 @@ export function useGroupVersionBreakdown(group) {
...
+export function getErrorsAndFlags(errorCode){
+  const errorMessage = [];
+  const flags = [];
+  let flag = Object.keys(flagsCodes).find((flagCode) => (errorCode & flagCode) !== 0);
+  let errorCodeVal = errorCode & ~flag;
+  if (flag) {
+    flags.push(flagsCodes[flag]);
+  }
+  //extract first all the flags from the error code
+  while (errorCodeVal > 3000){
```
pothos, line 282: This can loop endlessly. I'm no JS expert but I would do a for loop here. For each flag, test whether it's set and push it, then clean this flag.

frontend/src/js/constants/helpers.js
```diff
@@ -204,3 +213,84 @@ export function useGroupVersionBreakdown(group) {
...
+  for (;!!flag;) {
+    flags.push(flagsCodes[flag]);
+    errorCodeVal = errorCodeVal & ~flag;
+    flag = Object.keys(flagsCodes).find((flagCode) => (errorCodeVal & flagCode) !== 0);
+  }
+  if (actionCodes[errorCodeVal]) {
+    errorMessage.push(actionCodes[errorCodeVal]);
+  } else if (errorCodeVal > 2000 && errorCodeVal < 3000) {
+    errorMessage.push(`Http error code(${errorCodeVal - 2000})`);
+  }
```
pothos, line 288: Maybe add a third case where `errorCodeVal` is turned into a string `Unknown Error 123`

pothos: Then `errorMessage` doesn't need to be a list and the code in `getInstanceStatus` doesn't need to handle an empty list.

frontend/src/js/constants/helpers.js
pothos: An additional idea would be to have a nicer error name here like `Response empty` instead of `OmahaRequestEmptyResponseError` etc.

frontend/src/js/constants/helpers.js
pothos: Displaying the number for unknown errors may help if we add a new error in the future and Nebraska is not yet updated.

pothos: Thanks

pothos (approved):
Looks good now overall, just need to decide if we want to display `PayloadSizeMismatchError` as is or rename it here to a nicer name like `Payload size did not match` @joaquimrocha

frontend/src/js/components/Instances/Details.js
```diff
@@ -69,7 +69,15 @@ function StatusLabel(props) {
   const classes = useStatusStyles();
   const statusDefs = makeStatusDefs(useTheme());
   const {status, activated} = props;
-  const {icon = null, label = 'Unknown', color} = status && statusDefs[status.type] || {};
+  const statusObj = status && statusDefs[status.type] || {};
+  const {icon = null, color} = statusObj;
+  let label = 'Unknown';
+  console.log(statusObj);
```
joaquimrocha, line 75: Left over debug print.

joaquimrocha: Add a newline instead.

frontend/src/js/components/Instances/Details.js
```diff
@@ -69,7 +69,15 @@ function StatusLabel(props) {
...
+  if (statusObj.label === 'Error'){
```
joaquimrocha, line 76: Add a space between `)` and `{`.

joaquimrocha (changes requested):
Almost there. Please check the small comment I left + the changes we talked about on Slack.

frontend/src/js/__tests__/Activity/Container.spec.js
```diff
@@ -25,7 +26,11 @@ describe('Activity Container', () => {
       instance_id: null
     }];
 
-    const {getByText} = render(<Container/>);
+    const {getByText} = render(
+      <BrowserRouter>
+        <Container/>
+      </BrowserRouter>
```
joaquimrocha, line 32: Do you really need to use `BrowserRouter` here? Wouldn't it work without?
If it doesn't work without it, what are the implications of wrapping it around the base app instead of individual components?

joaquimrocha (changes requested):
Some comments, but it's getting there.

frontend/src/js/__tests__/Activity/Container.spec.js
ashu8912: So while testing or in normal cases as well => a component which uses components from react-router or react-router-dom its mandatory to wrap it in a Parent Router which can be either BrowserRouter or HashRouter

frontend/src/js/constants/helpers.js
```diff
@@ -204,3 +206,86 @@ export function useGroupVersionBreakdown(group) {
...
+  '34': 'OmahaResponseInvalid',
+  '35': 'OmahaUpdateIgnoredPerPolicy',
+  '36': 'OmahaUpdateDeferredPerPolicy',
+  '37': 'OmahaErrorInHTTPResponse',
+  '38': 'DownloadOperationHashMissingError',
+  '39': 'DownloadMetadataSignatureMissingError',
+  '40': 'OmahaUpdateDeferredForBackoff',
+  '41': 'PostinstallPowerwashError',
+  '42': 'NewPCRPolicyVerificationError',
+  '43': 'NewPCRPolicyHTTPError',
```
pothos, line 254: I have one more error code added in this (yet unmerged) PR: https://github.com/flatcar-linux/update_engine/pull/6

ashu8912: @pothos should i wait for it to merge?

pothos: No need to wait

ashu8912: added

pothos: ok^^ I meant it's ok to merge without but thanks for adding it, I will keep you posted

joaquimrocha:
@ashu8912 In the interest of saving time I removed the change of `Error update` -> `Error` because what we needed to do was not touch that label in a patch that's not about that. I kind of mixed things and had understood you had made `Error` -> `Error Update` but you were just capitalizing the word (and either change should not be in this commit).

PR #272, opened 2020-08-22
Fix instance list fetch

ashu8912:


joaquimrocha (changes requested):

joaquimrocha (changes requested):
Looking at the patch now, we're apparently just using the states (filters, page, rowsPerPage) for showing the values once. We could do away with that and make the `fetchFiltersFromURL` return an object with the things we're interested (status, version, page, etc.), and rename the `fetchFiltersFromURL` as `getParamsFromURL`.

I am fine either way, so I will leave it up to you. What I failed to ask before and I would like to see changed, is the `Fix instance list fetch` to something a bit more detailed as this is too generic and repeatable.
How about something like: `frontend: Fetch instance list directly from URL params` or `frontend: Avoid fetching instance list with wrong params initially` (not sure if it fits in <= 72 chars).

ashu8912:
Making URL the single source of truth is making the instance list view very slow to load, May be we can tackle this later, for now the above fix works fine.

joaquimrocha (approved):

Issue #273, opened 2020-08-24
Show the correct legend for the currently selected period in the version breakdown chart


joaquimrocha:
In the version breakdown chart, the default legend (no period selected, hence it's the last period) is always showing the values for the **1 day** interval, even if the user chooses a different interval.

**How to reproduce:**
1. Go to the group extended view and check the legend showing for the default interval (1 day);
2. Click on the `7 day` selector: the legend remains the same as before (and it's the same when selecting other intervals).

joaquimrocha:
Fixed in master.

PR #274, opened 2020-08-25
Fix default timeline table data display

ashu8912:


ashu8912:
@joaquimrocha as mentioned yesterday that the version_breakdown can be updated to accept duration query param but I realised later that we have to show the last time point from the version timeline by default and thus making a call to version_breakdown is an unnecessary call in that perspective, or if we do want to show by default for the entire duration than we need to change the message for version breakdown. But for now this looks like a good patch coz we already specify that we are showing for the last time point.

ashu8912:
fixes #273

joaquimrocha (changes requested):
Have some comments on the message:

Missing a comma after `only`:
```
If the version timeline chart is not clicked we were showing the
default timeline table for 1 day duration only even when a different
duration was selected.
```
This is confusing and missing a period:
``` Instead we should  by default show the last date
time point data for the selected duration
```
Should be something like.
```
Instead we should, by default, show the date for the last time point of the currently selected time interval.
```

I will test the patch now, as the code seems fine.

joaquimrocha (approved):

PR #277, opened 2020-08-28
Fix status timeline chart selection bug

ashu8912:


joaquimrocha (approved):
LGTM and I tested it.

PR #285, opened 2020-10-08
Syncer configuration

pothos:
- pkg/syncer/syncer: Add parameters for sync interval and source URL
    The interval was hardcoded to one hour which is inconvenient when new
    settings should be quickly propagated during a test phase. The sync
    URL was also hardcoded and required source code changes when a different
    source URL should has to be used.
    Expose both values as parameters with the current values as default.
- pkg/syncer/syncer: Reduce waiting time between two channels
    The syncer was sleeping for a minute before the next channel was
    synced. This makes it look like a bug and even if known it is not
    really useful but rather disturbing to wait 8 minutes until all
    channels are synced.
    Wait for 5 seconds which is a bit waiting time to not hammer the
    server but still allows to sync all channels in under a minute.

### DIFF
--- a/cmd/nebraska/controller.go
+++ b/cmd/nebraska/controller.go
@@ -52,6 +52,8 @@ type controllerConfig struct {
 	nebraskaURL         string
 	noopAuthConfig      *auth.NoopAuthConfig
 	githubAuthConfig    *auth.GithubAuthConfig
+	flatcarUpdatesURL   string
+	checkFrequency      time.Duration
 }
 
 func newController(conf *controllerConfig) (*controller, error) {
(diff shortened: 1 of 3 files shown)


cmd/nebraska/nebraska.go
```diff
@@ -44,6 +45,8 @@ var (
 	ghReadOnlyTeams     = flag.String("gh-ro-teams", "", "comma-separated list of read-only GitHub teams in the org/team format")
 	ghEnterpriseURL     = flag.String("gh-enterprise-url", "", fmt.Sprintf("base URL of the enterprise instance if using GHE; can be taken from %s env var too", ghEnterpriseURLEnvName))
 	logger              = log.New("nebraska")
+	flatcarUpdatesURL   = flag.String("sync-update-url", "https://public.update.flatcar-linux.net/v1/update/", fmt.Sprintf("Flatcar update URL to sync from"))
+	checkFrequencyVal   = flag.String("sync-interval", "1h", fmt.Sprintf("Sync check interval (the minimum depends on the number of channels to sync, e.g., 8m for 8 channels incl. different architectures)"))
```
joaquimrocha, line 49: I realize other lines are using the `Sprintf` function here, but that was probably a mistake and they can/should be plain strings, right?

cmd/nebraska/controller.go
```diff
@@ -52,6 +52,8 @@ type controllerConfig struct {
 	nebraskaURL         string
 	noopAuthConfig      *auth.NoopAuthConfig
 	githubAuthConfig    *auth.GithubAuthConfig
+	flatcarUpdatesURL   string
+	checkFrequency      time.Duration
```
joaquimrocha, line 56: Nitpick: The sync interval and source URL are independent, so this should be slipt into two commits.

joaquimrocha (changes requested):
Small comment.

cmd/nebraska/controller.go
pothos: Semantically yes, but due to the reformatting they wouldn't be independent from each other anyway and the first can't be reverted without conflicts.

cmd/nebraska/nebraska.go
pothos: Good catch, updated

joaquimrocha (approved):

PR #287, opened 2020-10-19
pkg/syncer/syncer: Extend list of synced channels with LTS

pothos:
The LTS channels for the different streams should be synced but not the
placeholder LTS channel used to manage the desired version in Nebraska.
Extend teh list of hard-coded synced channels with those that start
with "lts-" denoting an LTS stream channel. Ideally this would be a
boolean field on the channel itself, to set whether to sync or not.

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -140,7 +140,7 @@ func (s *Syncer) initialize() error {
 	}
 
 	for _, c := range flatcarApp.Channels {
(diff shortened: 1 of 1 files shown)


vbatts:
SGTM

joaquimrocha (approved):
Right. Ideally we'd get this as a setting per group, but it's fine to have this change ATM too.

vbatts:
I see this a draft, but it seems fine to merge.
And now @pothos is gone for two weeks :stuck_out_tongue_closed_eyes:

joaquimrocha:
When your drafts are good to be merged, then you must be a great engineer!

PR #289, opened 2020-10-23
Optimize update instance status

joaquimrocha:
This PR has the following optimizations affecting the `GetUpdatePackage` method:
* When updating the instance status, update `instance` + `instance_application` in the same query (instead of 2 queries);
* Don't update the `instance_application` twice when granting an update (it was setting the version + timestamp, etc. and then updating the status) which means 1 write instead of 2;
* Don't update the `instance_application` when an instance checks in if the last check in is less than 5 minutes ago (this reduces drastically the number of constant writes when using the `updateservicectl` fake instances which constantly ping the server)
* Don't fetch the group's stats (it's a count query that may get complex if we have lots of instances registered in the last 24h) unless we need them for enforcing the policy (i.e. if we allow more than 900000 parallel updates, then we don't need the stats because we're not using them for enforcing anything)

Here is a chart showing the number of writes with the latest stable release (throwing 2k instances pings every 0-10 seconds):
Screenshot from 2020-10-30 17-59-34

Here is the chart showing the same but with the new logic in this PR:
Screenshot from 2020-10-30 18-00-00

The CPU still reached 100% in both cases with this configuration in the very week machine I used:
(Old logic + the decline after I stopped the test case)
Screenshot from 2020-10-30 18-01-04

(New logic)
Screenshot from 2020-10-30 18-01-14

So even if we're still reaching 100% CPU and there is room for more optimization, this PR dramatically reduces the number of writes and that's a good thing.

### DIFF
--- a/pkg/api/activity.go
+++ b/pkg/api/activity.go
@@ -151,6 +151,53 @@ func (api *API) activityQuery(teamID string, p ActivityQueryParams) *goqu.Select
 	return query
 }
 
+func (api *API) hasRecentActivity(class int, p ActivityQueryParams) bool {
+	recent := time.Now().UTC().Add(-24 * time.Hour)
+
+	query := goqu.From("activity").
+		Select("*").
+		Where(goqu.C("class").Eq(class)).
+		Where(goqu.C("created_ts").Gt(recent))
+
+	if p.Severity != 0 {
+		query = query.Where(goqu.C("severity").Eq(p.Severity))
+	}
+
+	if p.Version != "" {
+		query = query.Where(goqu.C("version").Eq(p.Version))
+	}
+
+	if p.GroupID != "" {
+		query = query.Where(goqu.C("group_id").Eq(p.GroupID))
+	}
+
+	if p.AppID != "" {
+		query = query.Where(goqu.I("application_id").Eq(p.AppID))
+	}
+
+	if p.ChannelID != "" {
+		query = query.Where(goqu.I("channel_id").Eq(p.ChannelID))
+	}
+
+	if p.InstanceID != "" {
+		query = query.Where(goqu.C("instance_id").Eq(p.InstanceID))
+	} else {
+		query = query.Where(goqu.L(ignoreFakeInstanceCondition("instance_id")))
+	}
+
+	sql, _, err := query.ToSQL()
+	if err != nil {
+		return false
+	}
+
+	rows, err := api.db.Queryx(sql)
+	if err != nil {
+		return false
+	}
+	defer rows.Close()
+	return rows.Next()
+}
+
 // newGroupActivityEntry creates a new activity entry related to a specific
 // group.
 func (api *API) newGroupActivityEntry(class int, severity int, version, appID, groupID string) error {
--- a/pkg/api/instances.go
+++ b/pkg/api/instances.go
@@ -111,6 +111,66 @@ func (api *API) RegisterInstance(instanceID, instanceIP, instanceVersion, appID,
 	if appID, groupID, err = api.validateApplicationAndGroup(appID, groupID); err != nil {
 		return nil, err
 	}
+
+	// We want to avoid having to create an unneeded DB transaction, so we check whether it
+	// is necessary (we need it when writing into the two tables, instance and
+	// instance_application).
+
+	updateInstance := true
+	updateInstanceApplication := true
+
+	instance, err := api.GetInstance(instanceID, appID)
+	if err == nil {
+		// The instance exists, so we just update it if its IP changed
+		updateInstance = instance.IP != instanceIP
+
+		recent := nowUTC().Add(-5 * time.Minute)
+
+		// And we only update the instance_application if the latest registry is outdated or
+		// older than what we establish as recent.
+		updateInstanceApplication = instance.Application.LastCheckForUpdates.UTC().Before(recent) ||
+			instance.Application.Version != instanceVersion || instance.Application.GroupID.String != groupID
+
+		// Skip updating anything unnecessary
+		if !updateInstance && !updateInstanceApplication {
+			return instance, nil
+		}
+	}
+
+	upsertInstance, _, err := goqu.Insert("instance").
+		Cols("id", "ip").
+		Vals(goqu.Vals{instanceID, instanceIP}).
+		OnConflict(goqu.DoUpdate("id", goqu.Record{"id": instanceID, "ip": instanceIP})).
+		ToSQL()
+	if err != nil {
+		return nil, err
+	}
+
+	upsertInstanceApplication, _, err := goqu.Insert("instance_application").
+		Cols("instance_id", "application_id", "group_id", "version", "last_check_for_updates").
+		Vals(goqu.Vals{instanceID, appID, groupID, instanceVersion, nowUTC()}).
+		OnConflict(goqu.DoUpdate("ON CONSTRAINT instance_application_pkey", goqu.Record{"group_id": groupID, "version": instanceVersion, "last_check_for_updates": nowUTC()})).
+		ToSQL()
+	if err != nil {
+		return nil, err
+	}
+
+	// If we only have one table to update, then we just do that here directly and avoid the
+	// transaction below.
+	if updateInstance != updateInstanceApplication {
+		queryToExec := upsertInstance
+		if updateInstanceApplication {
+			queryToExec = upsertInstanceApplication
+		}
+		_, err := api.db.Exec(queryToExec)
+		if err != nil {
+			return nil, err
+		}
+
+		return instance, nil
+	}
+
+	// If this is an instance we haven't seen yet, then we write into instance + instance_application
 	tx, err := api.db.Begin()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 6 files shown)


pkg/api/instances.go
```diff
@@ -357,27 +363,33 @@ func (api *API) updateInstanceStatus(instanceID, appID string, newStatus int) er
...
+
+	const helperTableName = "inst_app"
+
+	// This insert is used with values returned from the update call, so we do one transaction
+	// in the DB only.
 	insertQuery, _, err := goqu.Insert("instance_status_history").
-		Cols("status", "version", "instance_id", "application_id", "group_id").
-		Vals(goqu.Vals{newStatus, lastUpdateVersion, instanceID, appID, groupID}).
+		Cols("status", "instance_id", "application_id", "version", "group_id").
+		FromQuery(goqu.From(goqu.L(helperTableName)).
```
vbatts, line 382: i wonder if some of these intermediary queries could get cached, with a TTL of like 30sec.
Many times, each host pinging is only interested in its own row, so if the sheet was only fresh within 30sec, and there had been other updates in that time, it would not have affected its own row.
This cache would save the bombardment of queries, if multiple callers could reuse the same cache.

joaquimrocha: Sure. Caching and queueing is a good strategy that it's in our plans, but we need to understand exactly what it means if we use the stored data for deciding on the policy, i.e. if you have "update only 200 instances every 15 minutes" and you use a 30-second valid cache for the data but you receive 1000 requests in 15 seconds, we cannot block them beyond 200 instances.

This is an example of why I want to take a step at a time. For other stuff not implied in the policy enforcing, for sure we could cache them.

pkg/api/instances.go
```diff
@@ -344,11 +390,19 @@ func (api *API) updateInstanceStatus(instanceID, appID string, newStatus int) er
 	if err != nil {
 		return err
 	}
+	return api.updateInstanceObjStatus(instance, newStatus)
+}
+
+func (api *API) updateInstanceData(instance *Instance, data *map[string]interface{}) error {
```
krnowak, line 396: There is no need to make `data` a pointer to map. Map types are pointer-like themselves, so passing them "by value" is cheap. So:
```suggestion
func (api *API) updateInstanceData(instance *Instance, data map[string]interface{}) error {
```

pkg/api/instances.go
```diff
@@ -357,30 +411,43 @@ func (api *API) updateInstanceStatus(instanceID, appID string, newStatus int) er
...
-	if err != nil {
-		return err
-	}
+
+	const helperTableName = "inst_app"
+
+	// This insert is used with values returned from the update call, so we do one transaction
+	// in the DB only.
 	insertQuery, _, err := goqu.Insert("instance_status_history").
 		Cols("status", "version", "instance_id", "application_id", "group_id").
```
krnowak, line 429: You could see https://pkg.go.dev/github.com/doug-martin/goqu/v9@v9.9.0#InsertDataset.With for the `WITH` support.

I suppose that it would look something like this:
```go
// we are not calling ToSQL() for updateQuery here!
updateQuery := goqu.Update("instance_application").YaddaYadda().Returning(…)
insertQuery, _, err := goqu.Insert("instance_status_history").
	With(helperTableName, updateQuery).
	FromQuery(…).
	Select(…).
ToSQL()
```

pkg/api/updates.go
```diff
@@ -5,7 +5,10 @@ import (
 	"time"
 
 	"github.com/blang/semver"
-	"github.com/doug-martin/goqu/v9"
+)
+
+const (
+	MaxParallelUpdates = 900000
```
krnowak, line 11: Where does this number come from?

krnowak: And does it need to be exported? (Starts with an upper case.)

pkg/api/updates.go
```diff
@@ -109,23 +112,18 @@ func (api *API) GetUpdatePackage(instanceID, instanceIP, instanceVersion, appID,
...
+	version := group.Channel.Package.Version
+
+	if err := api.grantUpdate(instance, version); err != nil {
 		logger.Error("GetUpdatePackage - grantUpdate error (propagates as ErrGrantingUpdate):", err)
-		return nil, ErrGrantingUpdate
 	}
 
-	if updatesStats.UpdatesToCurrentVersionGranted == 0 {
-		if err := api.newGroupActivityEntry(activityRolloutStarted, activityInfo, group.Channel.Package.Version, appID, group.ID); err != nil {
+	if !api.hasRecentActivity(activityRolloutStarted, ActivityQueryParams{Severity: activityInfo, AppID: appID, Version: version, GroupID: group.ID}) {
```
krnowak, line 125: I can assume that this is a cheaper equivalent of getting `updatesStats.UpdatesToCurrentVersionGranted`, right?

krnowak (changes requested):
Looks mostly good. Had some nitpicks and questions.

pkg/api/instances.go
```diff
@@ -111,6 +111,66 @@ func (api *API) RegisterInstance(instanceID, instanceIP, instanceVersion, appID,
...
+		Cols("instance_id", "application_id", "group_id", "version", "last_check_for_updates").
+		Vals(goqu.Vals{instanceID, appID, groupID, instanceVersion, nowUTC()}).
+		OnConflict(goqu.DoUpdate("ON CONSTRAINT instance_application_pkey", goqu.Record{"group_id": groupID, "version": instanceVersion, "last_check_for_updates": nowUTC()})).
+		ToSQL()
+	if err != nil {
+		return nil, err
+	}
+
+	// If we only have one table to update, then we just do that here directly and avoid the
+	// transation below.
```
pothos, line 159:
```suggestion
	// transaction below.
```

pkg/api/instances.go
joaquimrocha: I did test some combinations of `With` before and couldn't make it work with `Insert` but I guess I was not applying it correct because your suggestion works! Thanks!

pkg/api/instances.go
```diff
@@ -111,6 +111,66 @@ func (api *API) RegisterInstance(instanceID, instanceIP, instanceVersion, appID,
...
+		updateInstance = instance.IP != instanceIP
+
+		recent := nowUTC().Add(time.Duration(-5) * time.Minute)
+
+		// And we only update the instance_application if the latest registry is outdated or
+		// older than what we establish as recent.
+		updateInstanceApplication = instance.Application.LastCheckForUpdates.UTC().Before(recent) ||
+			instance.Application.Version != instanceVersion || instance.Application.GroupID.String != groupID
+
+		// Skip updating anything unnecessary
```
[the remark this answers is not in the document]
joaquimrocha: I don't think mine is incorrect. There were unnecessary things being updated. So we skip things that are unnecessary :slightly_smiling_face:

pkg/api/instances.go
joaquimrocha: This change gets me nervous, but you're coming from C++ too, so I will accept it.

pkg/api/activity.go
```diff
@@ -151,6 +151,52 @@ func (api *API) activityQuery(teamID string, p ActivityQueryParams) *goqu.Select
 	return query
 }
 
+func (api *API) hasRecentActivity(class int, p ActivityQueryParams) bool {
+	recent := time.Now().UTC().Add(time.Duration(-24*60) * time.Minute)
```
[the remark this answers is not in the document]
joaquimrocha: Thanks!

pkg/api/updates.go
joaquimrocha: It's not cheap against checking a boolean in memory, but it's cheap in most of the cases, especially when one is not limiting the updates, because we don't need to fetch the whole stats for the group (as it's a complex query).

pkg/api/updates.go
joaquimrocha: It's just an insanely huge number for checking updates in parallel, so instead of having a boolean check for whether we care for updates, I am using this number. We will limit to this number in the UI soon too.

joaquimrocha:
PR updated. Thanks for the reviews.

pkg/api/instances.go
```diff
@@ -111,6 +111,66 @@ func (api *API) RegisterInstance(instanceID, instanceIP, instanceVersion, appID,
 	if appID, groupID, err = api.validateApplicationAndGroup(appID, groupID); err != nil {
 		return nil, err
 	}
+
+	// We want to avoid having to create an unneeded DB transation, so we check whether it
```
krnowak, line 115:
```suggestion
	// We want to avoid having to create an unneeded DB transaction, so we check whether it
```

pkg/api/instances.go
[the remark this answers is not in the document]
krnowak: Yep, you are right. Let's leave it as is then.

krnowak (approved):
One typo, otherwise looks good.

pkg/api/instances.go
krnowak: Oh yeah, from that angle I can understand why it makes you nervous. Coming from that background, maps can have a bit weird semantics, see https://play.golang.org/p/RMG9kLd4d2L.

pkg/api/instances.go
joaquimrocha: Fixed. Thanks.

Issue #286, opened 2020-10-09
Add configurable Omaha track names for groups


pothos:
The mapping of Omaha track to a group ID was hardcoded for the stable,
    beta, alpha, and edge channel using only the initial the group IDs.
    When the groups were recreated resolving the track was broken and new
    groups always had to use the group ID as track which was confusing.
With https://github.com/kinvolk/nebraska/pull/284 we calculate the mapping dynamically based on the channel
architecture and the track name as the channel name, which is also what
a user would expect. However, this doesn't work when additional groups
are used for a single channel to specify alternative rollout policies or
disable updates. A configurable, unique name is needed for each group
and exposed in the UI. This can be implemented as a slug field for a group
or similar, and a migration to set it for the initial Flatcar groups.
Then the restriction to the hardcoded Flatcar app can also be lifted.

pothos:
Alternative suggestion: Replace the database type of the group ID to be a regular unique string that the user can change instead of the auto-generated UUID. If the database key shouldn't be touched, a new auto-generated one could be added.

pothos:
Replacing the ID is probably not a good idea and it should be kept as primary key, but the track name can be added as secondary key and most Go code would use it and it would be displayed in the UI instead of the current UUID. Just in case someone used the group UUID to refer to the channels, we can add a mapping of the known UUIDs to the new secondary track names.

PR #290, opened 2020-11-04
Add configurable group track names

pothos:
- Add configurable group track names:
    The mapping of Omaha track to a group ID was hardcoded for the stable,
    beta, alpha, and edge channel using only the initial the group IDs.
    When the groups were recreated, resolving the track was broken and then
    the groups always had to use the group ID as track which is confusing.
    For user-created groups there should be a way to add a track name.
    Add track name fields to the groups and expose them in the UI instead
    of the group ID. The edit and add dialogs allow to enter values but if
    omitted, the group ID will be used to populate the field. This is also
    done for existing entries through a migration. For requests which refer
    a non-existing track, a more clear error code is now returned. The test
    data needed some fixes for missing architecture information because the
    test requests used the amd64 architecture but the database defaults to
    the "all" architecture.

- Add cache for group track names and channel architectures:
    The lookup of the group track name and the channel architecture to find
    the group ID has to use an SQL JOIN statement each time which with many
    combinations could have a noticable performance impact.
    Cache the mapping of group track names and channel architectures to
    group IDs in a table and explicitly invalidate it when the group
    database entries are touched.

Closes https://github.com/kinvolk/nebraska/issues/286
Replaces https://github.com/kinvolk/nebraska/pull/284

I realize that the UI doesn't explain at all what these group UUIDs strings are. Since I added displaying the track name there, too, I think we could also rework this to have a proper label like `Group UUID and optional track name: …` but currently `Card.js` is shared between Applications and Groups.
Edit: At least the edit/add dialogs now have an explanation for the track and only the track is displayed instead of the UUIDs.


pothos:
The CI generates a different bindata Go file. Can somebody reproduce this?
(Such behavior is one reason why I don't like to have computer-generated stuff be committed…)

vbatts:
the Makefile for this has too many `.PHONY` and ought to do a little more for detecting changes...

anyways, when running `make backend && git diff` locally (with go1.15.3):

```patch
diff --git a/pkg/api/bindata.go b/pkg/api/bindata.go
index 987bea0..002ffc6 100644
--- a/pkg/api/bindata.go
+++ b/pkg/api/bindata.go
@@ -94,7 +94,7 @@ func dbDrop_all_tablesSql() (*asset, error) {
                return nil, err
        }
 
-       info := bindataFileInfo{name: "db/drop_all_tables.sql", size: 781, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+       info := bindataFileInfo{name: "db/drop_all_tables.sql", size: 781, mode: os.FileMode(0664), modTime: time.Unix(1, 0)}
        a := &asset{bytes: bytes, info: info, digest: [32]uint8{0xde, 0xcf, 0xbe, 0xfb, 0xd5, 0xe3, 0xea, 0xab, 0xec, 0x78, 0x43, 0x41, 0x2, 0x31, 0xfc, 0x52, 0x9a, 0x8b, 0x30, 0x84, 0xc8, 0xab, 0x83, 0x9c, 0x5b, 0x64, 0xf8, 0xbe, 0xbc, 0xfd, 0x0, 0x97}}
        return a, nil
 }
@@ -114,7 +114,7 @@ func dbSample_dataSql() (*asset, error) {
                return nil, err
        }
 
-       info := bindataFileInfo{name: "db/sample_data.sql", size: 15819, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+       info := bindataFileInfo{name: "db/sample_data.sql", size: 15819, mode: os.FileMode(0664), modTime: time.Unix(1, 0)}
        a := &asset{bytes: bytes, info: info, digest: [32]uint8{0x33, 0xc, 0x4c, 0x4c, 0x3, 0xb6, 0xaf, 0x7d, 0x14, 0x8b, 0xa6, 0x97, 0xe4, 0x22, 0xfd, 0xa3, 0x37, 0xdb, 0xb7, 0xf9, 0xa8, 0xbb, 0x94, 0xbe, 0x5e, 0xd3, 0x92, 0xab, 0xfe, 0xe5, 0x7e, 0x87}}
        return a, nil
 }
@@ -134,7 +134,7 @@ func dbMigrations0001_initialSql() (*asset, error) {
                return nil, err
        }
 
-       info := bindataFileInfo{name: "db/migrations/0001_initial.sql", size: 7125, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+       info := bindataFileInfo{name: "db/migrations/0001_initial.sql", size: 7125, mode: os.FileMode(0664), modTime: time.Unix(1, 0)}
        a := &asset{bytes: bytes, info: info, digest: [32]uint8{0x3e, 0x34, 0xf7, 0x7a, 0x71, 0x9b, 0x8b, 0x7b, 0xc6, 0x47, 0x8a, 0x1b, 0x3e, 0xb5, 0xb5, 0x2f, 0x29, 0xf5, 0xc3, 0xce, 0x5b, 0x2f, 0xc1, 0xd8, 0xdd, 0x23, 0xa0, 0x20, 0x34, 0x50, 0x6, 0x9f}}
        return a, nil
 }
@@ -154,7 +154,7 @@ func dbMigrations0002_event_dataSql() (*asset, error) {
                return nil, err
        }
 
-       info := bindataFileInfo{name: "db/migrations/0002_event_data.sql", size: 729, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+       info := bindataFileInfo{name: "db/migrations/0002_event_data.sql", size: 729, mode: os.FileMode(0664), modTime: time.Unix(1, 0)}
        a := &asset{bytes: bytes, info: info, digest: [32]uint8{0x93, 0xe9, 0x46, 0x83, 0xf5, 0x9c, 0x90, 0x5b, 0xa7, 0xd8, 0xdf, 0xb6, 0xb9, 0xf4, 0x9e, 0x1a, 0x6b, 0x5f, 0x66, 0x58, 0x61, 0xb9, 0x54, 0x3c, 0x93, 0x3b, 0xd9, 0x5b, 0xae, 0xf, 0x92, 0x8f}}
        return a, nil
 }
@@ -174,7 +174,7 @@ func dbMigrations0003_longer_team_namesSql() (*asset, error) {
                return nil, err
        }
 
-       info := bindataFileInfo{name: "db/migrations/0003_longer_team_names.sql", size: 142, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+       info := bindataFileInfo{name: "db/migrations/0003_longer_team_names.sql", size: 142, mode: os.FileMode(0664), modTime: time.Unix(1, 0)}
        a := &asset{bytes: bytes, info: info, digest: [32]uint8{0xea, 0xbb, 0x50, 0xf, 0x77, 0x25, 0x2e, 0xe6, 0x66, 0x6, 0x9c, 0xed, 0x44, 0xb2, 0x8a, 0xa6, 0x3c, 0xe7, 0xef, 0xb2, 0x49, 0x4, 0x81, 0x57, 0xc7, 0x6d, 0x91, 0x47, 0xe, 0xba, 0xb0, 0x57}}
        return a, nil
 }
@@ -194,7 +194,7 @@ func dbMigrations0004_rename_coreos_actionSql() (*asset, error) {
                return nil, err
        }
 
-       info := bindataFileInfo{name: "db/migrations/0004_rename_coreos_action.sql", size: 139, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+       info := bindataFileInfo{name: "db/migrations/0004_rename_coreos_action.sql", size: 139, mode: os.FileMode(0664), modTime: time.Unix(1, 0)}
        a := &asset{bytes: bytes, info: info, digest: [32]uint8{0x3c, 0x8b, 0xdd, 0x9e, 0x19, 0x1e, 0xdd, 0xa3, 0xd6, 0x9c, 0x2d, 0x5, 0x74, 0xe6, 0x9c, 0x1, 0xac, 0xde, 0x80, 0x9b, 0xa1, 0xb, 0xa8, 0xe2, 0xa4, 0x70, 0x48, 0x9b, 0xed, 0xe5, 0xae, 0xc6}}
        return a, nil
 }
@@ -214,7 +214,7 @@ func dbMigrations0005_default_team_idSql() (*asset, error) {
                return nil, err
        }
 
-       info := bindataFileInfo{name: "db/migrations/0005_default_team_id.sql", size: 333, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+       info := bindataFileInfo{name: "db/migrations/0005_default_team_id.sql", size: 333, mode: os.FileMode(0664), modTime: time.Unix(1, 0)}
        a := &asset{bytes: bytes, info: info, digest: [32]uint8{0x19, 0xdb, 0x8b, 0x9, 0xd, 0x3e, 0x9f, 0x64, 0xf9, 0x8f, 0x94, 0xe1, 0xf7, 0x95, 0x28, 0x12, 0xe7, 0xe6, 0x39, 0x30, 0xda, 0x5c, 0x68, 0x3d, 0x1f, 0xd0, 0xcd, 0x82, 0xd0, 0x8c, 0x4c, 0x52}}
        return a, nil
 }
@@ -234,7 +234,7 @@ func dbMigrations0006_initial_applicationSql() (*asset, error) {
[...]
```

vbatts:
and https://github.com/kinvolk/nebraska/pull/294 fixes that mode change

vbatts:
You can rebase on master now

pothos:
Thanks, ah, one of these errors again, good to have it fixed.

vbatts (commented):
this SGTM, but functionally I'll defer to @joaquimrocha for closer review

joaquimrocha:
I will review this tomorrow. Sorry I wwas busy today.

pkg/api/groups.go
```diff
@@ -135,6 +136,7 @@ func (api *API) AddGroup(group *Group) (*Group, error) {
 			group.PolicyPeriodInterval,
 			group.PolicyMaxUpdatesPerPeriod,
 			group.PolicyUpdateTimeout,
+			group.Track,
```
joaquimrocha, line 139: The group names cannot be duplicated: we should probably make this restriction per arch in the future.
The same should be applied to the `track`, as we should have duplicates but not for the same architecture.

pkg/omaha/omaha.go
```diff
@@ -106,16 +80,19 @@ func (h *Handler) buildOmahaResponse(omahaReq *omahaSpec.Request, ip string) (*o
...
+		// is optional and can be an empty string.
 		group := reqApp.Track
-		if reqAppUUID, err := uuid.Parse(reqApp.ID); err == nil {
-			if reqAppUUID.String() == flatcarAppID {
-				descriptor := chd(group, getArch(omahaReq.OS, reqApp))
-				if flatcarGroupID, ok := flatcarGroups[descriptor]; ok {
-					group = flatcarGroupID
-				}
-			}
+		groupID, err := h.crAPI.GetGroupID(group, getArch(omahaReq.OS, reqApp))
```
joaquimrocha, line 88: We're already doing many `GetGroup` calls but those are simple reads.
The `GetGroupID` however is performing a `JOIN` so I am affraid of the extra overhead for each request at scale. The alternative is to either cache the groups (which would benefit also other queries but we have to care of invalidating when the group or the channel changes), or to change some uses of the GetGroup to fetch it by `track`/ID depending on the usage (i.e. you'd pass the ID/track down but this gets easily complicated because we always have to do joins with the channel in order to get the arch).

Maybe the solution is to assign an arch to the group (coming from the channel when the channel is assigned to it) as sort of a cache. Not a big fan of the idea but we'd bypass the join this way too...

pkg/omaha/omaha_test.go
```diff
@@ -18,6 +18,8 @@ import (
 const (
 	defaultTestDbURL string = "postgres://postgres:nebraska@127.0.0.1:5432/nebraska_tests?sslmode=disable&connect_timeout=10"
 
+	flatcarAppID = "e96281a6-d1af-4bde-9a0a-97b76e56dc57"
```
joaquimrocha, line 21: It would also be nice to test the new `GetGroupID` or whatever new function you may use after my comments above.

frontend/src/js/components/Groups/EditDialog.js
```diff
@@ -133,6 +134,15 @@ function EditDialog(props) {
               fullWidth
             />
           </Box>
+          <Box mt={2} px={2}>
+            <Field
+              name="track"
+              component={TextField}
+              margin="dense"
+              label="Track"
+              fullWidth
```
joaquimrocha, line 143: This should have a placeholder or a small label below informing the user a track ID will be generated automatically if not filled.

joaquimrocha (changes requested):
The changes LGTM at an initial view but I raised some concerns in the comments that I think we need to address or at least discuss.

frontend/src/js/components/Common/Card.js
```diff
@@ -98,7 +98,7 @@ export function CardHeader(props) {
           </Grid>
           <Grid item xs={12}>
             <Box bgcolor={theme.palette.lightSilverShade} px={1} display="inline-block">
-              <Typography className={classes.idLabel} arial-label="group-id" noWrap>{props.cardId}</Typography>
+              <Typography className={classes.idLabel} arial-label="group-id" noWrap>{props.cardTrack ? props.cardId + ' (' + props.cardTrack + ')' : props.cardId}</Typography>
```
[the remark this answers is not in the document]
pothos: As the ID can still be used, it should have a place in the UI. We can add a migration for all non-default channels to copy the UUID over to the track name and forbid using the UUID for the default channels. But I didn't want to break this because I don't know if there are currently requests using the UUID.

pkg/omaha/omaha_test.go
pothos: It's covered by all test that use the Flatcar groups. We could expand the sample test data to cover ARM channels as well because the ARM architecture is not covered by tests (the test data and the initial data are different).

frontend/src/js/components/Groups/EditDialog.js
pothos: The track name is optional and the UUID can be used instead (in the current implementation). This explanation would be good to show, yes.

pkg/omaha/omaha.go
pothos: The other PR was written with minimal impact in mind but turned out to miss important cases if the channel name is used, and the slug was also not so nice until the group name suffix for architectures is removed.

How would I run a small benchmark? I don't expect the JOIN to be too significant because there are just a few groups and channels.

Fetching a group by track is actually what is done in the function, it just needs to have the additional information which architecture it is and that's encoded for the channel. Instead of adding this to the database I can try to combine the two PRs and add a cache here that uses the track name and the channel architecture.

pothos: Passing it down is only possible if we know what it is: either a UUID or a track name. If we forbid the usage of UUIDs we could spare the first lookup but I think it's easier to have a clear key for the follow-up functions…

frontend/src/js/components/Common/Card.js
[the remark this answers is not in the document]
joaquimrocha: What if we create a hardcoded check for the UUIDs in use. Sort of the currently track names check but for the UUID as a track name. We could then report somehow whether this is being used and remove it when we see it isn't anymore.
The main question is: is there a use-case for not using the track name (assuming we'll migrate UUID into the track name for existing groups, and generate a UUID for new groups if the user doesn't override the trackname)?

pkg/omaha/omaha.go
joaquimrocha: > The other PR was written with minimal impact in mind but turned out to miss important cases if the channel name is used, and the slug was also not so nice until the group name suffix for architectures is removed.
> 
> How would I run a small benchmark? I don't expect the JOIN to be too significant because there are just a few groups and channels.
> 
> Fetching a group by track is actually what is done in the function, it just needs to have the additional information which architecture it is and that's encoded for the channel. Instead of adding this to the database I can try to combine the two PRs and add a cache here that uses the track name and the channel architecture.

I can set up the staging server with a build from this PR and you can spawn a few thousand instances to stress it (or maybe try locally first) but I think it's still difficult to measure exactly the degration (maybe track in Postgres how much wait time there is).

I am leaning towards using the track name always and doing the migrations needed for it to work (hardcoded check for the default/known group's UUID + copying of the existing group UUIDs to the track column).

If you add a cache we'll be able to accelerate all reads that request groups' infos anyway (and we could add the migration later, since both things make sense), so I am also fine with that and it may be a simpler solution for now.

pkg/omaha/omaha_test.go
joaquimrocha: Ok. It'd be nice to have other arch tests, but that's beyond the scope of this PR, so don't feel obliged to do it.

frontend/src/js/components/Groups/EditDialog.js
joaquimrocha: My understanding is that we shouldn't expose both to the user (unless I am missing a use-case).
I think it's much simpler to just call it group ID or group track name, but havnig both seems confusing and unnecessary.

We can think of the track name as a mutable ID, right? And how we call it or handle it is an internal implementation detail.

frontend/src/js/components/Common/Card.js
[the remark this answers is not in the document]
pothos: I thought about the hard-coded reverse mapping and a warning in the log but once it's there it's probably staying forever ;)

frontend/src/js/components/Groups/EditDialog.js
pothos: Done, now only the track is shown and it is filled with the group ID if it was empty.

pkg/omaha/omaha_test.go
pothos: I had to add fix the architecture values of the test data anyway but they still only cover amd64.

pkg/omaha/omaha.go
pothos: I reworked the PR now to use the track name and have it filled through a migration. But since it's not a unique key I still needed to lookup the UUID and don't think we should change all functions.

Adding the cache now is not much work but I didn't do it in this iteration and think we should try to get some numbers first.

pothos: Ok, added the cache in an additional commit, so that we can revert this easily if we don't want it anymore.

pkg/api/groups.go
```diff
@@ -270,26 +292,61 @@ func (api *API) GetGroup(groupID string) (*Group, error) {
...
+	cachedGroupsLock.RLock()
+	if cachedGroups != nil {
+		// Keep a reference to the map that we found.
+		cachedGroupsRef = cachedGroups
+	}
+	cachedGroupsLock.RUnlock()
+	// Generate map on startup or if invalidated.
+	if cachedGroupsRef == nil {
+		cachedGroupsLock.Lock()
+		cachedGroups := make(map[GroupDescriptor]string)
```
joaquimrocha, line 305: Should we check the value of the `cachedGroups` here again?
Maybe I imagine that since we're performing the query inside the lock, if it takes time (and thus makes other calls wait for the lock here), if the other "threads" enter this part, maybe the value of `cachedGroups` changed from the previous check and we'd be performing unnecessary work (i.e. repeating the query)?

joaquimrocha (commented):
Thanks Kai!
Left a small comment WRT how we're handling the locked code. Ping me on Slack if I am missing something. I will test this now.

pkg/omaha/omaha.go
```diff
@@ -8,38 +8,25 @@ import (
...
-
-		chd("alpha", api.ArchAArch64):  "e641708d-fb48-4260-8bdf-ba2074a1147a",
-		chd("beta", api.ArchAArch64):   "d112ec01-ba34-4a9e-9d4b-9814a685f266",
-		chd("stable", api.ArchAArch64): "11a585f6-9418-4df0-8863-78b2fd3240f8",
-		chd("edge", api.ArchAArch64):   "b4b2fa22-c1ea-498c-a8ac-c1dc0b1d7c17",
+	initialFlatcarGroups = map[string]string{
+		// amd64
+		"5b810680-e36a-4879-b98a-4f989e80b899": "stable",
+		"3fe10490-dd73-4b49-b72a-28ac19acfcdc": "beta",
+		"9a2deb70-37be-4026-853f-bfdd6b347bbe": "alpha",
```
joaquimrocha, line 23: I tested and just noticed that `stable` is switched here with `alpha` in comparison to before (c&p mistake I'm sure).

joaquimrocha (changes requested):
Noticed a bug.

pkg/api/groups.go
pothos: It could be double work sometimes but also covers the case where the cache was invalidated again while it wasn't even completely built yet.

pkg/omaha/omaha.go
pothos: Good, updated :)

joaquimrocha (approved):
LGTM now. Thanks @pothos .

PR #304, opened 2020-11-23
db/migrations: Unify group default settings with public instance

pothos:
The defaults were varying between the groups and too limiting which
caused confusion when after one request no updates were served anymore.
Also the Safe Mode was set which does not behave as supposed because
when the updates are finally turned off by the Safe Mode they can't be
activated again as the faulty instance is still considered.
Adjust the defaults, affecting only new installations.

# How to use

# Testing done

Verified that running the new binary does not execute the migrations again. Verified that for an empty database the new settings are used.


joaquimrocha (approved):
LGTM.

PR #307, opened 2020-11-23
frontend: Show the full date in the instance details view

### LINKED ISSUE #299: Dates are shown like 8 Tue without a month (and year) in the event timeline
At least the month should be added to the format in which the date and time is displayed in the event timeline of an instance.

joaquimrocha:
The instances' details view was showing the date of the last update
check without the month information, regardless of whether we were in
the same month currently as the date in question. This patch fixes that
by showing the full numeric date. We can add a more human-friendly date
format later if needed.

fixes #299


ashu8912 (approved):
Tested, LGTM

PR #308, opened 2020-11-25
Remove generated JSON logos and exclude them from git

pothos:
After "make" this file gets modified.


ashu8912 (commented):
@pothos I think we don't want to commit this file because it's generated based on the logo we have we might in future want to change the logos, I think the entire frontend/src/js/icons folder shouldn't be committed as it's always generated when we start the server.
So I think the better solution will be to add the dir frontend/src/js/icons to .gitignore

pothos:
Ok, updated it

joaquimrocha (approved):

PR #310, opened 2020-11-25
Add instance alias

joaquimrocha:
We want users to be able to rename instances in order to more easily identify them.

This PR adds a new `alias` column to the instance table, the respective update endpoint, and the UI changes.


pothos:
I added the Omaha extension on top of this here: https://github.com/kinvolk/nebraska/commit/c93f1d906bac60b3339371d7cf41c23331c7d3ff (`kai/add_instance_alias-plus-omaha `)

pothos (approved):
From my point of view this is ready

PR #320, opened 2021-01-08
Enhance application metrics and Add http endpoint level metrics 

yolossn:
In the existing implementation, the metrics are calculated when the metrics endpoint is called. This can be a problem in 2 scenarios.

1. Since the instrumentation is done in the request-response cycle it has a possibility of timing out if the number of channels/version/application increases.
2. Prometheus has a default scrape interval of 15s and I doubt if these metrics need to be that granular. If not, we are putting unwanted pressure on the db.

This implementation uses Prometheus `client_golang` package and separates the instrumentation logic from the request-response lifecycle. Also adds endpoint level metrics which will be useful to identify the endpoints which doesn't meet the performance requirement.


cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// getMetricsRefreshInterval returns the metrics update Interval key is set in the environment as time.Duration,
+// If not returns the default update interval.
+func getMetricsRefreshInterval() time.Duration {
+	refreshInterval, err := strconv.Atoi(os.Getenv("NEBRASKA_METRICS_UPDATE_INTERVAL"))
```
krnowak, line 58: How about using `time.ParseDuration`? That way you need to pass a time unit in the env var (like `5s` or `1h`).

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
```
krnowak, line 59: I think that you would still need to check if the duration is positive if there was no problem parsing the number

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	err = prometheus.Register(failedUpdatesGaugeMetric)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// getMetricsRefreshInterval returns the metrics update Interval key is set in the environment as time.Duration,
+// If not returns the default update interval.
```
krnowak, line 56:
```suggestion
// getMetricsRefreshInterval returns the metrics update interval. It takes it from environment variable
// NEBRASKA_METRICS_UPDATE_INTERVAL. The variable must be a string acceptable by time.ParseDuration
// and represent a positive duration. Otherwise returns a default interval.
```

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	defaultMetricsUpdateInterval = time.Second * time.Duration(5)
```
krnowak, line 12: This should be fine too:

```suggestion
	defaultMetricsUpdateInterval = 5 * time.Second
```

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+
+// getMetricsRefreshInterval returns the metrics update Interval key is set in the environment as time.Duration,
+// If not returns the default update interval.
+func getMetricsRefreshInterval() time.Duration {
+	refreshInterval, err := strconv.Atoi(os.Getenv("NEBRASKA_METRICS_UPDATE_INTERVAL"))
+	if err != nil {
+		return defaultMetricsUpdateInterval
+	}
+
+	return time.Second * time.Duration(refreshInterval)
```
krnowak, line 63: If you use `ParseDuration`, then you'll get a `time.Duration` from it, so you could just do `return refreshInterval`.

cmd/nebraska/metrics.go
krnowak: Also, in case of an invalid duration, maybe log a warning? The variable being an empty string should be handled silently, though.

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+func registerAndInstrumentMetrics(ctl *controller) error {
+	// register application metrics
+	err := registerNebraskaMetrics()
+	if err != nil {
+		return err
+	}
+
+	refreshInterval := getMetricsRefreshInterval()
+
+	metricsTicker := time.NewTicker(refreshInterval)
```
krnowak, line 76: You could use `metricsTickChannel := time.Tick(refreshInterval)`, because you are not using any functionality of `time.Ticker`.

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return nil
+}
+
+// calculateMetrics calculates the application metrics and updates the respective metric.
+func calculateMetrics(ctl *controller) error {
+	aipcMetrics, err := ctl.api.GetAppInstancesPerChannelMetrics()
+	if err != nil {
+		logger.Error("getMetrics - getting app instances per channel metrics", "error", err.Error())
+		return err
```
krnowak, line 96: I'd just return a descriptive error only - it's going to be logged by the caller of `calculateMetrics`.

```suggestion
		return fmt.Errorf("failed to get app instances per channel metrics: %w", err)
```

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	for _, metric := range aipcMetrics {
+		appInstancePerChannelGaugeMetric.WithLabelValues(metric.ApplicationName, metric.Version, metric.ChannelName).Set(float64(metric.InstancesCount))
+	}
+
+	fuMetrics, err := ctl.api.GetFailedUpdatesMetrics()
+	if err != nil {
+		logger.Error("getMetrics - getting failed updates metrics", "error", err.Error())
+		return err
```
krnowak, line 106: Same here, could you return a descriptive error here instead?

krnowak (changes requested):
Found some small things to fix. But looks ok overall.

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+// NEBRASKA_METRICS_UPDATE_INTERVAL. The variable must be a string acceptable by time.ParseDuration
+// If not returns the default update interval.
+func getMetricsRefreshInterval() time.Duration {
+	refreshIntervalEnvValue := os.Getenv("NEBRASKA_METRICS_UPDATE_INTERVAL")
+	if refreshIntervalEnvValue == "" {
+		return defaultMetricsUpdateInterval
+	}
+
+	refreshInterval, err := time.ParseDuration(refreshIntervalEnvValue)
+	if err != nil {
```
krnowak, line 64: Checking for negative duration is missing.
```suggestion
	if err != nil || refreshInterval <= 0 {
```

cmd/nebraska/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+		logger.Warn("invalid NEBRASKA_METRICS_UPDATE_INTERVAL value, it must be acceptable by time.ParseDuration")
```
krnowak, line 65: Consider printing the invalid value too.

```suggestion
		logger.Warn("invalid NEBRASKA_METRICS_UPDATE_INTERVAL value, it must be acceptable by time.ParseDuration and positive", "value", refreshIntervalEnvValue)
```

krnowak (changes requested):
Two nits, with them fixed, should be good to go.

cmd/nebraska/metrics.go
krnowak: Or consider having a separate `if` for checking for negative value with an appriopriate error message.

krnowak (approved):

PR #321, opened 2021-01-11
Optimisations for slow queries in instance_application table

yolossn:
From the RDS performance insights, the query on instance_application table is slow. The composite index and removal of extra query parameters will boost the performance. See the attached pictures for the difference in query planner.

Query: 
---
SELECT "version", "status", "last_check_for_updates", "last_update_version", "update_in_progress", "application_id", "group_id" FROM "instance_application" WHERE (("instance_id" = ?) AND ("application_id" = ?) AND last_check_for_updates > now() at time zone ? - interval ? AND (instance_id IS NULL OR instance_id NOT SIMILAR TO ?))

Before composite index
image

After composite index
image
---

### DIFF
--- a/pkg/api/applications.go
+++ b/pkg/api/applications.go
@@ -147,7 +147,7 @@ func (api *API) DeleteApp(appID string) error {
 // GetApp returns the application identified by the id provided.
 func (api *API) GetApp(appID string) (*Application, error) {
 	var app Application
-	query, _, err := api.appsQuery().
+	query, _, err := goqu.From("application").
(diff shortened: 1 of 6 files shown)


joaquimrocha (approved):
The changes LGTM, but I'd like to ask whether you'd looked into all the places where those sorted queries in the last commit are being used, in order to see if we could skip some of them.

yolossn:
> The changes LGTM, but I'd like to ask whether you'd looked into all the places where those sorted queries in the last commit are being used, in order to see if we could skip some of them.

Did another round of check and made changes

PR #332, opened 2021-02-15
Add prettier

### LINKED ISSUE #326: Add prettier
**Current situation**
We're relying on eslint for the coding style, and stule issues are also caught by CI, but in order to reduce some back and forth, we should add prettier support.

**Impact**
Users will have autoomatic formating of code without having to run `make frontend-fixlint`.

ashu8912:
Just like we have prettier configured for headlamp we should do the same with nebraska 
fixes #326

### DIFF
--- a/frontend/.eslintrc.yml
+++ b/frontend/.eslintrc.yml
@@ -1,1 +1,0 @@
-extends: '@kinvolk'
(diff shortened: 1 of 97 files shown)


joaquimrocha:
I think this should be added after the Typescript PR, otherwise you'll have to change each commit.

PR #336, opened 2021-02-17
backend: Fix overwriting of accesslevel in github auth

### LINKED ISSUE #335: Users in the view and admin team cannot administer Nebraska
**Description**

If a user is in the admin and view teams, they should be able to get the admin permissions and not simply the view ones.

**Impact**

Users will be expect to manage Nebraska as they are in the admin team but the view team permissions seem to take priority.

**Environment and steps to reproduce**

1. Add a user to the admin and view teams.
2. Have the user log in and modify/add something to Nebraska.

**Expected behavior**
The user should be able to e.g. add a group.

Note: I am not 100% certain this is the view team getting priority, or a side effect of #334 (i.e. if the user logged in when they had only the view team permissions, and then log in with the admin ones, this 334 bug is making the view permissions persist).

yolossn:
Right now setting the accesslevel sets the value without
checking if the value already exists, this leads to ro
overwriting rw accesslevel. A check is added to prevent it.

fixes #335

### DIFF
--- a/cmd/nebraska/auth/githubauth.go
+++ b/cmd/nebraska/auth/githubauth.go
@@ -348,6 +348,10 @@ func (gha *githubAuth) doLoginDance(c *gin.Context, oauthClient *http.Client) (r
 		Page:    1,
 		PerPage: 50,
 	}
+	isRO := false
+	isRW := false
+
+checkLoop:
 	for {
 		ghTeams, response, err := client.Teams.ListUserTeams(c.Request.Context(), &listOpts)
 		if err != nil {
@@ -373,11 +377,15 @@ func (gha *githubAuth) doLoginDance(c *gin.Context, oauthClient *http.Client) (r
 			fullGithubTeamName := makeTeamName(*ghTeam.Organization.Login, *ghTeam.Name)
 			logger.Debug("login dance", "trying to find a matching ro or rw team", fullGithubTeamName)
 			for _, roTeam := range roTeams {
+				if isRO {
+					break
+				}
 				if fullGithubTeamName == roTeam {
 					logger.Debug("login dance", "found matching ro team", fullGithubTeamName)
 					teamData.org = *ghTeam.Organization.Login
 					teamData.team = ghTeam.Name
 					teamID = gha.defaultTeamID
+					isRO = true
 					session.Set("accesslevel", "ro")
 					break
 				}
@@ -388,13 +396,11 @@ func (gha *githubAuth) doLoginDance(c *gin.Context, oauthClient *http.Client) (r
 					teamData.org = *ghTeam.Organization.Login
 					teamData.team = ghTeam.Name
 					teamID = gha.defaultTeamID
+					isRW = true
 					session.Set("accesslevel", "rw")
-					break
+					break checkLoop
 				}
 			}
-		}
-		if teamID != "" {
-			break
 		}
 		// Next page being zero means that we are on the last
 		// page.
@@ -403,9 +409,10 @@ func (gha *githubAuth) doLoginDance(c *gin.Context, oauthClient *http.Client) (r
 		}
 		listOpts.Page = response.NextPage
 	}
-	if teamID == "" {
+	if !isRW {
 		logger.Debug("login dance", "no matching teams found, trying orgs")
 		listOpts.Page = 1
+	checkLoop2:
 		for {
 			ghOrgs, response, err := client.Organizations.List(c.Request.Context(), "", &listOpts)
 			if err != nil {
@@ -422,10 +429,14 @@ func (gha *githubAuth) doLoginDance(c *gin.Context, oauthClient *http.Client) (r
 				logger.Debug("login dance", "trying to find a matching ro or rw team", *ghOrg.Login)
 				nebraskaOrgName := *ghOrg.Login
 				for _, roTeam := range roTeams {
+					if isRO {
+						break
+					}
 					if nebraskaOrgName == roTeam {
 						logger.Debug("login dance", "found matching ro team", nebraskaOrgName)
 						teamData.org = nebraskaOrgName
 						teamID = gha.defaultTeamID
+						isRO = true
 						session.Set("accesslevel", "ro")
 						break
 					}
@@ -436,12 +447,9 @@ func (gha *githubAuth) doLoginDance(c *gin.Context, oauthClient *http.Client) (r
 						teamData.org = nebraskaOrgName
 						teamID = gha.defaultTeamID
 						session.Set("accesslevel", "rw")
-						break
+						break checkLoop2
 					}
 				}
-			}
-			if teamID != "" {
-				break
 			}
 			// Next page being zero means that we are on the last
 			// page.


krnowak:
So I suppose that the problem this PR seeks out to solve is the situation where a user belongs to both read-only team and read-write team (so the user should generally have a read-write access level), and the access level the user ends up with, depends on the order of teams we received from github. If the list of teams is `[read-only-team, read-write-team`] the user will have read-write access level. And on reverse order, the user will have read-only access level, which is clearly wrong.

With that in mind, may I suggest a different solution? I'd propose putting a label (let's call it `checkLoop1`) on this `for` loop in line 351:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L348-L353

Then change `break` to `break checkLoop` in line 392: https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L389-L394

That would mean that once we get a read-write access-level (and you can't get any better than that), we stop checking the teams completely. Similar changes (with label like `checkLoop2`) needs to be done also below, in line 409:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L407-L411

And changing `break` to `break checkLoop2` in line 439:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L436-L440

And while we are here, I'd add two more fixes. First one would be to remember if we are already in a read-only team, so we don't need to check that again. I'd add a `isRO := false` variable somewhere before the `for` loop we just labeled with `checkLoop1`, and add
```go
if isRO {
	break
}
``` 
before the `if fullGithubTeamName == roTeam {` in:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L373-L377
And in:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L423-L427

(That way we won't have a six or even seven levels of indentation in the function as five is already ludicrous and I'm to blame here. But you could also put the `for _, roTeam := range roTeams` inside the `if !isRO`, I don't mind.)

And then add `isRO = true` somewhere in:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L377-L381
And in:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L426-L429

(Instead of having `isRO` you probably could check the session directly if access level is already `ro`, I don't mind.)

And the second fix would be to drop these `if`s completely, they are just wrong:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L395-L399
And:
https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L442-L446

This will trip up when the user belongs to more than 50 teams (say 64), and then we will get two batches of teams. First batch of 50 teams will contain a read-only team, second batch of 14 teams will contain a read-write team. With this `if` in place, then second batch won't be considered at all, because we will already get a nonempty `teamID`, because there was some read-only team. With that, the user will have a read-only access level instead of a read-write one.

There is a bunch of code duplication going on here, and this code screams for a refactor, but for this PR I'd rather focus on just fixing bugs. Refactoring can happen some other time.

yolossn:
> With that in mind, may I suggest a different solution? I'd propose putting a label (let's call it `checkLoop1`) on this `for` loop in line 351:
> https://github.com/kinvolk/nebraska/blob/27b3a4c7d7cf4f2e4d475c68a1216710eed0af13/cmd/nebraska/auth/githubauth.go#L348-L353
> 
> Then change `break` to `break checkLoop` in line 392:

Awesome suggestions was discussing the loop break solution with @joaquimrocha. The `isRO` will help reduce redundant checks, will Include these.

cmd/nebraska/auth/githubauth.go
```diff
@@ -378,7 +378,9 @@ func (gha *githubAuth) doLoginDance(c *gin.Context, oauthClient *http.Client) (r
 					teamData.org = *ghTeam.Organization.Login
 					teamData.team = ghTeam.Name
 					teamID = gha.defaultTeamID
-					session.Set("accesslevel", "ro")
+					if !session.Has("accesslevel") {
+						session.Set("accesslevel", "ro")
+					}
```
joaquimrocha, line 383: So, what may be happening is we go through a team that's the read-write one first, and only then go through the read-only team which ends up reassigning the accesslevel, right?

In that case, I think we should rather stop the whole loop once a read-write team is found. It's more efficient, and we don't prevent reassigning the seccion's accesslevel with read-only if the user is removed from the read-write team and is left only in the read-only team (which I think may happen if we add this logic).

joaquimrocha (changes requested):

krnowak:
This looks good, thanks!

But I just found one more issue there. Basically the access level can depend on whether a user belongs to some team or to some organization (hence the duplication). So it might be possible that being a member of some team gives the user a read-only access, but membership to some organization may grant you a read-write access. So in current situation, when we find that user is a member of a read-only team, we assign some value to the `teamID` variable, which causes the function to not even consider the organizations any more, so the user won't be granted the read-write access level.

I think it could be fixed by adding a `isRW := false` variable just next to its `isRO` counterpart here:
https://github.com/kinvolk/nebraska/blob/44061e5573dcdfe09bbeeab36a4cf4bb3d39e501/cmd/nebraska/auth/githubauth.go#L349-L353

Then assigning `true` to it somewhere here:
https://github.com/kinvolk/nebraska/blob/44061e5573dcdfe09bbeeab36a4cf4bb3d39e501/cmd/nebraska/auth/githubauth.go#L393-L399

And replacing this `teamID == ""` with `if !isRW` here:
https://github.com/kinvolk/nebraska/blob/44061e5573dcdfe09bbeeab36a4cf4bb3d39e501/cmd/nebraska/auth/githubauth.go#L409-L413

I think all these issue have a root at the fact that this code initially gave either no access at all or read-write access. I think that later I bolted on a read-only access, but haven't thought it out very carefully, hence the issues like that cropping up. Sorry about that.

With that fixed, it should be good to go.

krnowak:
And I think that the commit message will need an update.

yolossn:
> And I think that the commit message will need an update.

Have made the changes and updated the commit message.

krnowak (approved):
Thanks!

PR #337, opened 2021-02-22
backend: Drop all user sessions on gh team addition

yolossn:
Right now if the user is added to the rw team the
active sessions are not deleted and the access level
remains same. The fix handles the added github
webhook event and drops all the active sessions of
the user.

### DIFF
--- a/cmd/nebraska/auth/githubauth.go
+++ b/cmd/nebraska/auth/githubauth.go
@@ -678,22 +678,38 @@ func (gha *githubAuth) loginWebhookMembershipEvent(c *gin.Context, payloadReader
 		return
 	}
 	logger.Debug().Str("got membership event with action", payload.Action).Msg("webhook")
-	if payload.Action != "removed" {
-		logger.Debug().Str("ignoring membership event with action", payload.Action).Msg("webhook")
-		return
-	}
-	logger.Debug().Str("got membership remove event with scope", payload.Scope).Msg("webhook")
+
+	logger.Debug().Str("got membership event with scope", payload.Scope).Msg("webhook")
 	if payload.Scope != "team" {
-		logger.Debug().Str("ignoring membership remove event with scope", payload.Scope).Msg("webhook")
+		logger.Debug().Str("ignoring membership event with scope", payload.Scope).Msg("webhook")
 		return
 	}
 	username := payload.Member.Login
 	org := payload.Org.Login
 	team := payload.Team.Name
-	sessionIDs := gha.stealUserSessionIDsForOrgAndTeam(username, org, team)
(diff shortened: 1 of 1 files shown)


cmd/nebraska/auth/githubauth.go
```diff
@@ -678,10 +678,7 @@ func (gha *githubAuth) loginWebhookMembershipEvent(c *gin.Context, payloadReader
 		return
 	}
 	logger.Debug("webhook", "got membership event with action", payload.Action)
-	if payload.Action != "removed" {
-		logger.Debug("webhook", "ignoring membership event with action", payload.Action)
-		return
-	}
+
 	logger.Debug("webhook", "got membership remove event with scope", payload.Scope)
```
krnowak, line 682:
```suggestion
	logger.Debug("webhook", "got membership event with scope", payload.Scope)
```

cmd/nebraska/auth/githubauth.go
```diff
@@ -678,10 +678,7 @@ func (gha *githubAuth) loginWebhookMembershipEvent(c *gin.Context, payloadReader
...
 	if payload.Scope != "team" {
 		logger.Debug("webhook", "ignoring membership remove event with scope", payload.Scope)
```
krnowak, line 684:
```suggestion
		logger.Debug("webhook", "ignoring membership event with scope", payload.Scope)
```

cmd/nebraska/auth/githubauth.go
```diff
@@ -690,11 +687,25 @@ func (gha *githubAuth) loginWebhookMembershipEvent(c *gin.Context, payloadReader
 	username := payload.Member.Login
 	org := payload.Org.Login
 	team := payload.Team.Name
-	sessionIDs := gha.stealUserSessionIDsForOrgAndTeam(username, org, team)
-	for _, sessionID := range sessionIDs {
-		logger.Debug("webhook", "dropping session", sessionID, "user", username)
-		gha.sessionsStore.MarkOrDestroySessionByID(sessionID)
+
+	if payload.Action == "added" {
+		logger.Debug("webhook", "action", payload.Action, "dropping all the sessions of user", username)
```
krnowak, line 692: Would it make sense to check if the team the user is added to is in either `gha.readWriteTeams` or in `gha.readOnlyTeams`? If the team is in neither, then dropping all the sessions is unnecessary - won't make a difference for the user.

krnowak (commented):
I'd make dropping the sessions conditional, otherwise looks good. But CI failed (probably a formatting issue).

krnowak (approved):

joaquimrocha (approved):

PR #341, opened 2021-03-02
backend: Cache Group Version Count Timeline

yolossn:
A cache layer for Group Version Count Timeline is added
with cache life span of 1 minute.

### DIFF
--- a/pkg/api/groups.go
+++ b/pkg/api/groups.go
@@ -59,9 +59,22 @@ var (
 	// itself. An alternative is to use atomic loads instead of RLock()
 	// and using atomic stores inside Lock() of a normal Mutex to serialize
 	// the writes (or use channel handshakes instead of a mutex).
-	cachedGroups     map[GroupDescriptor]string
-	cachedGroupsLock sync.RWMutex
+	cachedGroups                    map[GroupDescriptor]string
+	cachedGroupsLock                sync.RWMutex
+	cachedGroupVersionCount         = make(map[groupDurationCacheKey]groupVersionCountCache)
+	cachedGroupVersionCountLock     sync.RWMutex
+	cachedGroupVersionCountLifespan = time.Minute
 )
+
+type groupDurationCacheKey struct {
+	GroupID  string
+	Duration string
+}
+
+type groupVersionCountCache struct {
+	data     map[time.Time](VersionCountMap)
+	storedAt time.Time
+}
 
 type GroupDescriptor struct {
 	Track string
@@ -588,14 +601,27 @@ func durationParamToPostgresTimings(duration durationParam) (postgresDuration, p
 	return durationCodeToPostgresTimings(code)
 }
 
-func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (map[time.Time](VersionCountMap), error) {
+func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (map[time.Time](VersionCountMap), bool, error) {
+	cacheKey := groupDurationCacheKey{GroupID: groupID, Duration: duration}
+
+	cachedGroupVersionCountLock.RLock()
+	val, ok := cachedGroupVersionCount[cacheKey]
+	cachedGroupVersionCountLock.RUnlock()
+	if ok {
+		if time.Since(val.storedAt) < cachedGroupVersionCountLifespan {
+			logger.Debug().Str("cacheStatus", "HIT").Str("groupID", groupID).Str("duration", duration).Msg("GetGroupVersionCountTimeline")
+			return val.data, true, nil
+		}
+		logger.Debug().Str("cacheStatus", "STALE").Str("groupID", groupID).Str("duration", duration).Msg("GetGroupVersionCountTimeline")
+	}
+
 	var timelineEntry []VersionCountTimelineEntry
 	durationString, interval, err := durationParamToPostgresTimings(durationParam(duration))
 	// Get the number of instances per version until each of the time-interval
 	// divisions. This is done only for the instances that pinged the server in
 	// the last time-interval.
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 
 	instanceQuery, _, err := goqu.From("instance_application").Select(goqu.L("array_agg(DISTINCT (\"instance_id\")) \"instance_id\"")).Where(goqu.C("group_id").Eq(groupID),
@@ -603,12 +629,12 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
 		goqu.L(ignoreFakeInstanceCondition("instance_id"))).ToSQL()
 
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 	ids := []string{}
 	err = api.db.QueryRowx(instanceQuery).Scan(pq.Array(&ids))
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 	values := "VALUES "
 	for _, id := range ids {
@@ -643,19 +669,19 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
 
 	rows, err := api.db.Queryx(query, groupID)
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 	defer rows.Close()
 	for rows.Next() {
 		var timelineEntryEntity VersionCountTimelineEntry
 		err := rows.StructScan(&timelineEntryEntity)
 		if err != nil {
-			return nil, err
+			return nil, false, err
 		}
 		timelineEntry = append(timelineEntry, timelineEntryEntity)
 	}
 	if err := rows.Err(); err != nil {
-		return nil, err
+		return nil, false, err
 	}
 	allVersions := make(map[string]struct{})
 	timelineCount := make(map[time.Time]VersionCountMap)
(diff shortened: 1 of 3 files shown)


pkg/api/groups.go
```diff
@@ -59,10 +59,21 @@ var (
...
+	cachedGroupsLock                sync.RWMutex
+	cachedGroupVersionCount         map[groupDurationCacheKey]groupVersionCountCache
+	cachedGroupVersionCountLock     sync.RWMutex
+	cachedGroupVersionCountLifespan = time.Minute
 )
 
+type groupDurationCacheKey struct {
+	GroupID  string
+	Duration string
+}
```
joaquimrocha, line 72: I think we should add a newline between structs. I am surprised that golangci didn't complain about this.

pkg/api/groups_test.go
```diff
@@ -207,7 +207,7 @@ func TestGetVersionCountTimeline(t *testing.T) {
 	_ = a.updateInstanceStatus(instanceID, tApp.ID, InstanceStatusComplete)
 
 	// get VersionCountTimeline from 1 hr before now
-	versionTimelineMap, err := a.GetGroupVersionCountTimeline(tGroup.ID, "1h")
+	versionTimelineMap, _, err := a.GetGroupVersionCountTimeline(tGroup.ID, "1h")
```
joaquimrocha, line 210: It'd be great if we could test the cache as well.

pkg/api/groups.go
```diff
@@ -588,27 +599,45 @@ func durationParamToPostgresTimings(duration durationParam) (postgresDuration, p
...
 }
 
-func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (map[time.Time](VersionCountMap), error) {
+func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (map[time.Time](VersionCountMap), bool, error) {
+	if cachedGroupVersionCount == nil {
+		cachedGroupVersionCountLock.Lock()
+		cachedGroupVersionCount = map[groupDurationCacheKey]groupVersionCountCache{}
+		cachedGroupVersionCountLock.Unlock()
+	}
+	cacheKey := groupDurationCacheKey{GroupID: groupID, Duration: duration}
```
joaquimrocha, line 608: I am surprised this can be used as the cache key (I'd expect that the map would use this obj's location as the key, which could mean we'd never have a cache hit).

joaquimrocha (commented):
Left a nitpick and a couple of comments. If they don't make sense, we should merge it.

pkg/api/groups.go
krnowak: For the map to use object's location it would need to be of type `map[*groupDurationCacheKey]groupVersionCountCache`.

joaquimrocha: Right, but it's still creating a "new" struct above, so I am surprised how go can use that as the key. I am glad (and surprised) it works :wink:

krnowak: Key is taken by value, so map is doing the comparisons by value. You can think of `map[groupDurationCacheKey]groupVersionCountCache` as if it was `std::map<groupDurationCacheKey, groupVersionCountCache>*` in C++. ;)

pkg/api/groups.go
```diff
@@ -59,10 +59,21 @@ var (
 	// itself. An alternative is to use atomic loads instead of RLock()
 	// and using atomic stores inside Lock() of a normal Mutex to serialize
 	// the writes (or use channel handshakes instead of a mutex).
-	cachedGroups     map[GroupDescriptor]string
-	cachedGroupsLock sync.RWMutex
+	cachedGroups                    map[GroupDescriptor]string
+	cachedGroupsLock                sync.RWMutex
+	cachedGroupVersionCount         map[groupDurationCacheKey]groupVersionCountCache
```
krnowak, line 64: You could initialize the map here:
```suggestion
	cachedGroupVersionCount         = make(map[groupDurationCacheKey]groupVersionCountCache)
```

That way, you could drop the `nil` check in `GetGroupVersionCountTimeline`.

pkg/api/groups.go
```diff
@@ -588,27 +599,45 @@ func durationParamToPostgresTimings(duration durationParam) (postgresDuration, p
...
+	cachedGroupVersionCountLock.RLock()
+	if val, ok := cachedGroupVersionCount[cacheKey]; ok {
+		if time.Since(val.storedAt) < cachedGroupVersionCountLifespan {
+			logger.Debug().Str("cacheStatus", "HIT").Str("groupID", groupID).Str("duration", duration).Msg("GetGroupVersionCountTimeline")
+			cachedGroupVersionCountLock.RUnlock()
+			return val.data, true, nil
+		}
+		logger.Debug().Str("cacheStatus", "STALE").Str("groupID", groupID).Str("duration", duration).Msg("GetGroupVersionCountTimeline")
+	}
+	cachedGroupVersionCountLock.RUnlock()
```
krnowak, line 619: Since you take a shallow copy of `groupVersionCountCache` object, which means that you get a copy of the `storedAt` member, you could rewrite this in a following way to avoid remembering about RUnlocking the mutex in different places:

```go
cachedGroupVersionCountLock.RLock()
val, ok := cachedGroupVersionCount[cacheKey]
cachedGroupVersionCountLock.RUnlock()
if ok {
	…
}
```

pkg/api/groups.go
iaguis: Explaining Go through C++ :exploding_head:

pkg/api/groups.go
```diff
@@ -694,7 +723,18 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
 		}
 	}
 
-	return timelineCount, nil
+	go func() {
+		cachedGroupVersionCountLock.Lock()
+		cacheCopy := make(map[time.Time]VersionCountMap)
+		for key, val := range timelineCount {
+			cacheCopy[key] = val
+		}
```
krnowak, line 731: Is making a copy needed? If the returned map is not modified, we could store `timelineCount` directly. When we have a cache hit, we also don't return a copy. And if the map is modified by the caller then on the cache hit, we also should make a copy.

pkg/api/groups.go
```diff
@@ -694,7 +723,18 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
...
+		logger.Debug().Str("cacheStatus", "SET").Str("groupID", groupID).Str("duration", duration).Msg("GetGroupVersionCountTimeline")
+		cachedGroupVersionCount[cacheKey] = groupVersionCountCache{cacheCopy, time.Now()}
+		cachedGroupVersionCountLock.Unlock()
```
krnowak, line 734: If you are inside the goroutine, then you could do a deferred call to Unlock, right after the call to Lock.

krnowak (commented):
Some nitpicks.

pkg/api/groups.go
```diff
@@ -694,7 +723,18 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
 		}
 	}
 
-	return timelineCount, nil
+	go func() {
+		cachedGroupVersionCountLock.Lock()
+		cacheCopy := make(map[time.Time]VersionCountMap)
```
krnowak, line 728: Oh, one more thing. Would it make sense to check if the cache was updated in the meantime to avoid updating it for the second time? Basically, ensure that we are updating when there is no record for the key or if it's stale.

pkg/api/groups.go
joaquimrocha: Thanks @krnowak .

pkg/api/groups.go
yolossn: the map is not modified so removing the copy

pkg/api/groups_test.go
```diff
@@ -206,9 +207,31 @@ func TestGetVersionCountTimeline(t *testing.T) {
...
+	assert.Equal(t, false, isCache)
+
+	time.Sleep(time.Second * 10)
+	_, isCache, err = a.GetGroupVersionCountTimeline(tGroup.ID, "1h")
+	assert.NoError(t, err)
+
+	// the cache must be hit
+	assert.Equal(t, true, isCache)
+
+	time.Sleep(time.Second * 60)
```
krnowak, line 227: Could you file an issue about using some fake clock for testing, so we don't have tests that are sleeping for a minute?

pkg/api/groups.go
```diff
@@ -694,7 +720,17 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
 		}
 	}
 
-	return timelineCount, nil
+	go func() {
+		cachedGroupVersionCountLock.Lock()
+		defer cachedGroupVersionCountLock.Unlock()
+		val, ok := cachedGroupVersionCount[cacheKey]
+		if !ok || time.Since(val.storedAt) > cachedGroupVersionCountLifespan {
```
krnowak, line 727: The data in cache is valid if time since it was updated is less than a minute ago. Thus data in cache is stale if it's one minute or more. So:
```suggestion
		if !ok || time.Since(val.storedAt) >= cachedGroupVersionCountLifespan {
```

krnowak (approved):
Almost there.

krnowak (approved):

pkg/api/groups_test.go
yolossn: done https://github.com/kinvolk/nebraska/issues/342

PR #345, opened 2021-03-18
chart: Add option for separate /v1/update and /flatcar ingress

robinelfrink:
# Add option for separate /v1/update and /flatcar ingress

Optionally add an extra ingress handling paths /v1/update and /flatcar,
allowing different whitelisting options to the user interface and the
update endpoints.

# How to use

Enable extra ingress and set different annotations. Example `values.yaml` snippet:

```
ingress:
  enabled: true
  annotations:
    # address anonymized
    nginx.ingress.kubernetes.io/whitelist-source-range: 172.23.68.21
  update:
    enabled: true
    annotations: {}
```

# Testing done

Web interface is only accessible from `172.23.68.21`, flatcar nodes from anywhere can be upgraded.

### DIFF
--- a/charts/nebraska/Chart.yaml
+++ b/charts/nebraska/Chart.yaml
@@ -19,7 +19,7 @@ sources:
 maintainers:
   - name: kinvolk
     url: https://kinvolk.io/
-version: 0.1.1
+version: 0.1.2
 appVersion: "2.3.2"
 
 dependencies:
--- a/charts/nebraska/templates/update-ingress.yaml
+++ b/charts/nebraska/templates/update-ingress.yaml
@@ -46,7 +46,12 @@ spec:
               serviceName: {{ $fullName }}
               servicePort: {{ $svcPort }}
             {{- end }}
+          {{- if $.Values.config.hostFlatcarPackages.enabled }}
+          {{- if regexFind ( printf "://%s(:[0-9]+)?/" . ) ( default "" $.Values.config.hostFlatcarPackages.nebraskaURL ) }}
+          - path: {{ regexReplaceAll "^[^/]+://[^/]+" $.Values.config.hostFlatcarPackages.nebraskaURL "" | trimAll "/" }}/flatcar
+          {{- else }}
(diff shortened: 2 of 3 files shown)


joaquimrocha:
Hi @robinelfrink , can you give me a bit more context about this change (we don't have a `/flatcar` endpoint)?

robinelfrink:
> Hi @robinelfrink , can you give me a bit more context about this change (we don't have a `/flatcar` endpoint)?

This what my ingress controller says (anonimized) when I leave that path out:

```
[ingress-nginx-controller-5j4q6] 10.20.30.40 - - [23/Mar/2021:14:31:02 +0000] "GET /flatcar/flatcar-amd64-2765.2.1.gz HTTP/1.1" 403 153 "-" "-" 90 0.000 [nebraska-nebraska-80] [] - - - - f18726db811732e0075e0c06a33fce01
```

Maybe it's because we have `config.syncer.enabled: true` ?

robinelfrink:
Found it [here](https://github.com/kinvolk/nebraska/blob/5b2f43f1b8f5f7b85d3bcabd90f39e38a3a22289/cmd/nebraska/controller.go#L84):

```
PackagesURL:       conf.nebraskaURL + "/flatcar/",
```

joaquimrocha:
You're right!! The hosting option!
Cool. I will let @knrt10 review.

robinelfrink:
Now that I know where it's coming from, I think it should be configurable.

The least I can do is wrap an `{{- if .Values.config.hostFlatcarPackages.enabled }}` around it, and if possible even check if `ingress.hosts` matches `config.hostFlatcarPackages.nebraskaURL` and try to get the absolute path from that.

robinelfrink:
I added logic to extract the endpoint from `config.hostFlatcarPackages.nebraskaURL`.

knrt10 (approved):
+1. Looks great.

joaquimrocha:
Merged. Thanks @robinelfrink and @knrt10 .

PR #348, opened 2021-03-23
backend: Do not fail when trying to log without a session

joaquimrocha:
We have the noop auth backend which has no session set, and due to the
recent changes to set the user in the log context, it was failing.


cmd/nebraska/controller.go
```diff
@@ -60,6 +60,11 @@ type controllerConfig struct {
 
 func loggerWithUsername(l zerolog.Logger, c *gin.Context) zerolog.Logger {
 	session := ginsessions.GetSession(c)
+
```
iaguis, line 63: Usually in go there's no newline between these statements.

iaguis (approved):
Nit. LGTM otherwise.

iaguis (approved):

PR #349, opened 2021-03-24
chart: Bump Nebraska and chart version

joaquimrocha:


charts/nebraska/Chart.yaml
```diff
@@ -19,8 +19,8 @@ sources:
 maintainers:
   - name: kinvolk
     url: https://kinvolk.io/
-version: 0.1.2
-appVersion: "2.3.2"
+version: 0.1.3
+appVersion: "2.3.4"
```
knrt10, line 23: @joaquimrocha  any reason for it not to be `2.3.3`?

joaquimrocha: Because 2.3.4 is the latest release (from yesterday) and has bug fixes.

knrt10 (approved):

PR #358, opened 2021-04-21
Port remaining components to Typescript

### LINKED ISSUE #311: Port Nebraska to Typescript
**Description**
We need to port Nebraska to Typescript in order to prevent many issues that will be caught by simple type checking.

**Implementation options**
While at it, we shall also port any React class components to functional ones.

**Update:**
A lot of modules have been updated. Here is the remaining list:
- [x] ./frontend/src/js/components/Instances/
- [x] ./frontend/src/js/TestHelpers/
- [x] ./frontend/src/setupTests.js (trivial)
~./frontend/src/setupProxy.js~ (This one seems to not be supported in CRA: https://github.com/facebook/create-react-app/issues/6794 )

joaquimrocha:
- frontend: Port components/Instances/Charts.js to Typescript
- frontend: Port components/Instances/Details.js to Typescript
- frontend: Remove .react ext from components/Instances/Item.react.js
- frontend: Use the InstanceStatusHistory and Instance types where needed
- frontend: Port components/Instances/Item.js to Typescript
- frontend: Remove unused makeStyles import
- frontend: Remove unnecessary props argument from InstanceLayout
- frontend: Remove .react extension from components/Header.react.js
- frontend: Port components/Header.js to Typescript
- frontend: Do not use a new setting for the cached config header style
- frontend: Fix "Manage Access" button styling
- frontend: Let the "Manage Access" button get the correct color
- frontend: Do not use two different config objects in the Header
- frontend: Align the Account button to the right corner of the header
- frontend: Port TestHelpers/theme.js and setupTests.js to Typescript

Also fixes a few issues with the Header (showing the button in the right place with the right color).

fixes #311


frontend/src/js/components/Instances/Details.tsx
```diff
@@ -115,15 +122,26 @@ function StatusLabel(props) {
   );
 }
 
-function StatusRow(props) {
+interface StatusEvent {
+  status: number;
+  version: string;
+  error_code?: number;
+  created_ts: string | Date | number;
```
illume, line 129: created_ts is typed as string only in other places. Should it also be typed this way elsewhere?

frontend/package.json
```diff
@@ -60,6 +60,7 @@
     "@kinvolk/eslint-config": "^0.4.1",
     "@types/react-router-dom": "^5.1.7",
     "@types/underscore": "^1.10.24",
+    "@types/victory": "^33.1.4",
```
illume, line 63: Seems they are [distributing their own types now](https://github.com/FormidableLabs/victory/issues/1033#issuecomment-650405838) and that since then `@types/victory` [isn't getting updates](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/victory).

illume (approved):
Nice improvements. I left a couple of questions.

joaquimrocha:
Updated with the latest `victory` package. I will open an issue for us to replace victory with recharts (what we use for Headlamp).

PR #360, opened 2021-04-26
docs: Container Linux Config examples

pothos:
- docs: document how to set a machine alias    
    The machine alias is a recent Nebraska feature which wasn't documented
    yet.
-    docs: Use Container Linux Config in example   
    The cloud-init cloud config example should not be prominent because
    Container Linux Config (Ignition) is the recommended way for
    provisioning. Since both are YAML it's confusing to mention the
    cloud-init format in the example.

### DIFF
--- a/docs/managing-updates.md
+++ b/docs/managing-updates.md
@@ -60,6 +60,33 @@ In may take a few minutes to see an update request coming through. If you want t
 
 	update_engine_client -update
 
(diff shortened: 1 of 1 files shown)


joaquimrocha (approved):

joaquimrocha:
Good stuff, @pothos .

PR #366, opened 2021-05-04
Relicense the project as Apache 2.0

joaquimrocha:
We want to allow the integration of Nebraska with as many other projects
as possible, and to make its license consistent with the rest of the
Kinvolk's projects, so we're relicensing it as Apache 2.0.

Acked-By:
  Leon Plazier <leon.plazier@kpn.com>
  Marco Kilchhofer <marco@kilchhofer.info>
  Ricardo Katz <ricardo.katz@gmail.com>
  Robin Elfrink <robin@15augustus.nl, robin.elfrink@eu.equinix.com>
  Tom Offermann <toffermann@newrelic.com>


vbatts (approved):
LGTM

Doesn't have to be part of this PR, but we should double check that all source files have the appropriate comment header as well.

joaquimrocha:
> Doesn't have to be part of this PR, but we should double check that all source files have the appropriate comment header as well.

I did that search and couldn't find anything beyond this (unless I missed something...).

PR #372, opened 2021-05-10
OIDC Auth mode

yolossn:


frontend/src/js/stores/ApplicationsStore.ts
```diff
@@ -36,9 +38,16 @@ class ApplicationsStore extends Store {
...
       .catch(error => {
-        if (error.status === 404) {
-          this.applications = [];
-          this.emitChange();
+        switch (error.status) {
+          case 404:
+            this.applications = [];
+            this.emitChange();
+            break;
+          case 401:
```
illume, line 46: Are these the only status codes that need to be handled?

frontend/package.json
```diff
@@ -40,6 +42,7 @@
     "react": "^16.13.0",
     "react-color": "^2.17.3",
     "react-dom": "^16.13.0",
+    "react-redux": "^7.2.4",
```
illume, line 45: For the last 1.5 years or so the way to do idiomatic redux is different to what has been done here. The new way is a significant improvement for readability and reduces errors. IMHO, I think it's worth doing since this is the first time redux is being used in this codebase.

From the redux getting started guide: https://redux.js.org/introduction/getting-started
> Redux Toolkit is our official recommended approach for writing Redux logic.

frontend/src/js/__tests__/Applications/__snapshots__/Item.spec.js.snap
```diff
@@ -17,7 +17,7 @@ exports[`Application Item should render item correctly for no data 1`] = `
         class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12"
       >
         <div
-          class="MuiGrid-root makeStyles-gridHeader-125 MuiGrid-container MuiGrid-justify-xs-space-between"
+          class="MuiGrid-root makeStyles-gridHeader-4 MuiGrid-container MuiGrid-justify-xs-space-between"
```
illume, line 20: Seems like the only change here is the auto-generated classname IDs.

There's a generateClassName attribute that allows these IDs to be stable under test used in the Headlamp codebase here: https://github.com/kinvolk/headlamp/pull/202/files#diff-cc96163d7afd1ab1ae4d95870e5479069434913a583bf87706b2c429c56dacaaR22

frontend/src/js/components/Footer.tsx
```diff
@@ -1,23 +1,13 @@
...
+import { useTypedSelector } from '../stores/redux/reducers';
 
 function Footer() {
-  const [nebraskaConfig, setNebraskaConfig] = React.useState<null | NebraskaConfig>(null);
-  React.useEffect(() => {
-    API.getConfig().then(config => {
-      setNebraskaConfig(config);
-    });
-  }, []);
+  const {title='', nebraska_version=''} = useTypedSelector(state => state.config);
```
illume, line 6: Seems like these defaults would be better set inside the selector?

backend/cmd/nebraska/auth/oidcauth.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	q := redirectURL.Query()
+	q.Set("id_token", idToken)
+	redirectURL.RawQuery = q.Encode()
+	redirectTo(c, redirectURL.String())
+}
+
+func (oa *oidcAuth) login(c *gin.Context) {
+	// check if login_redirect_url is present in query params
+	loginRedirectURL := c.Request.URL.Query().Get("login_redirect_url")
```
illume, line 185: What if login_redirect_url is not there?

Does the URL need to be validated here?

backend/cmd/nebraska/nebraska.go
```diff
@@ -47,13 +47,21 @@ var (
 	ghReadWriteTeams    = flag.String("gh-rw-teams", "", "comma-separated list of read-write GitHub teams in the org/team format")
 	ghReadOnlyTeams     = flag.String("gh-ro-teams", "", "comma-separated list of read-only GitHub teams in the org/team format")
 	ghEnterpriseURL     = flag.String("gh-enterprise-url", "", fmt.Sprintf("base URL of the enterprise instance if using GHE; can be taken from %s env var too", ghEnterpriseURLEnvName))
+	oidcClientID        = flag.String("oidc-client-id", "", "OIDC client ID used for authentication")
+	oidcClientSecret    = flag.String("oidc-client-secret", "", "OIDC client Secret used for authentication")
+	oidcIssuerURL       = flag.String("oidc-issuer-url", "", "OIDC issuer URL used for authentication")
+	oidcCallbackURL     = flag.String("oidc-callback-url", "", "OIDC callback URL configured in your URL provider, /login/cb")
+	oidcAdminRoles      = flag.String("oidc-admin-roles", "", "comma-separated list of accepted roles with admin access")
```
illume, line 54: I feel these oidc- flags could be parsed to see if they are valid.

backend/cmd/nebraska/auth/oidcauth.go
```diff
@@ -0,0 +1,349 @@
...
+	issuerURL     string
+	callbackURL   string
+	provider      *oidc.Provider
+	verifier      *oidc.IDTokenVerifier
+	oauthConfig   *oauth2.Config
+	defaultTeamID string
+	adminRoles    []string
+	viewerRoles   []string
+	stateMap      map[string]string
+	rolesPath     string
```
illume, line 50: It could maybe make sense to document these fields.

backend/cmd/nebraska/auth/oidcauth.go
```diff
@@ -0,0 +1,349 @@
...
+			if !ok {
+				logger.Debug().Str("request_id", requestID).Msg("New Token doesn't contain ID Token")
+				httpError(c, http.StatusInternalServerError)
+				return "", true
+			}
+			c.Writer.Header().Set("id_token", idToken)
+			session.Set("refresh_token", newToken.RefreshToken)
+			tk, err = oa.verifier.Verify(ctx, idToken)
+			if err != nil {
+				logger.Error().Str("request_id", requestID).AnErr("error", err).Msg("Can't Verify New ID Token")
```
illume, line 296: I wonder if logging the Verify error would be useful info here?

backend/cmd/nebraska/auth/oidcauth.go
```diff
@@ -0,0 +1,349 @@
...
+		RedirectURL:  config.CallbackURL,
+		Scopes:       []string{oidc.ScopeOpenID},
+	}
+
+	verifier := provider.Verifier(oidcProviderConfig)
+	// state map is used keep track of login and callback requests
+	stateMap := map[string]string{}
+
+	// setup session store
+	cache := memcache.New(memcachegob.New())
```
illume, line 82: What are the security implications for using memcache for the session store?

illume:
I see lots of documentation in the code, and the command line arguments are documented already :+1: Is updating the documentation (docs/authorization.md) in scope for this PR? 

Is there any manual testing you would like me to do? (I know we were talking about testing with Azure Active Directory.)

A PR description would be useful for passers by.

backend/cmd/nebraska/controller.go
```diff
@@ -133,6 +139,16 @@ func NewClientConfig(conf *controllerConfig) *ClientConfig {
 			config.AccessManagementURL = GithubAccessManagementURL
 		}
 	}
+
+	if conf.oidcAuthConfig != nil {
+		config.AuthMode = "oidc"
+		if conf.nebraskaURL == "" {
+			config.LoginURL = "http://localhost:8000/login"
```
illume, line 146: I wonder if this default should be set somewhere else? Maybe in the command line args?

frontend/package.json
joaquimrocha: Interesting. I used https://react-redux.js.org/introduction/getting-started since we are using react too, but I agree the toolkit looks better (1.5 years! jeez, that's like 15 years in real, non-js life).

joaquimrocha:
> I see lots of documentation in the code, and the command line arguments are documented already  Is updating the documentation (docs/authorization.md) in scope for this PR?
> 

Yes. I had mentioned to Santhosh that we need the docs.

I will address the frontend changes.

frontend/src/js/stores/ApplicationsStore.ts
joaquimrocha: They are for now. It's an improvement over what we had before. But surely we should still find a better way to communicate when errors happen (for a different PR).

joaquimrocha:
I have updated the frontend part for including the redux changes.
@ashu8912 when you got some bandwidth (can be in another PR), please make the theme related changes so the tests run with static class names as René mentioned.
We should make the theme provider affect the whole app and not just the header, BTW.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
 After that, follow the same procedures as with GitHub.
 
+# Preparing Keycloak as OIDC provider for nebraska
+
+
+- Run `Keycloak` using docker:
+    - `docker run -p 8080:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin -d quay.io/keycloak/keycloak:13.0.1`
+
+- Open http://localhost:8080 in your browser to access keycloak UI and login with the username admin and password as admin.
+## Creating Roles
```
joaquimrocha, line 241: Separate with newline.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
+
+### Member Role
+
+1. Click on Roles menu option and select Add Role
+
+2. Provide a name for the member role, here we will use nebraska_member
+
+3. Click Save
```
joaquimrocha, line 249: I think these should go without empty lines in between.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
+
+### Admin Role
+
+1. Click on `Roles` menu option and select `Add Role`.
+
+2. Provide a name for the admin role, here we will use `nebraska_admin`.
+
+3. Click `Save`.
+
+4. After the admin role is created enable composite role to ON. In the Composite Roles section select the member role, In our case it is nebraska_member and click Add Selected.
```
joaquimrocha, line 259: Ditto.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
+
+Now the member and admin roles are created, the admin role is a composite role which comprises of member role.
+
+<p align="center">
+  <img width="751" height="281" src="https://github.com/kinvolk/nebraska/blob/oidc/docs/images/keycloak-roles.gif">
```
joaquimrocha, line 264: You should omit the height to get its aspect ratio preserved + maybe using `width="100%"` is a good idea.
More importantly, use a relative path for the image.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
+
+## Creating a client.
+
+1. Click on `Clients` menu option and click `Create`.
+
+2. Set the client name as `nebraska` and click `Save`.
+
+3. Change the `Access Type` to `Confidential`
+
+4. Set `Valid Redirect URIs` to `http://localhost:8000/login/cb`.
```
joaquimrocha, line 275: Same thing about newlines.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
+
+4. Set `Valid Redirect URIs` to `http://localhost:8000/login/cb`. 
+
+## Adding roles scope to token.
+
+1. Click on `Mappers` tab in Client Edit View. Click on `Create`.
+
+2. Set the name as `roles`, Select the `Mapper Type` as `User Realm Role`, `Token Claim Name` as `roles` and Select `Claim JSON Type` as String.
+
+3. Click `Save`
```
joaquimrocha, line 283: Ditto.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
+
+## Attaching Roles to User.
+
+1. Click on `Users` menu option and click `View all users`.
+
+2. Once the user list appears select the user and click on `Edit`.
+
+3. Go to `Role Mapping` tab and select `nebraska_admin` role and click on add selected to attach role to user. If you want to provide only member access access select the member role.
```
joaquimrocha, line 291: Ditto.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
 
 After that, follow the same procedures as with GitHub.
 
+# Preparing Keycloak as OIDC provider for nebraska
+
+
```
joaquimrocha, line 236: Use only one newline.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
 
 After that, follow the same procedures as with GitHub.
 
+# Preparing Keycloak as OIDC provider for nebraska
```
joaquimrocha, line 234: There should be an intro saying this is the recommended way for authenticating into Nebraska.

docs/authorization.md
```diff
@@ -231,6 +231,67 @@ to the URL of your GitHub Enterprise instance, for example
...
+## Attaching Roles to User.
+
+1. Click on `Users` menu option and click `View all users`.
+
+2. Once the user list appears select the user and click on `Edit`.
+
+3. Go to `Role Mapping` tab and select `nebraska_admin` role and click on add selected to attach role to user. If you want to provide only member access access select the member role.
+
+
+
```
joaquimrocha, line 294: Add one or two newlines, not 3.

joaquimrocha (changes requested):
It's looking good. Left some comments.
Also, I think this section should be the 1st one after the noop dev one. And we should add another commit where the github auth prints a deprecation warning + delete github from the docs.

docs/authorization.md
yolossn: since we use the raw image URL from Github I am not sure if using a relative path will work here.

joaquimrocha: It does!

backend/cmd/nebraska/auth/oidcauth.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	// Store refresh_token in session
+	session := ginsessions.GetSession(c)
+	session.Set("refresh_token", token.RefreshToken)
+	sessionSave(c, session, "login_cb")
+
+	// delete state from map as the callback is processed.
+	// also prevents reusing the same callback URL once again.
+	delete(oa.stateMap, state)
```
krnowak, line 168: I think we should be more stringent in cleaning up the state map. There can be errors during authentication that will result in keeping some stale records in the state map. Too many of those errors and the map will eat too much memory.

backend/cmd/nebraska/auth/oidcauth.go
yolossn: Added a way to specify valid redirect URLs using command line args. Right now it is a direct string match, should we opt-in for a domain level check and not worry about what the redirect path is?

joaquimrocha: If we could allow simple wildcards, then we could support both direct matches + prefixes (`https://kinvolk.io/*`).

backend/cmd/nebraska/auth/oidcauth.go
yolossn: We have been using Memcache as a session store for GitHub auth, AFAIK there isn't any.

backend/cmd/nebraska/auth/oidcauth.go
yolossn: If any user is not able to access the endpoints due to auth we can match request_id in the response header and the logs to find the exact reason. Ideally, the probability of this case occurring is relatively low as token expiry is handled already. So It will be useful to know these errors.

backend/cmd/nebraska/auth/oidcauth.go
illume: Just be careful it does not allow things past like: https://kinvolk.io.evil.example.com/ 

It should always have a `/` on the end before the wildcard? So `https://kinvolk.io/*` would be ok, but `https://kinvolk.io*` would not be.

frontend/src/js/__tests__/Applications/__snapshots__/Item.spec.js.snap
illume: I'll create a new issue for this.

frontend/src/js/__tests__/Common/__snapshots__/SectionHeader.spec.js.snap
```diff
@@ -37,7 +37,6 @@ exports[`Section Header renders section header correctly 1`] = `
               aria-hidden="true"
               class="MuiSvgIcon-root MuiSvgIcon-fontSizeSmall"
               focusable="false"
-              role="presentation"
```
illume, line 40: Not sure if removing this was intentional?

frontend/src/js/components/Main..tsx
```diff
@@ -49,6 +53,17 @@ const nebraskaTheme = createMuiTheme({
 });
 
 export default function Main() {
+  const dispatch = useDispatch();
+
+  React.useEffect(() => {
+    API.getConfig().then(config => {
+      console.debug("Got config", config)
```
illume, line 60: Maybe remove this console.debug now? The data is perhaps viewable in the redux dev tools.

backend/cmd/nebraska/auth/oidcauth.go
joaquimrocha: We're talking about a user defined parameter, so the user (and us) should take that into account when deploying, I don't think we can enforce some rule that prevents this while being flexible enough for all cases.

docs/authorization.md
```diff
@@ -40,6 +40,122 @@ the `postgres` container as follows:
 
 - In the browser, access `http://localhost:8000`
 
+# Deploying Nebraska for testing on local computer (OIDC authentication)
```
illume, line 43: Cool! I'll try these instructions out.

backend/pkg/util/wildcard.go
```diff
@@ -0,0 +1,35 @@
+package util
+
+// WildcardMatch is used to match * wildcard pattern in a string
+func WildcardMatch(pattern string, str string) bool {
+	return wildcardRuneMatch([]rune(pattern), []rune(str))
+}
+
+// Modified version of wildcard string match without support for ? operator
+// refer https://bit.ly/3557auF
+func wildcardRuneMatch(pattern []rune, str []rune) bool {
```
krnowak, line 10: Frankly speaking, I'm not a fan of a recursive nature of this function. Is there a chance that this function will be called with a string coming from a user (for example, from some specially crafted HTTP request)? If so, then I'd say that there is a risk of a stack exhaustion because of a too deep recursion when `str` is really super duper long. Maybe try converting it to an iterative version.

backend/pkg/util/wildcard_test.go
```diff
@@ -0,0 +1,31 @@
+package util
+
+import (
+	"testing"
+)
+
+func TestWildcardMatch(t *testing.T) {
+	testCases := []struct {
+		patten         string
```
krnowak, line 9:
```suggestion
		pattern        string
```

backend/pkg/util/wildcard_test.go
```diff
@@ -0,0 +1,31 @@
...
+		{"http://localhost:8000/*", "http://localhost:8000/config", true},
+		{"http://kinvolk.io/app/*", "http://kinvolk.io/app/groups", true},
+		{"http://kinvolk.io/app/*", "http://kinvolk.io/app/", true},
+		{"http://kinvolk.io/app/*/groups/*", "http://kinvolk.io/app/appID/groups/groupID", true},
+		{"http://kinvolk.io/app/*/groups/*", "http://kinvolk.io/app/appID/groups/", true},
+		{"http://kinvolk.io/app/*/groups/*", "http://kinvolk.io/app/groups/sfsdf", false},
+	}
+
+	for i, tc := range testCases {
+		output := WildcardMatch(tc.patten, tc.str)
```
krnowak, line 26:
```suggestion
		output := WildcardMatch(tc.pattern, tc.str)
```

krnowak (changes requested):

backend/pkg/util/wildcard.go
yolossn: Even with an iterative version, a significant number of specially crafted HTTP requests can hog the server. 

Maybe we can reduce the scope of the match just to allow * operator at the end of a pattern (e.g., support http://localhost:8000/nebraska/* and not http://localhost:8000/nebraska/*/app/*). Then the logic will reduce to this. 

```go
func WildcardMatch(pattern string, str string) bool {
	i := strings.Index(pattern, "*")
	if i == -1 {
		return str == pattern
	}

	if len(str) > i {
		return str[:i] == pattern[:i]
	}
	return str == pattern[:i]
}

```
This should work for the majority of the use cases.

Cc: @joaquimrocha Is this what we wanted from the beginning?

backend/cmd/nebraska/auth/oidcauth.go
```diff
@@ -0,0 +1,422 @@
...
+
+	result := gjson.Get(string(out), rolesPath)
+	result.ForEach(func(key, value gjson.Result) bool {
+		roles = append(roles, value.String())
+		return true
+	})
+	return roles, nil
+}
+
+func (oa *oidcAuth) Authenticate(c *gin.Context) (teamID string, replied bool) {
```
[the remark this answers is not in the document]
yolossn: Yes, a refactor was already under discussion. It will be done after Github auth is fully deprecated.

backend/pkg/util/wildcard.go
joaquimrocha: Assuming we're only allowing * wildcards, wouldn't it work and be simpler if we replace the pattern given URL with a regex expression and see if it matches?
Given URL `http://*.kinvolk.io/*` -> make it something like `pttrn := '^http://.*\.kinvolk\.io/.*'` + `matched, err := regexp.Match(pttrn, received_url)`

joaquimrocha (changes requested):

backend/pkg/util/wildcard.go
yolossn: The issue with converting the given URL to regex is that we will have to take into account all Regex tokens and handle them. I think I found a package that supports * wildcard with a better simple solution, Refer [this](https://github.com/ryanuber/go-glob/blob/51a8f68e6c24dc43f1e371749c89a267de4ebc53/glob.go#L11)

docs/authorization.md
```diff
@@ -40,6 +40,66 @@ the `postgres` container as follows:
...
+
+- Open http://localhost:8080 in your browser to access keycloak UI and login with the username admin and password as admin.
+
+## Creating Roles
+
+### Member Role
+
+1. Click on Roles menu option and select Add Role
+2. Provide a name for the member role, here we will use nebraska_member
+3. Click Save
```
joaquimrocha, line 70: nitpick: Special names below were written using backquotes for highlighting. I think we should do the same in this group as well.

docs/authorization.md
```diff
@@ -99,6 +99,48 @@ Now the member and admin roles are created, the admin role is a composite role w
...
+
+  idTokenClaims[`${namespace}/roles`] = assignedRoles;
+  accessTokenClaims[`${namespace}/roles`] = assignedRoles;
+
+  context.idToken = idTokenClaims;
+  context.accessToken = accessTokenClaims;
+  callback(null, user, context);
+}
+```
+Now the rule to add the roles to the token is setup, the roles will be available in the key `http://kinvolk.io/roles`.
```
joaquimrocha, line 153: Should we add the corresponding roles path value for the nebraska parameter in this and KeyCloak's docs?
Or are the deployment_docs enough?

yolossn: Added a note in Deploying Nebraska with OIDC section

backend/cmd/nebraska/auth/oidcauth.go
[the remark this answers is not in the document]
krnowak: Cool, thanks for confirming.

backend/cmd/nebraska/auth/oidcauth.go
```diff
@@ -0,0 +1,434 @@
...
+
+	now := time.Now()
+	oa.stateMap.Range(func(key, value interface{}) bool {
+		val, ok := value.(stateMessage)
+		if !ok {
+			oa.stateMap.Delete(key)
+			return true
+		}
+		if now.After(val.timeout) {
+			logger.Debug().Str("message", fmt.Sprintf("Deleting expired key %s from sate map", key))
```
krnowak, line 429:
```suggestion
			logger.Debug().Str("message", fmt.Sprintf("Deleting expired key %s from state map", key))
```

krnowak (approved):
One nitpick, otherwise ok.

joaquimrocha (approved):

joaquimrocha:
Thanks to everyone involved!
I have rebased and will merge now. Cheers!

PR #377, opened 2021-05-21
frontend: Fix labels for instance status area chart

ashu8912:
We don't want to show any label on instance status pie chart, the
correct way to not render any label is to pass a component to labels
prop which renders null.

## Testing done
Yes


illume (approved):
👍 🎉

Issue #350, opened 2021-03-30
Allow to sort the instances' list table


joaquimrocha:
**Current situation**
Users cannot sort the instances list by any criteria.

**Impact**
This makes it difficult to quickly find the instances users may be interested in.

**Ideal future situation**
Instances can be sorted by any of the columns' data.

**Implementation options**
The SimpleTable implementation in Headlamp already supports a very easy way to sort data.

Issue #228, opened 2020-06-12
Check if SQL queries for applications, channels, groups, instances and packages could be optimized


krnowak:
There is a list of things that currently needs data from other tables:

- application needs groups, channels and packages
- channel needs a package
- group needs a channel
- instance needs an instance application
- package needs a flatcar action

Currently, we do around 4 queries to get an application - 1 for getting data from application table, 1 for getting data from groups table, 1 for getting data from channel table, and 1 for getting data from package table. When getting N applications we do 1 + 3N queries - 1 for getting data from application table and then for N queries for channels, groups and packages each. This could likely be optimized to just 4 queries.

Similar goes for other examples (like getting N groups with 2 queries instead of 1 + N of them).

PR #378, opened 2021-05-24
Add sort to instance list

ashu8912:
# [Title: describe the change in one sentence]
Adds sort filter to instances list

## How to use
Go to instance list view and apply the various sort filters available, also use the sortOrder filter to change in what order you want the sorted data to appear(desc or asc)

## Testing done
Yes

### DIFF
--- a/backend/cmd/nebraska/controller.go
+++ b/backend/cmd/nebraska/controller.go
@@ -847,6 +847,8 @@ func (ctl *controller) getInstances(c *gin.Context) {
 	p.Status, _ = strconv.Atoi(c.Query("status"))
 	p.Page, _ = strconv.ParseUint(c.Query("page"), 10, 64)
 	p.PerPage, _ = strconv.ParseUint(c.Query("perpage"), 10, 64)
+	p.SortFilter = c.Query("sort")
+	p.SortOrder = c.Query("sortOrder")
 	duration := c.Query("duration")
 	result, err := ctl.api.GetInstances(p, duration)
 	if err == nil {
--- a/backend/pkg/api/instances.go
+++ b/backend/pkg/api/instances.go
@@ -3,6 +3,7 @@ package api
 import (
 	"database/sql"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/doug-martin/goqu/v9"
@@ -101,6 +102,66 @@ type InstancesQueryParams struct {
 	Version       string `json:"version"`
 	Page          uint64 `json:"page"`
 	PerPage       uint64 `json:"perpage"`
+	SortFilter    string `json:"sort_filter"`
+	SortOrder     string `json:"sort_order"`
+}
+
+type instanceFilterItem int
+
+const (
+	id instanceFilterItem = iota
+	ip
+	lastCheckForUpdates
+)
+
+var sortFilterMap = map[instanceFilterItem]string{
+	id:                  "alias",
+	ip:                  "ip",
+	lastCheckForUpdates: "last_check_for_updates",
+}
+
+type sortOrder int
+
+const (
+	sortOrderAsc sortOrder = iota
+	sortOrderDesc
+)
+
+func sortOrderFromString(str string) sortOrder {
+	val, err := strconv.Atoi(str)
+
+	/*
+		In case value is other than 0 or 1 or there is a wrong type of sortOrder passed
+		fallback to sortOrderDesc
+	*/
+	if (val != 0 && val != 1) || err != nil {
+		return sortOrderDesc
+	}
+	return sortOrder(val)
+}
+
+func sanitizeSortFilterParams(sortFilter string) string {
+	sortFilterNumericValue, _ := strconv.Atoi(sortFilter)
+	if value, ok := sortFilterMap[instanceFilterItem(sortFilterNumericValue)]; ok {
+		return value
+	}
+	return sortFilterMap[id]
+}
+
+func (api *API) checkIfColumnExistInTableQuery(tableName string, columnName string) (bool, error) {
+	query, _, err := goqu.Select(goqu.Func("EXISTS", goqu.From("information_schema.columns").Where(goqu.Ex{
+		"table_name":  tableName,
+		"column_name": columnName,
+	}))).ToSQL()
+
+	if err != nil {
+		return false, err
+	}
+
+	var isColumnAvailableInTable bool
+	err = api.db.QueryRowx(query).Scan(&isColumnAvailableInTable)
+
+	return isColumnAvailableInTable, err
 }
 
 // RegisterInstance registers an instance into Nebraska.
@@ -228,7 +289,10 @@ func (api *API) GetInstance(instanceID, appID string) (*Instance, error) {
 	if err != nil {
 		return nil, err
 	}
-	instanceApplication, err := api.getInstanceApp(appID, instance.ID, validityInterval)
+	/* passing "" to sortFilter while invoking getInstanceApp signifies we are not interested
+	in a sort
+	*/
+	instanceApplication, err := api.getInstanceApp(appID, instance.ID, validityInterval, "", 0)
 	switch err {
 	case nil:
 		instance.Application = *instanceApplication
@@ -240,10 +304,29 @@ func (api *API) GetInstance(instanceID, appID string) (*Instance, error) {
 
 	return &instance, nil
 }
-
-func (api *API) getInstanceApp(appID, instanceID string, duration postgresDuration) (*InstanceApplication, error) {
+func (api *API) getInstanceApps(appID, instanceID string, duration postgresDuration, sortFilter string, orderOfSort sortOrder, limit, offset uint) ([]*InstanceApplication, error) {
+	var instanceApps []*InstanceApplication
+	query, _, err := api.instanceAppQuery(appID, instanceID, duration, sortFilter, orderOfSort).Limit(limit).Offset(offset).ToSQL()
+	if err != nil {
+		return nil, err
+	}
+	rows, err := api.db.Queryx(query)
+	for rows.Next() {
+		var instanceApp InstanceApplication
+		err = rows.StructScan(&instanceApp)
+		if err != nil {
+			return nil, err
+		}
+		instanceApps = append(instanceApps, &instanceApp)
+	}
+	if err != nil {
+		return nil, err
+	}
+	return instanceApps, nil
+}
+func (api *API) getInstanceApp(appID, instanceID string, duration postgresDuration, sortFilter string, orderOfSort sortOrder) (*InstanceApplication, error) {
 	var instanceApp InstanceApplication
-	query, _, err := api.instanceAppQuery(appID, instanceID, duration).ToSQL()
+	query, _, err := api.instanceAppQuery(appID, instanceID, duration, sortFilter, orderOfSort).ToSQL()
 	if err != nil {
 		return nil, err
 	}
@@ -252,6 +335,18 @@ func (api *API) getInstanceApp(appID, instanceID string, duration postgresDurati
 		return nil, err
 	}
 	return &instanceApp, nil
+}
+func (api *API) getInstanceFromID(instanceID string) (*Instance, error) {
+	query, _, err := goqu.From("instance").Where(goqu.C("id").Eq(instanceID)).ToSQL()
+	if err != nil {
+		return nil, err
+	}
+	var instance Instance
+	err = api.db.QueryRowx(query).StructScan(&instance)
+	if err != nil {
+		return nil, err
+	}
+	return &instance, nil
 }
 
 // GetInstanceStatusHistory returns the status history of an instance in the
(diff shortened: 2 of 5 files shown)


backend/pkg/api/instances.go
```diff
@@ -302,7 +304,14 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
...
 	}
 	limit, offset := sqlPaginate(p.Page, p.PerPage)
+	columnToSortExpression := goqu.I(p.SortFilter).Asc().NullsLast()
+
+	if p.SortOrder == 1 {
+		columnToSortExpression = goqu.I(p.SortFilter).Desc().NullsLast()
+	}
+
 	query, _, err := api.instancesQuery(p, dbDuration).
+		Order(columnToSortExpression).
```
yolossn, line 314: Can you check if any indexes will be required for this ? 
Reference: https://www.postgresql.org/docs/8.3/indexes-ordering.html

joaquimrocha (changes requested):
I'd like to understand whether it's too complicated to add the sorting triggered from the column title cell's up/down buttons, as we have for Headlamp. I know that the filtering in Headlamp is client-side ATM, so this one needs more changes, but in terms of UX that's what was expected from the issue (#350 ).

ashu8912:
Screenshot from 2021-05-31 21-59-28
Screenshot from 2021-05-31 21-59-44

ashu8912:
Screenshot from 2021-05-31 22-00-53
Screenshot from 2021-05-31 22-01-04

joaquimrocha:
@ashu8912 , from the screenshots, I think the UX needs to be improved. Let's look into using the up/down arrows on the SimpleTable (take it and adapt it from Headlamp is needed) as we discussed on Slack.

backend/pkg/api/instances.go
```diff
@@ -101,6 +101,8 @@ type InstancesQueryParams struct {
 	Version       string `json:"version"`
 	Page          uint64 `json:"page"`
 	PerPage       uint64 `json:"perpage"`
+	SortFilter    string `json:"sort_filter"`
+	SortOrder     int    `json:"sort_order"`
```
krnowak, line 105: Maybe it would be better to have a separate enumeration for sort order like:
```go
type sortOrder int

const (
	sortOrderAsc sortOrder = iota
	sortOrderDesc
)

func sortOrderFromString(str string) (sortOrder, error) {
	…
}
```

That way, you would have a clearly defined allowed values for the sort order parameter (right now `1` is for descending sort, anything else is ascending sort), and you could avoid some magic numbers below.

ashu8912:
Screenshot from 2021-06-15 17-43-11

backend/cmd/nebraska/controller.go
```diff
@@ -847,6 +848,8 @@ func (ctl *controller) getInstances(c *gin.Context) {
 	p.Status, _ = strconv.Atoi(c.Query("status"))
 	p.Page, _ = strconv.ParseUint(c.Query("page"), 10, 64)
 	p.PerPage, _ = strconv.ParseUint(c.Query("perpage"), 10, 64)
+	p.SortFilter = sortFilter
```
illume, line 851: I guess this could just be this? `p.SortFilter = c.Query("sort")`

frontend/src/js/components/Instances/List.tsx
```diff
@@ -235,7 +248,7 @@ function ListView(props: { application: Application; group: Group }) {
 
   function applyFilters(_filters = {}) {
     const newFilters: { [key: string]: any } =
-      Object.keys(_filters).length !== 0 ? _filters : { status: '', version: '' };
+      Object.keys(_filters).length !== 0 ? _filters : { status: '', version: '', sort: 'id' };
```
illume, line 251: Sorting by id by default?

backend/pkg/api/instances.go
```diff
@@ -301,14 +394,63 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
...
+		return InstancesWithTotal{}, err
+	}
+
+	if existsInInstanceTable {
+		if sortOrder == sortOrderAsc {
+			instancesQuery = instancesQuery.Order(goqu.I(sortFilter).Asc().NullsLast())
+		} else if sortOrder == sortOrderDesc {
+			instancesQuery = instancesQuery.Order(goqu.I(sortFilter).Desc().NullsLast())
+		}
+	}
```
yolossn, line 414: This can also be moved below the `existsInInstanceAppTable` conditional block.

backend/pkg/api/instances.go
```diff
@@ -301,14 +394,63 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
...
+		return InstancesWithTotal{}, fmt.Errorf("The sort filter %s is invalid", sortFilter)
+	}
+	query, _, err := instancesQuery.
 		Limit(limit).
 		Offset(offset).
 		ToSQL()
+
 	if err != nil {
 		return InstancesWithTotal{}, err
 	}
```
yolossn, line 431: Maybe move the instancesQuery below the existsInInstanceAppTable conditional block for better readability

backend/pkg/api/instances.go
```diff
@@ -301,14 +394,63 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
...
+		}
+	}
+
+	//check if the sortFilter is part of the instance_app
+	existsInInstanceAppTable, err := api.checkIfColumnExistInTableQuery("instance_application", sortFilter)
+	if err != nil && (!existsInInstanceTable && !existsInInstanceAppTable) {
+		/* we are sure here that the sort filter passed doesn't exist in instance or the
+		instance_application table
+		*/
+		return InstancesWithTotal{}, fmt.Errorf("The sort filter %s is invalid", sortFilter)
```
yolossn, line 422: The error message will be wrong in the case there was an error returned by `checkIfColumnExistInTableQuery`

backend/pkg/api/instances.go
```diff
@@ -101,6 +102,64 @@ type InstancesQueryParams struct {
...
+	case id:
+		return sortFilterMap[id]
+	case ip:
+		return sortFilterMap[ip]
+	case lastCheckForUpdates:
+		return sortFilterMap[lastCheckForUpdates]
+	default:
+		// if the sortFilter is invalid fallback to sort by ID
+		return sortFilterMap[id]
+	}
```
yolossn, line 146: this can just be 
``` go
if val,ok:=sortFilterMap[sortFilterNumericValue];ok{
   return sortFilterMap[sortFilterNumericValue]
}
return sortFilterMap[id]
```

backend/pkg/api/instances.go
```diff
@@ -101,6 +102,64 @@ type InstancesQueryParams struct {
...
+
+const (
+	sortOrderAsc sortOrder = iota
+	sortOrderDesc
+)
+
+func sortOrderFromString(str string) (sortOrder, error) {
+	val, err := strconv.Atoi(str)
+
+	return sortOrder(val), err
```
yolossn, line 131: Here the returned value can be any `int` and there is a chance the value will not match both the cases in

```go
               if sortOrder == sortOrderAsc {
			instancesQuery = instancesQuery.Order(goqu.I(sortFilter).Asc().NullsLast())
		} else if sortOrder == sortOrderDesc {
			instancesQuery = instancesQuery.Order(goqu.I(sortFilter).Desc().NullsLast())
		}
```

yolossn: Is there a need to return an error? In this case, the sortOrder will default to `sortOrderAsc`.

frontend/src/js/components/Instances/List.tsx
```diff
@@ -124,7 +128,13 @@ function ListView(props: { application: Application; group: Group }) {
   /*TODO: use the URL as the single source of truth and remove states */
   const [page, setPage] = React.useState(0);
   const [rowsPerPage, setRowsPerPage] = React.useState(10);
-  const [filters, setFilters] = React.useState<{ [key: string]: any }>({ status: '', version: '' });
+  //false sortOrder means it's in ascending order
+  const [sortOrder, setSortOrder] = React.useState(false);
```
joaquimrocha, line 132: The code would be more self-explainatory if this was e.g. `isAscSortOrder`, `setIsAscSortOrder` = false|true

frontend/src/js/components/Instances/List.tsx
joaquimrocha: Yeah, maybe sorting by last update check by default would be better.

frontend/src/js/components/Instances/List.tsx
```diff
@@ -319,12 +343,12 @@ function ListView(props: { application: Application; group: Group }) {
             </Grid>
             <Box width="100%" borderTop={1} borderColor={'#E0E0E0'} className={classes.root}>
               <Grid item container md={12} alignItems="stretch" justify="space-between">
-                <Grid item md>
+                <Grid item md={3}>
```
joaquimrocha, line 346: Is this left over from your original approach?

frontend/src/js/components/Instances/List.tsx
```diff
@@ -319,12 +343,12 @@ function ListView(props: { application: Application; group: Group }) {
...
                   <Box ml={2}>
                     <InstanceCountLabel countText={getInstanceCount()} instanceListView />
                   </Box>
                 </Grid>
-                <Grid item md>
+                <Grid item md={9}>
```
joaquimrocha, line 351: Is this left over from your original approach?

frontend/src/js/components/Instances/List.tsx
```diff
@@ -291,6 +308,13 @@ function ListView(props: { application: Application; group: Group }) {
   function isFiltered() {
     return filters.status || filters.version;
   }
+
+  function sortHandler(sortOrder: boolean, sortQuery: string) {
+    setSortOrder(sortOrder);
+    setSortQuery(sortQuery);
+    addQuery({ sort: sortQuery, sortOrder: Number(sortOrder) });
```
joaquimrocha, line 315: I'd prefer the sort values to be human readable, like `sort=name` or `sort=last-check`, etc.
This can be only in the frontend of course, where we match a number and do the request to the backend with that.

joaquimrocha (changes requested):

backend/pkg/api/instances.go
ashu8912: I have moved the error checking above this block

backend/pkg/api/instances.go
```diff
@@ -301,8 +394,62 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
...
+	if !existsInInstanceTable && !existsInInstanceAppTable {
+		/* we are sure here that the sort filter passed doesn't exist in instance or the
+		instance_application table
+		*/
+		return InstancesWithTotal{}, fmt.Errorf("The sort filter %s is invalid", sortFilter)
+	}
+
+	if err != nil {
+		return InstancesWithTotal{}, err
+	}
```
yolossn, line 421: Duplicate error handling block can be removed.

backend/pkg/api/instances.go
```diff
@@ -253,6 +334,18 @@ func (api *API) getInstanceApp(appID, instanceID string, duration postgresDurati
 	}
 	return &instanceApp, nil
 }
+func (api *API) getInstanceFromApp(instanceIDOfApp string) (*Instance, error) {
```
yolossn, line 337: This can be 
```
func (api *API) getInstanceFromID(instanceID string) (*Instance, error) {
```

yolossn: @ashu8912 looks like this got missed.

backend/pkg/api/instances.go
```diff
@@ -301,8 +394,60 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
 	if err != nil {
 		return InstancesWithTotal{}, err
 	}
+
 	limit, offset := sqlPaginate(p.Page, p.PerPage)
-	query, _, err := api.instancesQuery(p, dbDuration).
+	sortFilter := sanitizeSortFilterParams(p.SortFilter)
+	sortOrder := sortOrderFromString(p.SortOrder)
+	existsInInstanceTable, err := api.checkIfColumnExistInTableQuery("instance", sortFilter)
+
```
yolossn, line 402: small nit, the empty line is not needed here.

backend/pkg/api/instances.go
ashu8912: fixed

backend/pkg/api/instances.go
ashu8912: thanks,fixed

backend/pkg/api/instances.go
```diff
@@ -301,8 +394,59 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
...
+		}
+		for _, application := range applications {
+			var instance *Instance
+			instance, err = api.getInstanceFromID(application.InstanceID)
+			if err != nil {
+				return InstancesWithTotal{}, err
+			}
+			instance.Application = *application
+			instances = append(instances, instance)
+		}
```
krnowak, line 432: Shouldn't this be something to get with a single query that joins instance apps table and instances table? Or maybe with two queries, where one gets the instance apps, and second gets all the interesting instances? Because right now it looks like we are doing 1 + N queries here.

ashu8912: @krnowak yes I was also thinking about a join query at the beginning but the problem is we have been doing such queries almost everywhere and this 1+N querying is not something specific to this patch.

we already have an issue with that https://github.com/kinvolk/nebraska/issues/228, I didn't change it to a JOIN query because it was moving a lot of code pieces, here the plan was to add a separate patch that tackles all these query optimization things.

ashu8912: Also the table instance and instance_applications are huge so things like JOIN would be costly this is something we still need to investigate

krnowak: Heh, I even forgot that I already filed an issue about it. So I'm going to let it slide another time then. :)

frontend/src/js/utils/helpers.ts
```diff
@@ -19,6 +19,18 @@ export const ARCHES: {
   3: 'X86',
 };
 
+export const InstanceSortFilters:{
```
krnowak, line 22: This one could use a comment like the `ARCHES` constant above has.

backend/pkg/api/instances.go
```diff
@@ -101,6 +102,64 @@ type InstancesQueryParams struct {
...
 	PerPage       uint64 `json:"perpage"`
+	SortFilter    string `json:"sort_filter"`
+	SortOrder     string `json:"sort_order"`
+}
+
+const (
+	id int = iota
+	ip
+	lastCheckForUpdates
+)
```
krnowak, line 113: Could these constants be of a separate type, so they are not accidentally misused as some random number?

krnowak (changes requested):
Some minor nits, looks good otherwise.

joaquimrocha:
I have tested this again but it still doesn't sort the instances correctly if they have an alias.
Test:
1. Go to the instances list in some group, and select an instance in the middle of the table;
2. Go to the 3-dot menu and select Rename. Rename the instance as `zzz`.
3. Do the same with another instance from the middle of the instance list but rename it as `aaa`
4. Use the instance ID column's sorting buttons to sort it.

Expected outcome: the `aaa` instance shows up as the first one, and the `zzz` as the last one, when sorting ascendently. And the opposite when sorting in decending order.

Actual outcome: those instances are not shown at the first/last places.

Also, I've noticed I don't see the aaa/zzz instances with these names at all after pressing sort.

frontend/src/js/components/Instances/List.tsx
```diff
@@ -348,18 +370,24 @@ function ListView(props: { application: Application; group: Group }) {
...
+                    />
                     <TablePagination
                       rowsPerPageOptions={[10, 25, 50, 100]}
                       component="div"
                       count={instancesObj.total}
                       rowsPerPage={rowsPerPage}
                       page={page}
                       backIconButtonProps={{
-                        'aria-label': 'previous page',
+                        'aria-label': 'previous page'
```
joaquimrocha, line 387: This and other simimlar changes shouldn't be in the PR. Is husky/pretty not handling this?

PR #398, opened 2021-06-22
Fix a11y issues, Add eslint-plugin-jsx-a11y as a dev dependency

ashu8912:


illume:
I'm taking a look at the test error in here (which I think is probably unrelated to this PR).

illume:
The issue is not related to it using npm v7 somehow... but mainly about old deps (I guess).

So I'm going to try and update Nebraska dependencies in a separate PR first.

illume:
I fixed the dependency issue. In the end I simply redid the `npm install eslint-plugin-jsx-a11y` and it worked. 🤷 
Moved the commits around a bit so they all pass (added the fixes before enabling the lint rules), and tested them one at a time. I thought somehow that error would come back, but it didn't.

illume (approved):
👍 thanks

PR #401, opened 2021-06-25
frontend: Replace victory charts with recharts

ashu8912:
We want to use one common charting library among all our projects we are already using recharts on headlamp and also on some parts of nebraska, this patch removes the few occurrence where victory chart is used by rechart


ashu8912:
@joaquimrocha while adding this patch i found out that there are a few more depdendency that we don't use but we still ship with node_modules,
You can take a look at these by running ```npx depcheck``` from the frontend dir

joaquimrocha (changes requested):
I have commented on slack as it was easier. Also, please squash both commits, as after the first one, victory is not needed.

joaquimrocha:
@ashu8912 , Thanks for these changes. Really nice that everything looks just like it was.
The pie charts still don't animate for me but you added some attributes for the animation. Do they work for you in the end?

If not, we can still merge this after fixing the conflict. Cheers.

PR #403, opened 2021-06-29
Start adding I18n

illume:
# Add i18n infra

This PR adds the i18n infra based on the Headlamp did i18n. 
- [x] t()/Trans used in all the components
- [x] docs 
- [x] Make sure tests run (probably needs some jest configuration)
- [x] lint and type check passes

With this PR it should run the same for users as Nebraska, except it will use the i18n infrastructure. 

Further PRs will:
- add other locales (this PR does not even have an 'en' locale).
- add the locale selector into the UI

... but first a11y should be done, because that will likely add a bunch of strings.

## How to use

`make run-frontend`

## Testing done

- existing specs pass
- lint and typechecking passes
- visually it looks similar


joaquimrocha (approved):
The docs refer to make i18n but I don't see that target added to the makefile.
After that, it'd should be good to go.

Another difference I noticed is that in Headlamp we don't have capitalized namespaces. I'd prefer to keep it that way, but it's not a big deal :shrug:

illume:
Yeah, I'll uncapitalize. I just matched them to the Capitalized folder names. Will add the make target too.

illume:
I rebased against master, fixed the cases, and added the `make i18n` target.

joaquimrocha (approved):

PR #409, opened 2021-07-01
Humanize policy numbers

joaquimrocha:
- frontend: Fix type of argument in Group/Item's translation
- frontend: Share the format for rate policy in Groups/Item|ItemExtended
- frontend: Inform the user of unlimited updates in the update policy

Screenshot from 2021-07-01 12-30-52


illume (approved):
👍 thanks

PR #410, opened 2021-07-01
Frontend dependency updates, CI improvements

illume:
Dependencies baked fresh...

- frontend: some unused dependencies removed (with help from `npx depcheck`)
- frontend: dependency updates
- add some front end CI tests (test frontend:formatting/typechecking), and speed up CI
- update alpine to latest for container
- github: test building a container image

---

These frontend dependencies are left alone:
```
Package         Current   Wanted  Latest  Location
@iconify/tools   1.3.17   1.3.18  1.3.18  Nebraska
@types/node     14.17.4  14.17.4  16.0.0  Nebraska
husky             4.3.8    4.3.8   7.0.0  Nebraska
lint-staged      10.5.4   10.5.4  11.0.0  Nebraska
```

- Latest @iconify/tools gives a syntax error. 
- We are using node 14 so use @types/node 14.
- We are using the same version of husky/lint-staged as headlamp and because it changes the way it works. Maybe worth a look again, since husky changed its license back to MIT from a fund-raising-type-of-license now.

---

Also `@kinvolk/eslint-config` should probably get an update, but that will be done in a separate PR.


frontend/package.json
```diff
@@ -75,6 +75,7 @@
     "eject": "react-scripts eject",
     "lint": "eslint -c package.json 'src/**/*.{js,jsx,ts,tsx}'",
     "format": "prettier --config package.json --write src/",
+    "format-check": "prettier --config package.json --check src/",
```
joaquimrocha, line 78: I am getting a bit confused with having lint + format-check. Should we just use pretty, or maybe mix both into one command?

joaquimrocha (commented):
Left a comment about the format/lint, the rest looks fine to me but I will build an image and test it on staging.

frontend/package.json
illume: The intention is to check that the code has been formatted on CI.

format-check could be added into lint if you prefer.
```
"lint": "eslint -c package.json 'src/**/*.{js,jsx,ts,tsx}' && prettier --config package.json --check src/",
```

illume: I changed it to this. So the format check is done as part of the `npm run lint` / `make frontend-lint`.

PR #411, opened 2021-07-01
Improve instances sorting

joaquimrocha:
- frontend: Set arrow-up/down as asc/desc order in Instances/Table
- frontend: Add the swap icon to sort buttons in not-sorted columns
- frontend: Use asc/desc names for the sort order instead of a boolean
- frontend: Sort instances list from most recent ping to least recent


illume (approved):
👍

PR #416, opened 2021-07-08
Add unused import linter plugin also don't allow any unused vars

ashu8912:


illume (approved):
🎉 👍 great! Thanks.

frontend/package.json
```diff
@@ -110,7 +111,12 @@
       "@kinvolk",
       "prettier",
       "plugin:jsx-a11y/recommended"
-    ]
+    ],
+    "plugins": ["unused-imports"],
+    "rules": {
```
illume, line 116: Probably we should add this to the shared eslint config later. I could do that if you like... since I already wanted to do a fix for something else?

joaquimrocha: Yes, I'd rather have this in the shared one (assuming we can have that there instead of here). Please go ahead @illume .
Thank you @ashu8912 for spotting and fixing this.

ashu8912: @illume also to reduce some work for you, you can use the second commit from here after we have added the plugin to the shared config

illume: There's a PR here: https://github.com/kinvolk/eslint-config/pull/9 (I tested it with nebraska already).

I guess the first commit in here needs to change by updating the `@kinvolk/eslint-config` to 0.5.0 and those rules and plugin addition removed. (Once the new eslint-config has been released that is.)

joaquimrocha:
I have now published kinvolk/eslint-config:0.5.0

joaquimrocha (approved):

PR #417, opened 2021-07-08
Use main as default branch

joaquimrocha:


joaquimrocha:
Notice that we need to update the master references in the charts yaml but those are for getting the docs/icon and require also bumping the charts version to work, so we'll update the master references when we need to update the chart version.

illume (approved):
🎉 🎈

PR #418, opened 2021-07-09
Build simple binary on images

joaquimrocha:
- Dockerfile: Build the backend binary without generating changes
- .github: Run build-container action if Dockerfile is touched


illume (approved):
👍

PR #428, opened 2021-07-30
backend: Fix instance list filtering when filtered by status & version

### LINKED ISSUE #426: Filtering instances by status doesn't show the correct list of instances
## Description

When filtering the instances list view by status, the number of instances shown matches the number of instances with that status but it shows an incorrect list of instances in the table.

## Impact

Users cannot list the instances that have a certain status.

1. With multiple instances reported in a group, and some in a status like `error`, go to the instances view and select the error status.
2. Notice how the instances listed do not show the selected status.

## Acceptance criteria

- [ ] Upon filtering by status, only instances with that status are shown.
- [ ] If this is a backend issue, create a unit test for it

ashu8912:
By default the sortFilter is found in the instance app table and when
mapping instanceApplications to instance we were not applying the
filters, this lead to wrong list of instances to be shown in the UI.

Go to the instance list view apply the filters and you will see the correct list with filters getting applied 

## Testing done
YES

fixes #426

### DIFF
--- a/backend/pkg/api/instances.go
+++ b/backend/pkg/api/instances.go
@@ -315,6 +315,9 @@ func (api *API) getInstanceApps(appID, instanceID string, duration postgresDurat
 	if p.Version != "" {
 		query = query.Where(goqu.C("version").Eq(p.Version))
 	}
+	if p.GroupID != "" {
+		query = query.Where(goqu.C("group_id").Eq(p.GroupID))
+	}
 	instanceAppsQuery, _, err := query.ToSQL()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


ashu8912:
fixes #428

backend/pkg/api/instances.go
```diff
@@ -438,6 +447,7 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
 		}, nil
 	}
 	instancesQuery := api.instancesQuery(p, dbDuration)
+	fmt.Println(instancesQuery)
```
joaquimrocha, line 450: This should not be here.

joaquimrocha (changes requested):
One comment about a left-overv print.
Otherwise LGTM. Thanks!

joaquimrocha:
Turns out the instances list had a bigger problem: the group was also not being considered when fetching the instances.
I added a commit to fix that. Please remove the print line from yours.

joaquimrocha:
@ashu8912 , I have added the test coverage. Once the CI confirms it all works, feel free to review + merge if it LGTY.

PR #429, opened 2021-07-30
Makefile: Ensure the test targets are run from the backend dir

joaquimrocha:
Some targets were not updated when the backend code was moved into the
backend/ directory, and it was passing the CI tests still because the
diff that triggers the CI testing was running against files that did
not exist.


ashu8912 (approved):
Thanks :tada:

PR #430, opened 2021-08-03
version from source

joaquimrocha:
- Makefile: Allow to override the version label
- build: Set the version from the original source instead of the container


illume (approved):
👍 lgtm

PR #431, opened 2021-08-04
frontend: Fix persistent chart tooltip 

### LINKED ISSUE #424: Persistent hover effect in the status charts
## Description

I have noticed that I can leave the group's status charts (circle ones) in a state where they show the tooltips and highlighted state as if they're still hovered.

## Impact

The UI gets confusing and looks broken in this state.

1. Go to a group that has instances reported to it;
2. Hover quickly on the circle charts

## Expected behavior

If moving the cursor off a chart, its highlighted effect and tooltip will disapear.

Here is the wrong behavior's screenshot:

Screenshot from 2021-07-26 15-41-55

ashu8912:
Set the tooltip only when the tooltip boolean is not turned on
## Testing done
Go to the GroupExtended View and hover over InstanceStatusArea Charts quickly jump from one chart to other and the tooltip would not remain persistent.
fixes #424

### DIFF
--- a/frontend/src/js/components/Instances/Charts.tsx
+++ b/frontend/src/js/components/Instances/Charts.tsx
@@ -122,12 +122,14 @@ function ProgressDoughnut(props: ProgressDoughnutProps) {
             animationDuration={1000}
             animationEasing={'ease-in-out'}
             onMouseOver={(dataum, index) => {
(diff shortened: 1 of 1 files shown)


joaquimrocha (approved):
Works great. Thanks!

PR #432, opened 2021-08-04
charts: Update to Nebraska 2.4.1

joaquimrocha:


charts/nebraska/Chart.yaml
```diff
@@ -20,7 +20,7 @@ maintainers:
   - name: kinvolk
     url: https://kinvolk.io/
 version: 0.1.5
```
yolossn, line 22: To make a release the version should also be bumped.

yolossn (approved):
LGTM

PR #433, opened 2021-08-04
Allow other result success cases to complete the update

joaquimrocha:
We had a [limit](https://github.com/kinvolk/nebraska/blob/main/backend/pkg/api/events.go#L167) on how we consider an update to be done because we were only contemplating how Flatcar behaves.

Yet, as far as understand from the [protocol](https://github.com/google/omaha/blob/master/doc/ServerProtocolV3.md), there are [3 different ways](https://pkg.go.dev/github.com/kinvolk/go-omaha/omaha#EventResult) to represent success and we should definitely consider them.

**Update:** Turns out the event types are stored in the DB and I don't want to add a migration to include the ResultSuccessRestartBrowser case, so I dropped support for this event result type. I could add a hack where I always query the ResultSuccess if there's ResultSuccessRestartBrowser, but I don't think it's worth it and instead we should look into whether it's worth having the event_type data in the DB in the first place (having the info in the source should be enough).

Another issue with not understanding all success result cases is that when using the (deprecated but still useful) `updateservicectl`, the fake instances would reach an "unknown" state because of this [corner case logic](https://github.com/kinvolk/nebraska/blob/main/backend/pkg/api/events.go#L108).
With this PR it will successfully consider those fake instances updates to be completed, and I don't think the PR will change anything in the Flatcar case because we are [using only "ResultReboot" AFAICS](https://github.com/kinvolk/update_engine/blob/312ab260fce283e84635b73d92dd6d526d1e3d96/src/update_engine/omaha_request_action.cc#L136
). I'd like @pothos to double check this is the case.

### DIFF
--- a/backend/pkg/api/events.go
+++ b/backend/pkg/api/events.go
@@ -164,8 +164,9 @@ func (api *API) triggerEventConsequences(instanceID, appID, groupID, lastUpdateV
 		return err
 	}
 
-	// TODO: should we also consider ResultSuccess in the next check? Flatcar ~ generic conflicts?
-	if etype == EventUpdateComplete && result == ResultSuccessReboot {
+	// We allow the plain ResultSuccess here only if the app is not Flatcar because Flatcar is relying on
+	// having only the update-complete logic on ResultSuccessReboot.
+	if etype == EventUpdateComplete && (result == ResultSuccessReboot || (appID != flatcarAppID && result == ResultSuccess)) {
 		if err := api.updateInstanceStatus(instanceID, appID, InstanceStatusComplete); err != nil {
 			logger.Error().Err(err).Msg("triggerEventConsequences - could not update instance status")
 		}
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ i18n:
 	cd frontend && npm run i18n
 
 run-backend: backend-binary
-	cd backend && ./bin/nebraska -auth-mode noop
+	cd backend && ./bin/nebraska -auth-mode noop -debug
 
 .PHONY: backend
 backend: run-generators backend-code-checks build-backend-binary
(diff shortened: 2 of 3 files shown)


pothos:
I've done some tests and it looked good but it's not easy to answer: there is an initialization in update-engine with `kTypeUpdateComplete` and `kResultSuccess` is the default - I don't know yet whether this is will be sent and under which conditions.

pothos:
Yes, so the instance sends `<event eventtype="3" eventresult="1"></event>` after the update payload is applied (showing in Nebraska logs as `event="update complete.success"`) and then after the instance rebooted it sends `<event eventtype="3" eventresult="2" previousversion="2….x.y"></event>` (showing in Nebraska as `event="update complete.success reboot"`).

backend/pkg/api/events.go
```diff
@@ -164,8 +164,7 @@ func (api *API) triggerEventConsequences(instanceID, appID, groupID, lastUpdateV
 		return err
 	}
 
-	// TODO: should we also consider ResultSuccess in the next check? Flatcar ~ generic conflicts?
-	if etype == EventUpdateComplete && result == ResultSuccessReboot {
+	if etype == EventUpdateComplete && (result == ResultSuccessReboot || result == ResultSuccess) {
```
pothos, line 167:
```suggestion
	if etype == EventUpdateComplete && (result == ResultSuccessReboot || result == ResultSuccess) {
```
I would make it a property of the application whether `ResultSuccessReboot` or `ResultSuccess` is expected here.
Instead of adding a DB property you can also add a check here against the hardcoded Flatcar application ID.

pothos (changes requested):

backend/pkg/api/events.go
joaquimrocha: @pothos Did you mean to add a code suggestion?

So AFAIU, you're saying "keep this change you added but only if the app is not FCL".

joaquimrocha:
I have updated the branch to include the old FCL behavior and a test. We should start adding more tests, especially those covering special Flatcar behavior.

backend/pkg/api/events.go
pothos: Thanks, looks good now.

Yes, I didn't want to add a suggestion but just a comment on that line.

pothos (approved):

PR #437, opened 2021-08-12
backend: Change instance fetch query to a join query & add search

ashu8912:
# [Title: describe the change in one sentence]
Port instance fetching query to a join query

## Testing done
Go to instance list view and try using filters and pagination to see whether the current approach of getting instance works and is similarly fast as the earlier approach

- [x] Tests added

##Features
**Search without classifier**

- [x] Search by alias
- [x] Search by id
- [x] Search by ip

**Search With classifier**

- [x] Search by alias
- [x] Search by id
- [x] Search by ip

### DIFF
--- a/backend/pkg/api/instances.go
+++ b/backend/pkg/api/instances.go
@@ -104,6 +104,8 @@ type InstancesQueryParams struct {
 	PerPage       uint64 `json:"perpage"`
 	SortFilter    string `json:"sort_filter"`
 	SortOrder     string `json:"sort_order"`
+	SearchFilter  string `json:"search_filter"`
+	SearchValue   string `json:"search_value"`
 }
 
 type instanceFilterItem int
@@ -341,6 +343,29 @@ func prepareGetInstancesQuery(instanceQuery *goqu.SelectDataset, instanceAppQuer
 	).Select(goqu.L("*"))
 }
 
+func prepareSearchQuery(finalQuery *goqu.SelectDataset, p InstancesQueryParams) *goqu.SelectDataset {
+	searchFilter := p.SearchFilter
+	searchValue := p.SearchValue
+	searchExpression := "%" + searchValue + "%"
+	outputQuery := finalQuery
+	if searchFilter == "All" && searchValue != "" {
+		// search by alias -> ids -> ip -> date
+		outputQuery = finalQuery.Where(
+			goqu.Or(goqu.I("alias").ILike(searchExpression),
+				goqu.I("id").ILike(searchExpression),
+				goqu.L("text(ip)").Like(searchExpression)))
+	} else if searchFilter != "" && searchValue != "" {
+		if searchFilter == "ip" {
+			outputQuery = finalQuery.Where(
+				goqu.L("text(ip)").Like(searchExpression))
+		} else {
+			outputQuery = finalQuery.Where(
+				goqu.I(searchFilter).ILike(searchExpression))
+		}
+	}
+	return outputQuery
+}
+
 // GetInstances returns all instances that match with the provided criteria.
 func (api *API) GetInstances(p InstancesQueryParams, duration string) (InstancesWithTotal, error) {
 	var instances []*Instance
@@ -363,9 +388,7 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
 	instancesQuery = instancesQuery.Select("id", "ip", "created_ts", goqu.Case().
 		When(goqu.C("alias").Neq(""), goqu.C("alias")).Else(goqu.C("id")).As("alias"))
 
-	instanceAppQuery := goqu.From("instance_application").
-		Select("version", "status", "last_check_for_updates", "last_update_version", "update_in_progress", "application_id", "group_id", "instance_id")
-
+	instanceAppQuery := prepareInstanceAppQuery()
 	finalQuery := prepareGetInstancesQuery(instancesQuery, instanceAppQuery)
 	if sortOrder == sortOrderAsc {
 		finalQuery = finalQuery.Order(goqu.I(sortFilter).Asc().NullsLast())
@@ -373,6 +396,7 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
 		finalQuery = finalQuery.Order(goqu.I(sortFilter).Desc().NullsLast())
 	}
 
+	finalQuery = prepareSearchQuery(finalQuery, p)
 	query, _, err := finalQuery.
 		Limit(limit).
 		Offset(offset).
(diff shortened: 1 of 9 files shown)


ashu8912:
There is one small issue right now with the pr is the way it scans values into instance, I am facing some issues with StructScan right now will try to look into it tomorrow

ashu8912:
**Search without classifier**

- [x] Search by alias
- [x] Search by id
- [x] Search by ip
- [ ] Search by date

**Search With classifier**

- [ ] Search by alias
- [ ] Search by id
- [ ] Search by ip
- [ ] Search by date

backend/pkg/api/instances.go
```diff
@@ -412,55 +359,21 @@ func (api *API) GetInstances(p InstancesQueryParams, duration string) (Instances
...
-		if sortOrder == sortOrderAsc {
-			instancesQuery = instancesQuery.Order(goqu.I(sortFilter).Asc().NullsLast())
-		} else if sortOrder == sortOrderDesc {
-			instancesQuery = instancesQuery.Order(goqu.I(sortFilter).Desc().NullsLast())
-		}
+	instancesQuery = instancesQuery.Select("id", "ip", "created_ts", goqu.Case().
+		When(goqu.C("alias").Neq(""), goqu.C("alias")).Else(goqu.C("id")).As("alias"))
+
+	instanceAppQuery := goqu.From("instance_application").
+		Select("version", "status", "last_check_for_updates", "last_update_version", "update_in_progress", "application_id", "group_id", "instance_id")
```
joaquimrocha, line 392: Why isn't it selecting *?

backend/pkg/api/instances.go
```diff
@@ -414,7 +439,17 @@ func (api *API) GetInstancesCount(p InstancesQueryParams, duration string) (uint
 	if err != nil {
 		return 0, err
 	}
-	countQuery, _, err := api.getFilterInstancesQuery(goqu.L("COUNT (*)"), p, dbDuration).ToSQL()
+	instancesQuery := api.instancesQuery(p, dbDuration)
+	instancesQuery = instancesQuery.Select("id", "ip", "created_ts", goqu.Case().
+		When(goqu.C("alias").Neq(""), goqu.C("alias")).Else(goqu.C("id")).As("alias"))
+
+	instanceAppQuery := goqu.From("instance_application").
+		Select("version", "status", "last_check_for_updates", "last_update_version", "update_in_progress", "application_id", "group_id", "instance_id")
```
joaquimrocha, line 447: This is the repeated columns that I asked about in the previous comment. It's worth at least abstracting, if we cannot just use *.

frontend/src/js/utils/helpers.ts
```diff
@@ -8,6 +8,28 @@ import React from 'react';
...
+    queryValue: 'alias',
+  },
+  {
+    name: 'name',
+    queryValue: 'alias',
+  },
+  {
+    name: 'date',
+    queryValue: 'last_check_for_updates',
+  },
```
joaquimrocha, line 31: I understood you haven't covered the date search yet. If that's the case, please remove this criterium.

backend/pkg/api/instances.go
```diff
@@ -394,6 +336,47 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	searchFilter := p.SearchFilter
+	searchValue := p.SearchValue
+	searchExpression := searchValue + "%"
+	outputQuery := finalQuery
+	parsedTime, err := time.Parse(time.RFC1123, searchValue)
+	// get a day ahead time from searchValue so that we have a limit while searching
+	oneDayAheadTime := parsedTime.Add(24 * time.Hour)
+	dateSearchQuery := goqu.And(
+		goqu.I(searchFilter).Gte(parsedTime),
+		goqu.I(searchFilter).Lte(oneDayAheadTime))
```
yolossn, line 356: This can be moved to the `else if searchFilter == "last_check_for_updates"` block.

ashu8912: thanks, i placed it above thinking it will be reused with the All searchFilter type but we have decided for now that we want the classifier search by date and not without it.
I have moved it inside now

backend/pkg/api/instances.go
```diff
@@ -394,6 +336,47 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	searchFilter := p.SearchFilter
+	searchValue := p.SearchValue
+	searchExpression := searchValue + "%"
+	outputQuery := finalQuery
+	if searchFilter == "All" && searchValue != "" {
+		// search by alias -> ids -> ip -> date
+		outputQuery = finalQuery.Where(
+			goqu.Or(goqu.I("alias").Like(searchExpression),
+				goqu.I("id").Like(searchExpression),
+				goqu.L("text(ip)").Like(searchExpression)))
```
yolossn, line 356: Is the search case sensitive or case insensitive?  If it is case insensitive then we must use the ILike operator.

backend/pkg/api/instances.go
```diff
@@ -394,6 +336,47 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+	} else if searchFilter != "" && searchValue != "" {
+		if searchFilter == "ip" {
+			outputQuery = finalQuery.Where(
+				goqu.L("text(ip)").Like(searchExpression))
+		} else if searchFilter == "last_check_for_updates" {
+			parsedTime, err := time.Parse(time.RFC1123, searchValue)
+			// get a day ahead time from searchValue so that we have a limit while searching
+			oneDayAheadTime := parsedTime.Add(24 * time.Hour)
```
yolossn, line 364: I think the error should be handled before, else`parsedTime.Add(24 * time.Hour)` will fail.

backend/pkg/api/instances.go
```diff
@@ -394,6 +336,47 @@ func (api *API) GetInstanceStatusHistory(instanceID, appID, groupID string, limi
...
+			dateSearchQuery := goqu.And(
+				goqu.I(searchFilter).Gte(parsedTime),
+				goqu.I(searchFilter).Lte(oneDayAheadTime))
+			// apply date search query only when there was no error while parsing date
+			if err == nil {
+				outputQuery = finalQuery.Where(dateSearchQuery)
+			}
```
yolossn, line 371: I think you move this block also inside `err==nil` block.

joaquimrocha (approved):
Nitpick: Instead of "improve UI" as the last commit title, just mention that it's moving the search entry to an upper level in the group's details UI.

Also, maybe the search input needs some a11y attributes (aria-label). You can check this with the Lighthouse part of the chromium dev tools, or the MSFT A11y Insights extension.

Maybe also let @yolossn give one last look at the backend part to approve the PR on his side.

Otherwise LGTM and worked great in my tests. Thanks!

yolossn (approved):

PR #441, opened 2021-08-13
backend/pkg/api/instances tests

illume:
# Some test improvements for instances

Improved test coverage for:
- getInstanceApps
- GetInstanceStatusHistory

Also moved an instance related update function to be where the other instance update functions are.

### DIFF
--- a/backend/pkg/api/instances.go
+++ b/backend/pkg/api/instances.go
@@ -577,6 +577,17 @@ func (api *API) updateInstanceStatus(instanceID, appID string, newStatus int) er
 	return api.updateInstanceObjStatus(instance, newStatus)
 }
 
+// InstanceStatusUpdateGranted for an instance and version
+func (api *API) grantUpdate(instance *Instance, version string) error {
+	insertData := make(map[string]interface{})
+	insertData["last_update_granted_ts"] = nowUTC()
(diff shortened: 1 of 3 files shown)


joaquimrocha (commented):
Hi. Why was the function moved (I would assume you'd be able to grant the update from the tests in any case)?
@yolossn , maybe this will require changes to your API PR?

illume:
Mainly because it's closer to the data it operates on and because it decouples it from the updates module. It also makes it a bit clearer with the data dependency updateInstanceData relies on in updateGranted being closer.

backend/pkg/api/groups_test.go
```diff
@@ -204,7 +204,7 @@ func TestGetVersionCountTimeline(t *testing.T) {
 
 	instance, err := a.GetInstance(instanceID, tApp.ID)
 	assert.NoError(t, err)
-	_ = a.grantUpdate(instance, version)
+	_ = a.updateGranted(instance, version)
```
joaquimrocha, line 207: As we chatted today, I'd prefer to keep the imperative tense here to describe we're changing something rather than consulting.
It also makes the diff smaller that way.

joaquimrocha (approved):

PR #445, opened 2021-08-19
backend packages: Simplify check of channel arch in AddPackage

illume:
- simplified a check in packages.AddPackage
- added a test for the check

### DIFF
--- a/backend/pkg/api/packages.go
+++ b/backend/pkg/api/packages.go
@@ -49,6 +49,33 @@ type Package struct {
 	Arch              Arch           `db:"arch" json:"arch"`
 }
 
+// checkMatchingArch returns an error if the arch does not match the channels
+func (api *API) checkMatchingArch(channelIDs StringArray, arch Arch) error {
+	if len(channelIDs) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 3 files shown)


backend/pkg/api/packages.go
```diff
@@ -49,6 +49,33 @@ type Package struct {
...
+// checkMatchingArch returns an error if the arch does not match the channels
+func (api *API) checkMatchingArch(channelIDs StringArray, arch Arch) error {
+	if len(channelIDs) == 0 {
+		return nil
+	}
+
+	query, _, err := goqu.From("channel").
+		Select(goqu.COUNT("*")).
+		Where(goqu.Ex{"id": channelIDs}).
+		Where(goqu.Ex{"arch": goqu.Op{"neq": arch}}).
```
joaquimrocha, line 61: nitpick: Maybe  `goqu.C("arch").Neq(arch)` is clearer.

joaquimrocha (approved):
Approved with a nitpick comment. Address it if you want.

backend/pkg/api/packages.go
illume: Indeed it is! Thanks.

PR #449, opened 2021-08-23
Fix adding groups

joaquimrocha:
- frontend: Add a description for when the channels list is empty
- frontend: Fix setting the group's update per period
- frontend: Add a space between the channel's label and its contents
- frontend: Ensure that packages will show up in the list after created

### DIFF
--- a/frontend/src/js/components/Channels/List.tsx
+++ b/frontend/src/js/components/Channels/List.tsx
@@ -10,6 +10,7 @@ import API from '../../api/API';
 import { Application, Channel, Package } from '../../api/apiDataTypes';
 import { applicationsStore } from '../../stores/Stores';
 import { ARCHES } from '../../utils/helpers';
+import Empty from '../Common/EmptyContent';
 import Loader from '../Common/Loader';
 import ModalButton from '../Common/ModalButton';
 import SectionPaper from '../Common/SectionPaper';
@@ -35,6 +36,7 @@ function ChannelList(props: {
 }) {
   const { application, onEdit } = props;
   const classes = useStyles();
+  const { t } = useTranslation();
 
   function getChannelsPerArch() {
     const perArch: {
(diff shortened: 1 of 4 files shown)


frontend/src/js/components/Packages/List.tsx
```diff
@@ -49,7 +47,7 @@ function List(props: { appID: string }) {
     return function cleanup() {
       applicationsStore.removeChangeListener(onChange);
     };
-  }, [application, packages]);
+  }, [application]);
```
illume, line 50: I guess it's dependent on `props.appID` as well?

frontend/src/js/components/Groups/EditDialog/index.tsx
```diff
@@ -55,7 +55,7 @@ function EditDialog(props: {
       policy_updates_enabled: values.updatesEnabled,
       policy_safe_mode: values.safeMode,
       policy_office_hours: values.onlyOfficeHours,
-      policy_max_updates_per_period: values.maxUpdates,
+      policy_max_updates_per_period: parseInt(values.maxUpdates),
```
illume, line 58: Probably the structure values should be type checked rather than using any.

frontend/src/js/components/Packages/EditDialog.tsx
```diff
@@ -96,6 +96,7 @@ function EditDialog(props: { data: any; show: boolean; create?: boolean; onHide:
       .then(() => {
         props.onHide();
         actions.setSubmitting(false);
+        applicationsStore.emitChange();
```
illume, line 99: I think ApplicationStore.createPackage and ApplicationStore.updatePackage should call this emitChange?

Inside ApplicationStore I think there maybe an error with how promises are used. It seems to call this.getAndUpdateApplication without it being related to the returned promise. Meaning it could be called at any time, rather than resolving at this determined point.

frontend/src/js/components/Channels/List.tsx
```diff
@@ -56,26 +58,35 @@ function ChannelList(props: {
...
-              channel={channel}
-              packages={application.packages || []}
-              showArch={false}
-              handleUpdateChannel={onEdit}
-            />
-          ))}
-        </MuiList>
-      ))}
+      {noChannels ? (
+        <Empty>{t('channels|No channels created for this application yet')}</Empty>
```
illume, line 69: If you wanted less nesting you could use a return statement above the existing one:
```
if (noChannels) {
    return <Empty>{t('channels|No channels created for this application yet')}</Empty>
}
```

frontend/src/js/components/Groups/EditDialog/index.tsx
joaquimrocha: I tried adding types but it's still coming as a string. So I will keep this one with the parse.

joaquimrocha:
Updated after @illume 's suggestions.

frontend/src/js/components/Groups/EditDialog/index.tsx
illume: I added an issue for later about improving types where we use any here: https://github.com/kinvolk/nebraska/issues/451

PR #452, opened 2021-08-24
backend: Use app ID when getting the gorup by track name

### LINKED ISSUE #450: Groups are being identified by their track name in a global level and not per app
We introduced the notion of track name a while ago but this concept has been tied to how Flatcar uses the groups, and thus was not tested with multiple applications in the database. Turns out we're not specifying which app we're looking for when getting the group by track name intrernally, so if there are more than one app with groups having the same track name, it will lead to issues.

joaquimrocha:
Groups may have the same track name but be set to different apps, which
may make it return the wrong group if one doesn't specify the app ID. So
this commit accounts for that.

fixes #450


backend/pkg/api/groups.go
```diff
@@ -359,9 +360,15 @@ func (api *API) GetGroupID(trackName string, arch Arch) (string, error) {
 		cachedGroupsLock.Unlock()
 	}
 
-	cachedGroupID, ok := cachedGroupsRef[GroupDescriptor{Track: trackName, Arch: arch}]
+	// Trim the {} that may surround the ID
+	appIDNoBrackets := appID
```
pothos, line 364: I would trim space here

backend/pkg/api/groups.go
```diff
@@ -359,9 +360,15 @@ func (api *API) GetGroupID(trackName string, arch Arch) (string, error) {
...
+	if len(appIDNoBrackets) > 1 && appIDNoBrackets[0] == '{' {
+		appIDNoBrackets = appIDNoBrackets[1 : len(appIDNoBrackets)-1]
```
pothos, line 366: and then trim it again as last step here

pothos (commented):
Looks good otherwise but I didn't run a test with a Flatcar instance

PR #456, opened 2021-08-27
WIP: backend omaha: Extend update check with a _payload

illume:
# Extend update check with a _payload

Work in progress...

## How to use

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]

### DIFF
--- a/backend/pkg/omaha/omaha.go
+++ b/backend/pkg/omaha/omaha.go
@@ -87,8 +87,51 @@ func getArch(os *omahaSpec.OS, appReq *omahaSpec.AppRequest) api.Arch {
 	return api.ArchAMD64
 }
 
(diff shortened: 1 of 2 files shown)


joaquimrocha:
Since we control the `go-omaha` project, I am thinking maybe it's easier to add this custom field in there. It shouldn't break protocol because it's just extra data that can be interpreted by the client.
Maybe we can even make it a `CDATA` field...

joaquimrocha:
I have been wondering whether to add the extra data tag to the whole response, or to a package.
Having it in the package means that we can have more data from each package, or use it as the actual update payload data, if we're using just one package.
I opened a draft PR for this in https://github.com/kinvolk/go-omaha/pull/7 .

joaquimrocha:
Closing in favor of https://github.com/kinvolk/go-omaha/pull/7

Issue #451, opened 2021-08-24
[RFE] Improve types in frontend


illume:
## Current situation

There's still a few places that need types, and are currently using `any` instead (which effectively turns off the type checker).

## Impact

It will probably find some bugs like the parseInt-missing one found here: https://github.com/kinvolk/nebraska/pull/449

## Ideal future situation

Most places that are using `any` will have more specific types.

## **Implementation options

Add types in places where any is used.

PR #457, opened 2021-08-27
frontend: Improve API types

illume:
# Improve API types

Adding types to the API means that they flow through to the rest of the app and allow better checking.

It's a start on https://github.com/kinvolk/nebraska/issues/451

## How to use

App should behave as before, but also type-check.

## Testing done

Ran `npm run tsc` and the tests.


joaquimrocha:
Awesome, thanks!

PR #458, opened 2021-09-01
updater: Add debug flag to updater.

yolossn:
At present it is difficult to debug errors, the
debug flag can be set to print the raw response
to stdout.

### DIFF
--- a/updater/updater.go
+++ b/updater/updater.go
@@ -69,15 +69,16 @@ type Updater struct {
 
 	appID   string
 	channel string
+	debug   bool
 
 	omahaReqHandler OmahaRequestHandler
 }
 
-func New(omahaURL string, appID string, channel string, instanceID string, instanceVersion string) (*Updater, error) {
-	return NewWithOmahaRequestHandler(omahaURL, appID, channel, instanceID, instanceVersion, NewHttpOmahaReqHandler(omahaURL))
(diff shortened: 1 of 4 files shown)


updater/updater.go
```diff
@@ -69,15 +69,16 @@ type Updater struct {
...
 	channel string
+	debug   bool
 
 	omahaReqHandler OmahaRequestHandler
 }
 
-func New(omahaURL string, appID string, channel string, instanceID string, instanceVersion string) (*Updater, error) {
-	return NewWithOmahaRequestHandler(omahaURL, appID, channel, instanceID, instanceVersion, NewHttpOmahaReqHandler(omahaURL))
+func New(omahaURL string, appID string, channel string, instanceID string, instanceVersion string, debug bool) (*Updater, error) {
+	return NewWithOmahaRequestHandler(omahaURL, appID, channel, instanceID, instanceVersion, debug, NewHttpOmahaReqHandler(omahaURL, debug))
```
joaquimrocha, line 78: Don't make this an arg. Use a setter in the updater.New, and set some config as an arg for the OmahaHandler.
Lots of arguments, especiall when they can be optional, will make functions more difficult to remember and code to read.

joaquimrocha (changes requested):

joaquimrocha:
BTW, we should change this to the updater-lib, but it has conflicts.

PR #462, opened 2021-09-02
Apps with product id

### LINKED ISSUE #455: Support human readable application names (reverse domain names)
## Current situation

Apps are created with a generated universal ID when their stored.

## Impact

We cannot create an application with a given name, and it's almost impossible to remember an application ID.

## Ideal future situation

The [omaha protocol](https://chromium.googlesource.com/chromium/src.git/+/refs/heads/main/docs/updater/protocol_3_1.md#Application-IDs) allows for names to be human readable, and unique to each application by using the reverse domain in it.

## **Implementation options

We should request the application ID when creating it, and allow also to set it besides the UUID, for existing apps. So we start supporting both the flatcar UUID but also e.g. `io.kinvolk.Flatcar`.

joaquimrocha:
fixes #455


backend/pkg/handler/app.go
```diff
@@ -51,7 +52,12 @@ func (h *Handler) CreateApp(ctx echo.Context, params codegen.CreateAppParams) er
 		return ctx.NoContent(http.StatusBadRequest)
 	}
 
-	app := appFromRequest(request.Name, request.Description, "", teamID)
+	desc := ""
+	if request.Description != nil {
+		desc = *request.Description
+	}
```
yolossn, line 58: this logic can be moved to the appFromRequest function, make the description arg pointer to string.

backend/pkg/api/applications.go
```diff
@@ -141,14 +200,20 @@ func (api *API) DeleteApp(appID string) error {
...
 func (api *API) GetApp(appID string) (*Application, error) {
+	realAppID, err := api.GetAppID(appID)
+	if err != nil {
+		return nil, err
+	}
+
 	var app Application
 	query, _, err := goqu.From("application").
-		Where(goqu.C("id").Eq(appID)).ToSQL()
+		Where(goqu.C("id").Eq(realAppID)).ToSQL()
```
yolossn, line 216: The api.GetAppID() part can be skipped by modifying the query.

```go
query, _, err := goqu.From("application").
Where(goqu.Or(goqu.C("id").Eq(appID), goqu.C("product_id").Eq(appID))).ToSQL()

```

backend/pkg/api/applications.go
```diff
@@ -236,13 +301,85 @@ func (api *API) GetApps(teamID string, page, perPage uint64) ([]*Application, er
 	return apps, nil
 }
 
+// updateCachedApps invalidates the cached app IDs in cachedApps and
+// must be called whenever the apps entries are modified.
+func (api *API) updateCachedApps() {
```
yolossn, line 306: The function name and the implementation doesn't match, perhaps the name should be `clearCachedApps()`?

backend/pkg/api/applications.go
```diff
@@ -2,25 +2,47 @@ package api
...
+var (
+	// cachedApps caches the mapping of apps' product-ids and UUIDs to apps' UUIDs.
+	// The UUID -> UUID seeming redundancy is because this way we can use it to
+	// validate also the UUIDs against existing apps.
+	// It must not be modified directly but replaced (atomically or via lock)
+	// by a new map to prevent data races.
+	// An update must be triggered through updateCachedApps() each time
+	// an apps change. A RW lock was chosen to prevent data
+	// races over the pointer itself.
+	cachedApps     appsCache
```
yolossn, line 28: cachedApps gives the perception that apps are cached but only the mapping of productID and appID is.

```suggestion
	cachedProductIDAppID     map[string][string]
```

backend/pkg/api/applications.go
```diff
@@ -120,12 +170,21 @@ func (api *API) UpdateApp(app *Application) error {
...
 func (api *API) DeleteApp(appID string) error {
-	query, _, err := goqu.Delete("application").Where(goqu.C("id").Eq(appID)).ToSQL()
+	var matchStatement goqu.Expression
+	if _, err := uuid.Parse(appID); err == nil {
+		matchStatement = goqu.C("id").Eq(appID)
+	} else {
+		matchStatement = goqu.C("product_id").Eq(null.StringFrom(appID))
+	}
+
+	query, _, err := goqu.Delete("application").Where(matchStatement).ToSQL()
```
yolossn, line 187: The query can be converted to

```go
query, _, err := goqu.Delete("application").
Where(goqu.Or(goqu.C("id").Eq(appID), goqu.C("product_id").Eq(null.StringFrom(appID)))).ToSQL()
```

yolossn: If we dont want to rely on OR statements perhaps the `api.GetAppID(appID)` can be used and we can stick to using only appID for deleting the app?

backend/pkg/api/applications.go
```diff
@@ -236,13 +301,85 @@ func (api *API) GetApps(teamID string, page, perPage uint64) ([]*Application, er
...
+					if err != nil {
+						logger.Warn().Err(err).Msg("Failed to read app from DB")
+					}
+
+					if prodIDPtr := app.ProductID.Ptr(); prodIDPtr != nil {
+						cachedApps[*prodIDPtr] = app.ID
+					}
+
+					// So we can quickly validate the UUID based IDs
+					cachedApps[app.ID] = app.ID
```
yolossn, line 352: Should the logic to trim spaces and {} be added here also? I can see that being done when checking if the key exists in the map.

backend/pkg/api/applications.go
joaquimrocha: The reason why I did it this way was to save one DB op, but we don't delete apps every time, so it should be okay.

backend/pkg/api/applications.go
joaquimrocha: I'll go with `cachedAppIDs` because Map is the type and that's not too relevant as well as it can lead to more changes if we ever change it.

backend/pkg/api/applications.go
joaquimrocha: Good point. I took the same naming as in the groups' caching, but you got a point.

backend/pkg/api/applications.go
joaquimrocha: Not really because at this point we're supposed not to store the app id with the {}

joaquimrocha:
Updated. Thanks for the review @yolossn . Let me know how it looks now.

backend/api/spec.yaml
```diff
@@ -1064,13 +1064,15 @@ components:
       type: object
       required:
         - name
-        - description
+        - product_id
```
yolossn, line 1067: product_id is made required in the backend, but the frontend doesn't have a required check. Testing from frontend with empty product_id breaks.

backend/pkg/api/applications.go
```diff
@@ -90,17 +115,41 @@ func (api *API) AddAppCloning(app *Application, sourceAppID string) (*Applicatio
...
 		}
 	}
-
+	// Even though AddApp will invalidate the cache, we need to do it again here
+	// to prevent eventual race issues.
+	api.clearCachedAppIDs()
 	return app, nil
 }
 
+func validateProductID(productID null.String) error {
```
yolossn, line 124: Can you add a comment with examples of valid and invalid productIDs? Reduces the time taken to process the regex every time.

backend/api/spec.yaml
joaquimrocha: Wooops.

backend/pkg/api/applications.go
```diff
@@ -236,13 +298,85 @@ func (api *API) GetApps(teamID string, page, perPage uint64) ([]*Application, er
...
+	cachedAppsIDsLock.Unlock()
+}
+
+func (api *API) GetAppID(appOrProductID string) (string, error) {
+	var cachedAppsRef appsCache
+	cachedAppsIDsLock.RLock()
+	if cachedAppIDs != nil {
+		// Keep a reference to the map that we found.
+		cachedAppsRef = cachedAppIDs
+	}
```
yolossn, line 317: Looks like the reference to cachedAppIDs is kept even if cachedAppIDs is not nil. [Refer]( 
https://github.com/kinvolk/nebraska/pull/462/files#diff-fcdc98002f32a570ecb40a947009fca724887f3dd96ccacf59a8134ce4f194aeR322-R324) Maybe just keep a reference only once?

backend/pkg/api/applications.go
```diff
@@ -236,13 +298,85 @@ func (api *API) GetApps(teamID string, page, perPage uint64) ([]*Application, er
...
+	cachedAppsIDsLock.RUnlock()
+
+	// Generate map on startup or if invalidated.
+	if cachedAppsRef == nil {
+		cachedAppsIDsLock.Lock()
```
yolossn, line 322: Also instead of acquiring the lock for the whole query to run, can't we create a temp map and assign it to `cachedAppIDs`  at once ?

yolossn: I see that this logic is similar to the one in GetGroupID and I think making these changes will make the logic easy to understand, feel free to ignore it if you think otherwise.

yolossn:
Added few comments, The rest looks good.

joaquimrocha:
Updated.

backend/pkg/api/applications.go
joaquimrocha: If you don't block during the populating phase, then you may have several instances doing the populating phase at the same time. It wouldn't be a concurrency issue since we'd lock when assigning the main map, but it does end up in duplicating the effort (meaning more reads in the database, more memory use just to be discarded on assignment...).

yolossn (approved):

PR #465, opened 2021-09-06
Small UI improvements

joaquimrocha:
- frontend: Add missing "packages" field to the Application interface
- frontend: Complete some missing type information in several components
- frontend: Show an up to date list of packages when assigning a channel
- frontend: Check if version_breakdown is valid before using its length
- frontend: Reflect package creation's result in the new package form
- frontend: Use "applications" as the ns for apps' translations
- frontend: Make "Channels" in the GroupDetailsForm translatable


ashu8912 (approved):
Tested, LGTM!!!

PR #468, opened 2021-09-08
backend: Allow Config to be set both from env and flag

yolossn:
This patch allows the user to use both env and flags to
configure the nebraska server. Note that the flag
overrides the env.


backend/pkg/api/channels.go
```diff
@@ -156,45 +156,6 @@ func (api *API) GetChannel(channelID string) (*Channel, error) {
 	return &channel, nil
 }
 
-func (api *API) getSpecificChannels(channelID ...string) ([]*Channel, error) {
```
joaquimrocha, line 159: This commit was also in the product-id PR, so we should remove it AFAIU.

joaquimrocha (approved):
Assuming this is retro-compatible with the flags we use on `main`, and that the use of one flag doesn't invalidate any used env variables (just simply overrides the corresponding env var), it LGTM.

joaquimrocha (changes requested):
I have tried again and the env vars don't seem to be working.
BTW, I have also just realized that, unless I am confused, the config seems be exported as a package, but I am not sure we want this because config seems very intimate to running nebraska, not something that may be used by other modules IMO.

yolossn:
The code stopped working as expected once I changed from `spf13/pflag` provider to golang standard `flag` to support flags with single `-`. I'll investigate if there are any other ways to handle both cases. 

The `config` package is exported because the `handler` package which is exported is dependent on it. Making `config` internal makes `handler` unusable from any external code.

joaquimrocha (approved):

joaquimrocha (approved):

PR #470, opened 2021-09-13
Fix group edit validations

ashu8912:
## Testing done
Go to group edit view and try invalidating a field and you will be shown with the validation errors

### DIFF
--- a/frontend/src/js/components/Groups/EditDialog/GroupDetailsForm.tsx
+++ b/frontend/src/js/components/Groups/EditDialog/GroupDetailsForm.tsx
@@ -1,5 +1,6 @@
-import { FormControl, Grid, InputLabel, MenuItem, Select, TextField } from '@material-ui/core';
+import { FormControl, Grid, InputLabel, MenuItem, Select } from '@material-ui/core';
 import { Field } from 'formik';
+import { TextField } from 'formik-material-ui';
 import { useTranslation } from 'react-i18next';
 import { Channel } from '../../../api/apiDataTypes';
 import { ARCHES } from '../../../utils/helpers';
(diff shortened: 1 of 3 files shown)


frontend/src/js/components/Groups/EditDialog/FormErrorMessage.tsx
```diff
@@ -0,0 +1,14 @@
+import { Box } from '@material-ui/core';
+import { connect, getIn } from 'formik';
+
+// This component renders an error message if a field has
+// an error and it's already been changed.
+const ErrorMessage = (props: any) => {
+  console.log(props);
```
joaquimrocha, line 7: Don't log this.

frontend/src/js/components/Groups/EditDialog/GroupDetailsForm.tsx
```diff
@@ -64,6 +66,7 @@ export default function GroupDetailsForm(props: {
               setFieldValue('description', e.target.value);
             }}
           />
+          <FormErrorMessage name="description" />
```
joaquimrocha, line 69: I haven't tried the PR but I doubt this will be the same UX as the forms where Formik has in fact worked automatically to provide the errors.
I would like us to figure out why Formik isn't setting up the errors automatically here, but if we cannot, then it seems like we can emulate the same UX by setting the `error` property in `Field`, which should be passed on to the MUI `TextField` component, and set the `helpText` to the error message when there's an error.

If you need more context, please ping me on IM.

joaquimrocha (changes requested):

joaquimrocha (approved):
Awesome!

PR #471, opened 2021-09-13
Data packages

joaquimrocha:


illume:
- 1) 1st commit, fixed it so that the code gen would run on CI
- 2) fixed a few errors, finished the test, switched the `var char 1024` to `text`, and used MetadataContent everywhere rather than 
- 3) made the textarea resizable

The backend currently fails if the newly non-required fields (size, filename, hash) are not included. Not sure if that work is in scope for this PR?

Keeping cleanups, and refactoring out of this PR (such as splitting the long form up).

illume:
> The backend currently fails if the newly non-required fields (size, filename, hash) are not included. Not sure if that work is in scope for this PR?

These Package fields are not required anymore.

joaquimrocha:
Finally to try it out.
The metadata fields are much more usable now. I have noticed however, that some fields still show as mandatory (*) but are not, so they should drop the *.

The flatcar version of the package should keep the required fields as they were before (I think the URL may be have been mistakenly marked as not required).

The way to test is to switch the package type back and forth and tabbing through to see the errored fields.

illume:
> The metadata fields are much more usable now. I have noticed however, that some fields still show as mandatory (*) but are not, so they should drop the *.

Ok, let me check. Is the schema at backend/api/spec.yaml considered the source of truth? I mean, what's shown there as required should be shown in the form as required?

> The flatcar version of the package should keep the required fields as they were before (I think the URL may be have been mistakenly marked as not required).

The filename, hash, and size should stay required for Flatcar packages, but not for the "Other" package type? If so, I'll have to see if that's possible to specify in the schema somehow.


ps. One unrelated issue I noticed is that the create form doesn't show an error when there's an error in the backend. Also the update form simply closes when there is a backend error.

joaquimrocha:
> > The metadata fields are much more usable now. I have noticed however, that some fields still show as mandatory (*) but are not, so they should drop the *.
> 
> Ok, let me check. Is the schema at backend/api/spec.yaml considered the source of truth? I mean, what's shown there as required should be shown in the form as required?

Since the endpoint is the same for all types of packages, I understand we cannot do the distinction in the spec.
However, we can match the spec to the least mandatory type of packages, and then dynamically check it in the function that implements the endpoint.

> 
> > The flatcar version of the package should keep the required fields as they were before (I think the URL may be have been mistakenly marked as not required).
> 
> The filename, hash, and size should stay required for Flatcar packages, but not for the "Other" package type? If so, I'll have to see if that's possible to specify in the schema somehow.

I had already some client-side logic in the branch for that purpose.

> 
> ps. One unrelated issue I noticed is that the create form doesn't show an error when there's an error in the backend. Also the update form simply closes when there is a backend error.

Yeah, showing errors to clients in forms is one of the painpoints we currently have in Nebraska. Most of them just show a generic error, making it difficult to understand what exactly was wrong.
We should tackle this in a different PR.

illume:
I tried using the [OpenAPI discriminator feature](https://swagger.io/specification/#discriminator-object
) which should let us use a different schema based on the type. However, [it doesn't seem to work](https://github.com/deepmap/oapi-codegen#whats-missing-or-incomplete) with `oapi-codegen`. Oh well, that would have been nice. Just going to check that manually in the handler for now.

```diff
diff --git a/backend/api/spec.yaml b/backend/api/spec.yaml
index 2d97527..88b305f 100644
--- a/backend/api/spec.yaml
+++ b/backend/api/spec.yaml
@@ -684,7 +684,14 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/createPackageInfo"
+              oneOf:
+                - $ref: "#/components/schemas/createPackageInfo"
+                - $ref: "#/components/schemas/createPackageInfoFlatcar"
+              discriminator:
+                propertyName: type
+                mapping:
+                  4: '#/components/schemas/createPackageInfo'
+                  1: '#/components/schemas/createPackageInfoFlatcar'
       responses:
         "200":
           description: Create Package success response
@@ -748,7 +755,14 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/updatePackageInfo"
+              oneOf:
+                - $ref: "#/components/schemas/updatePackageInfo"
+                - $ref: "#/components/schemas/updatePackageInfoFlatcar"
+              discriminator:
+                propertyName: type
+                mapping:
+                  4: '#/components/schemas/updatePackageInfo'
+                  1: '#/components/schemas/updatePackageInfoFlatcar'
       responses:
         "200":
           description: Update Package success response
@@ -1161,10 +1175,52 @@ components:
           type: integer
         flatcarAction:
           $ref: "#/components/schemas/flatcarActionPackage"
-        
+
+    createPackageInfoFlatcar:
+      type: object
+      required:
+        - application_id
+        - arch
+        - channels_blacklist
+        - description
+        - filename
+        - hash
+        - size
+        - type
+        - url
+        - version
+      properties:
+        application_id:
+          type: string
+        arch:
+          type: integer
+        channels_blacklist:
+          type: array
+          items:
+            type: string
+        description:
+          type: string
+        filename:
+          type: string
+        hash:
+          type: string
+        size:
+          type: string
+        url:
+          type: string
+        version:
+          type: string
+        type:
+          type: integer
+        flatcarAction:
+          $ref: "#/components/schemas/flatcarActionPackage"
+
     updatePackageInfo:
       $ref:  "#/components/schemas/createPackageInfo"
 
+    updatePackageInfoFlatcar:
+      $ref:  "#/components/schemas/createPackageInfoFlatcar"
+
     updateInstanceInfo:
       type: object
       required:
```

illume:
Seems the proof_of_concept branch was force pushed(on purpose I guess). I'll rebase against the new branch?

joaquimrocha:
> Seems the proof_of_concept branch was force pushed(on purpose I guess). I'll rebase against the new branch?

Yes please. Shouldn't be complicated. As we mostly squashed stuff and rebadged with main.

illume:
Cool. It should be ready now, with the required fields changes.

frontend/src/js/components/Packages/EditDialog.tsx
```diff
@@ -313,7 +313,7 @@ function EditDialog(props: { data: any; show: boolean; create?: boolean; onHide:
   const validation: {
     [key: string]: any;
   } = Yup.object().shape({
-    url: Yup.string().url(),
+    url: Yup.string().url().required(t('frequent|Required')),
```
joaquimrocha, line 316: I am thinking we shouldn't make the URL mandatory for non-flatcar packages because a URL means we're expecting to download something from it, but maybe that info will be in the json passed as metadata.

frontend/src/js/components/Packages/EditDialog.tsx
```diff
@@ -230,7 +235,32 @@ function EditDialog(props: { data: any; show: boolean; create?: boolean; onHide:
...
             />
+          ) : (
+            <>
+              <Field
+                name="metadataType"
+                component={TextField}
+                margin="dense"
+                label={t('packages|Metadata Content Type')}
+                type="text"
+                helperText={t('packages|E.g. content/json')}
```
joaquimrocha, line 246: I know it was my mistake, but the example should read "text/json".

joaquimrocha (commented):
Thanks @illume .
I left two minor comments. I also tested it and unfortunately it seems like storing the metadata (and type) is not working (in my test I switched between packages types in the dialog, and stored an "Other" package with contents in the metadata's related fields).

illume:
@joaquimrocha this should be ready to review.

joaquimrocha (commented):
The changes LGTM. I found though that I cannot assign a channel to a group.
This is not caused by this PR, as far as I can see but should be addressed (in another PR is okay).

illume:
Cool, thanks.

I'll fix these conflicts.

illume:
The weirdness on this branch/PR because of the forced push to proof-of-concept has been fixed.

PR #475, opened 2021-09-17
Updater example tests blackbox

joaquimrocha:
I made the tests part of the updater package when I stopped exporting the NewUpdateInfo method but still needed to test it.
However, I mistakenly made the example tests also part of the updater package, and those should indeed be black box style.

This commit fixes that.


yolossn (approved):
LGTM

PR #477, opened 2021-09-20
frontend: Move the group's track above the description in edit form

joaquimrocha:
The group's track is more important than the group's description and
it's easy to miss as it is the last field in the form. So this patch
simply moves it above the description in the creation/edition form.


notpygame (approved):
👍

illume (approved):
👍

PR #481, opened 2021-09-22
Remove updaters legacy code

joaquimrocha:
We have not being maintaining this code and it's not used, so let's
removed it.


yolossn (approved):
LGTM

PR #490, opened 2021-10-04
WIP: Go SDK

yolossn:


go-sdk/application.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return fmt.Errorf("deleting app %q: %w", appID, err)
+	}
+	if resp.StatusCode() != http.StatusNoContent {
+		return fmt.Errorf("DeleteApp %q returned invalid response code: %d", appID, resp.StatusCode())
+	}
+	return nil
+}
+
+func (n *Nebraska) PaginateApps(ctx context.Context, page int, perPage int, reqEditors ...RequestEditorFn) (*ApplicationsResponse, error) {
```
invidian, line 137: (one-off comment, more to come)

Maybe rather than paginate, something conceptually simpler, like `ListApps(limit, offset)` ?

invidian:
Thanks for opening the PR quickly @yolossn! I checked out the code and played a bit with it (though without Nebraska instance) and I produced something like this: https://github.com/kinvolk/nebraska/compare/go-sdk...invidian:invidian/go-sdk?expand=1

The code is likely heavily inspired by Kubernetes [Clientset](https://pkg.go.dev/k8s.io/client-go/kubernetes#Clientset), but I believe it's a decent source of inspiration, as I worked with this client a lot, also from testing perspective and I never really run into some annoyances or limitations. Let me know what you think!

Ah, also sorry for stripping out the actual implementation, but due to time constraints, i didn't have time to restore it, I only wanted code to compile to make sure all types are compatible.

yolossn:
Hey, @invidian  Thanks for the example, the `ApplicationClient` and `GroupClient` make sense, unlike adding the `NebraskaClient` to each object, this method reduces the scope of functions available under each Object.

go-sdk/application.go
yolossn: Good idea but I doubt if we can support `limit` and `offset` from the existing APIs. We also can't repurpose the Paginate endpoint to support this, for example, if the `offset` is `50` and `limit` is `20` we cant support it. Still if we want to support this we will need to add a soft check if the offset is divisible by limit which can end up being weird for the end-user.

yolossn:
@invidian @joaquimrocha 
I have made the changes based on the comments. At present Applications and Groups are covered, wanted to get another round of feedback before moving to the Channels, Packages, etc. Please take a look.

go-sdk/applicationClient.go
```diff
@@ -0,0 +1,195 @@
...
+
+type applicationClient struct {
+	config *Config
+	client *codegen.ClientWithResponses
+}
+
+func (ac *applicationClient) Create(ctx context.Context, conf AppConfig, opts *CreateAppOptions) (Application, error) {
+	if opts == nil {
+		opts = &CreateAppOptions{}
+	}
```
invidian, line 53: Seems like passing `opts` by value could save us a lot of code and tests (similar to what Kubernetes clientset is doing).

go-sdk/application.go
```diff
@@ -0,0 +1,54 @@
...
+	Props() codegen.Application
+	CreateGroup(ctx context.Context, conf GroupConfig, opts *CreateGroupOptions) (Group, error)
+	PaginateGroups(ctx context.Context, page int, perpage int, opts *PaginateGroupOptions) (*Groups, error)
+	Response() *http.Response
+}
+
+type Applications struct {
+	TotalCount int
+	Apps       []Application
+	rawResponse
```
invidian, line 22: Is this ever used?

go-sdk/applicationClient.go
```diff
@@ -0,0 +1,195 @@
...
+	type appsPage struct {
+		Applications []json.RawMessage `json:"applications"`
+		Count        int               `json:"count"`
+		TotalCount   int               `json:"totalCount"`
+	}
+
+	var rawApps appsPage
+
+	if !strings.Contains(resp.Header.Get("Content-Type"), "json") {
+		return nil, fmt.Errorf("invalid applications response content-type: %q", resp.Header.Get("Content-Type"))
```
invidian, line 168: I'd add here what type did we expect, like:
```
expected response content-type %q, got %q
```

go-sdk/applicationClient.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	decoder := json.NewDecoder(resp.Body)
```
invidian, line 171: Why do we need to use the decoder here, if we create another decoders later? Can't we simply just use `Unmarshal`?

go-sdk/groupClient.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+type GroupConfig codegen.CreateGroupJSONBody
+
+type CreateGroupOptions struct {
+	commonOptions
+}
+type GetGroupOptions struct {
+	commonOptions
+}
```
invidian, line 21: I wonder if we need resource-specific type here or maybe we could just have `GetOptions`

go-sdk/groupClient.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (gc *groupsClient) Create(ctx context.Context, conf GroupConfig, opts *CreateGroupOptions) (Group, error) {
+	if opts == nil {
+		opts = &CreateGroupOptions{}
+	}
+
+	resp, err := gc.client.CreateGroup(ctx, gc.appID, codegen.CreateGroupJSONRequestBody(conf), convertReqEditors(opts.RequestEditors...)...)
+	if err != nil {
+		return nil, fmt.Errorf("creating group app id %q returned %w", gc.appID, err)
```
invidian, line 53: I'd stick to the standard error formatting:
```
creating group app id %q: %w
```

go-sdk/nebraska.go
```diff
@@ -0,0 +1,70 @@
...
+	config *Config
+	client *codegen.ClientWithResponses
+}
+
+type RequestEditorFn codegen.RequestEditorFn
+
+func New(conf Config) (Nebraska, error) {
+	// Validate Server URL
+	_, err := url.Parse(conf.ServerURL)
+	if err != nil {
```
invidian, line 32: This could be written as a single line.

go-sdk/nebraska.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, fmt.Errorf("invalid server url: %w", err)
+	}
+
+	// If the authMode is not noop UserName and Password is necessary.
+	client, err := codegen.NewClientWithResponses(conf.ServerURL)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't init client: %w", err)
+	}
+
+	serverConfig, err := client.GetConfigWithResponse(context.TODO())
```
invidian, line 42: We should take context as a parameter for this function.

invidian (commented):
Just some minor nits, otherwise looks quite good! I'd really like to see some more tests for it, to make sure bits and pieces can be easily mocked for tests, as I think some code might be redundant, e.g. testing content-type headers is done in couple of places, it could be refactored at least.

There might be a bit more styling which could be improved, but that can be done later.

PR #503, opened 2021-10-26
chore: Split frontend and backend build parts

guilhem:
# Split frontend and backend build parts

## Makefile

- `make` will call `Makefile` in _frontend_ and _backend_
- `make run` call `run` in both _frontend_ and _backend_

## Dockerfile

- use _distroless_ for run
  - use non-root user
- make binary static
  - and make it smaller
- use _node_ image for frontend build
- use caching
  - remove `non-cache` build parameter
  - improve `.dockerignore`
  - use docker/dockerfile:1.3 syntax for better caching

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -30,5 +30,5 @@ jobs:
 
     - name: Build, Lint & Check
       run: |
-        cd $GITHUB_WORKSPACE
-        make backend-ci
+        make ci
+      working-directory: backend
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -26,7 +26,7 @@ jobs:
       id: go
 
     - name: Check out code
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     - name: Build, Lint & Check
       run: |
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -13,9 +13,16 @@ jobs:
     name: Build Container Image
     runs-on: ubuntu-20.04
     steps:
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
     - name: Check out code
       uses: actions/checkout@v2
     - name: Build image
+      env:
+        DOCKER_BUILDKIT: 1
       run: |
         cd $GITHUB_WORKSPACE
         make container
--- a/frontend/Makefile
+++ b/frontend/Makefile
@@ -1,0 +1,42 @@
+.PHONY: frontend
+frontend: build
+
+.PHONY: watch
+watch: run
+
+.PHONY: run
+run: node_modules
+	npm start
+
+.PHONY: install
+install: node_modules
+
+.PHONY: install-ci
+install-ci:
+	npm ci
+
+.PHONY: build
+build: install
+	npm run build
+
+.PHONY: test
+test:
+	npm run test
+
+.PHONY: lint
+lint:
+	npm run lint
+
+.PHONY: tsc
+tsc:
+	npm run tsc
+
+.PHONY: i18n
+i18n:
+	npm run i18n
+
+.PHONY: ci
+ci: install-ci lint test build tsc
+
+node_modules:
+	npm install
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -17,7 +17,7 @@ jobs:
         node-version: [12.x, 14.x, 16.x]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v1
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -24,22 +24,7 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
 
-    - name: Install dependencies
+    - name: run ci
       run: |
-        make frontend-install-ci
-
-    - name: Run linter
-      run: |
-        make frontend-lint
-
-    - name: Run test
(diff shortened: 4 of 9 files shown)


.github/workflows/build-container.yml
```diff
@@ -13,9 +13,16 @@ jobs:
     name: Build Container Image
     runs-on: ubuntu-20.04
     steps:
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
```
joaquimrocha, line 20: Is this needed because of the changes to how the image is built? If that's the case then maybe the changes should be squashed together.

.github/workflows/backend.yml
```diff
@@ -26,9 +26,9 @@ jobs:
       id: go
 
     - name: Check out code
-      uses: actions/checkout@v1
```
joaquimrocha, line 29: Commenting here but it's for the whole patch: these changes are a good improvement, both the splitting & the overall improvements.
Since they have more than just moving things around, it's more difficult to review and catch any potential issues, but I know splitting the commit like that now is even a bigger chore.

I would like though to keep the old targets for compatibility (we are calling these from scripts and may have users doing so as well).

@guilhem , do you think you can do these updates?

joaquimrocha (changes requested):
@guilhem Thanks for another nice set of changes. I left some comments.

.github/workflows/backend.yml
guilhem: Yes, splitting van be challenging, but I can try :)

I tried to keep compatibility by forwarding target to nested Makefile... unsuccessfully.
But I can modify old targets for compatibility.

.github/workflows/build-container.yml
guilhem: "Set up QEMU" is not necessary.
It's here if we want to do multi arch (cross arch building).

.github/workflows/backend.yml
guilhem: @joaquimrocha in some way I spited commits.
It may be better, but let me know.

frontend/Makefile
```diff
@@ -0,0 +1,40 @@
+.PHONY: frontend
+frontend: build
+
+.PHONY: watch
+watch: run
+
+.PHONY: run
+run: install
```
joaquimrocha, line 8: You added this dependency which makes the run command very slow. I'd prefer to do it based on whether frontend/node_modules exist. This doesn't ensure we're running the latest install, but `make run` is a development command so we can expect a level of independence/responsibility for the developer.

joaquimrocha (changes requested):
Thanks for the changes @guilhem !
I left a couple of small comments. I also ran the image build command, both locally with podman and in a Github workflow, and I am getting an issue when mounting the cache:

Respectively:
```
Error: error building at STEP "RUN --mount=type=cache,target=$GOPATH/pkg/mod make build": error resolving mountpoints for container "99a4f2744073233ac657940770d935f669a8c03e8f55df3178c6d7a98d560d51": invalid filesystem type "cache"
make: *** [Makefile:109: image] Error 125
```

```
RUN --mount=type=cache,target=$GOPATH/pkg/mod 	make build
the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled
make: *** [Makefile:109: image] Error 1

Problem making Nebraska base image!
```

Maybe it's a good idea to split out the image build changes to a different PR, if you want the makefile related ones to be merged sooner.

frontend/Makefile
guilhem: For buildah, a patch fixed it... but not yet released :( https://github.com/containers/buildah/pull/3574

For GitHub actions, that's why I added a special action that install buildkit.

guilhem:
For buildah, a patch fixed it... but not yet released :( https://github.com/containers/buildah/pull/3574

For GitHub actions, that's why I added a special action that install buildkit.

joaquimrocha:
@guilhem , I have tried again with an updated podman and I could build it.
However, with docker in my Fedora box I have permissions issues when mounting the cache. Not sure I am doing something wrong or whether it's docker's version (from the moby-engine 20.10.10).

I wonder if the docker related changes will become an issue for other users too. What do you think?

We can also always extract the docker related changes into a new PR and merge the rest, as the Makefile LGTM now.

guilhem:
> I wonder if the docker related changes will become an issue for other users too. What do you think?

I tried to do it the "actual way" (it was some kind of challenge too).
But yes, some people without latest buildkit can have problems.

> We can also always extract the docker related changes into a new PR and merge the rest, as the Makefile LGTM now.

It sound a really good idea to me :)

joaquimrocha:
> > We can also always extract the docker related changes into a new PR and merge the rest, as the Makefile LGTM now.
> 
> It sound a really good idea to me :)

Thanks @guilhem . Since it was not clear who should do this, do you want to extract the Docker commits into a new PR yourself?
If you don't have time, let me know.

joaquimrocha:
> > > We can also always extract the docker related changes into a new PR and merge the rest, as the Makefile LGTM now.
> > 
> > 
> > It sound a really good idea to me :)
> 
> Thanks @guilhem . Since it was not clear who should do this, do you want to extract the Docker commits into a new PR yourself? If you don't have time, let me know.

@guilhem , do you have time to split the docker changes out?

guilhem:
> > > > We can also always extract the docker related changes into a new PR and merge the rest, as the Makefile LGTM now.
> > > 
> > > 
> > > It sound a really good idea to me :)
> > 
> > 
> > Thanks @guilhem . Since it was not clear who should do this, do you want to extract the Docker commits into a new PR yourself? If you don't have time, let me know.
> 
> @guilhem , do you have time to split the docker changes out?

Working on it :)

joaquimrocha (approved):

joaquimrocha:
@guilhem Awesome! 🥳

Issue #507, opened 2021-11-03
DB index, and table bloat


illume:
## Current situation

There are issues with database index and table size bloat. Leading to service slowdown, and the database running out of resources (memory) and downtime.

## What we know

- Without the REINDEX, The index are bloating from 2x to 7x the size it should be over several days. Note, no autovaccuum or autoanalyze ran.
- autovacuum is not running (most likely because it's based on table size, rather than number of rows changed). Meaning garbage collection can't happen.
- autoanalyze is not running(samesies as autovacuum). Meaning query statistics are not updated. 
- Cleanup script was running on boot, but didn't trigger a REINDEX after the script was changed.
- The cleanup script with REINDEX was making several ccnew invalid transient index which were using up a lot of resources. After removing the REINDEX and VACUUM ANALYZE from the cleanup script, no new ccnew index were made. This confirms the REINDEX was creating them. https://www.postgresql.org/docs/13/sql-reindex.html
- PostgreSQL 13 is available on RDS (we are using 12). This has index size improvements via deduplication.
- PostgreSQL 14 is not on AWS yet. This has autovacuum improvements on delete.
- There are index size improvements on a staging server when it was updated to pg 13.13 from 12.x and REINDEX'd. See [pg 13 release notes](https://www.postgresql.org/docs/release/13.0/). "Users upgrading with pg_upgrade will need to use REINDEX to make an existing index use this feature."
- auto vacuum and auto analyze threshold changes are enough to stop the table bloat, and significantly reduce the index bloat (from 10x to 2x).

## What we don't know.

- Something else may be creating invvalid *ccnew index (I don't think we have logging to find out what did this)
logging says the creation script ran, but with the REINDEX commented out. It could be that a REINDEX was still running from before the script stopped. Because reindexing concurrently can run for quite a long time.
- We don't know for sure what is blocking REINDEX on some tables. Possibly the vacuum analyze ran in the older script... or something else. ~~We could confirm if the VACUUM ANALYZE is stopping the REINDEX by running the original cleanup script on a staging server. [ED: this does not reproduce the issue]~~
- We can monitor the progress of a REINDEX with https://www.postgresql.org/docs/current/progress-reporting.html#CREATE-INDEX-PROGRESS-REPORTING
  - `pg_stat_progress_create_index` view and columns (lockers_total, lockers_done and current_locker_pid)
        select pid, 
            usename, 
            pg_blocking_pids(pid) as blocked_by, 
            query as blocked_query
[...]

PR #508, opened 2021-11-03
DB Autovacuum, and Autoanalyze thresholds

illume:
This hopefully addresses the DB table/index bloat size issue https://github.com/kinvolk/nebraska/issues/507

Because autovacuum and autoanalyze are effectively disabled when tables
are very large. This is because the default is 20% of a table (and 10%
of a table for analyze).

Instead we change it to when about 5,000 rows change. This value was chosen
based on getting the autovacuum to run every day. Because it's large enough
to not cause the autovac to run all the time, but about the right size
to make a difference for query statistics and reducing table bloat.

The analyze threshold was chosen at half the autovacuum threshold
because the defaults are set at half.

Ultimately there should be some sort of auto-tuning done where the table growth/change rates are calculated and then these thresholds are updated. It has happened where a table gets a 5x growth over other months inside one month. It could be even greater. If it's only 10x growth then the autovac would only run 10 times per day, which is acceptable.

## Testing done

Ran some tests to reproduce index bloat without the thresholds. https://github.com/kinvolk/nebraska/issues/507#issuecomment-964522622 Then confirmed that table bloat was almost reduced entirely, auto analyze was running and index bloat was reduced to 2x compared to 10x observed.

- [x] Check the table growth rates.
- [x] Check if there are other tables to add to this list. Only the tables from the cleanup script were chosen.

### DIFF
--- a/backend/pkg/api/db/migrations/0017_autovac_thresholds.sql
+++ b/backend/pkg/api/db/migrations/0017_autovac_thresholds.sql
@@ -1,0 +1,104 @@
+-- +migrate Up
+
+ALTER TABLE
+    instance
+SET
+    (
+        autovacuum_analyze_scale_factor = 0,
+        autovacuum_analyze_threshold = 3666,
+        autovacuum_vacuum_scale_factor = 0,
+        autovacuum_vacuum_threshold = 7333
+    );
+
+
+ALTER TABLE
+    instance_application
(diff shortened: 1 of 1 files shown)


illume:
The lint fail is because the proof-of-concept branch is failing `make backend-code-check`.

PR #513, opened 2021-11-17
WIP: proof-of-concept merge

illume:
I reduced the proof-of-concept down to 7 commits without the meta data stuff. I left the proof-of-concept branch there, and made a new proof-of-concept2 branch.

### Four groups of commits

- backend: Add OIDC Login Token auth endpoint

---

- Makefile: Run codegen in backend target

---

- frontend: Allow to change and show application.product_id
- backend: Validate apps' product-id format before setting the app
- backend: Use a cache for quickly getting the apps' IDs
- backend: Add human-readable applications.product_id

---

- backend: Replace gin with oapi-codegen and echo

## Testing done

Had CI run on each commit, so each commit passes all checks.


backend/pkg/api/applications.go
```diff
@@ -222,13 +303,85 @@ func (api *API) GetApps(teamID string, page, perPage uint64) ([]*Application, er
...
+// must be called whenever the apps entries are modified.
+func (api *API) clearCachedAppIDs() {
+	cachedAppsIDsLock.Lock()
+	cachedAppIDs = nil
+	// Generating the map is not always possible here because the database
+	// can be closed.
+	cachedAppsIDsLock.Unlock()
+}
+
+func (api *API) GetAppID(appOrProductID string) (string, error) {
```
illume, line 316: Should GetAppId be used in more places where ever appID is used? One example is channels.go@GetChannels which takes an appID.

backend/pkg/api/groups.go
```diff
@@ -383,6 +383,21 @@ func (api *API) updateCachedGroups() {
...
 }
 
+// GetGroupsCount retuns the total number of groups in an app
+func (api *API) GetGroupsCount(appID string) (int, error) {
+	query, _, err := goqu.From("groups").Where(goqu.C("application_id").Eq(appID)).Select(goqu.L("count(*)")).ToSQL()
+	if err != nil {
+		return 0, err
+	}
+	count := 0
+	err = api.db.QueryRow(query).Scan(&count)
```
illume, line 393: There's several copies of this count code. Could reduce much of it with a helper.

```golang
  func (api *API) GetGroupsCount(appID string) (int, error) {
    query, _, err := goqu.From("groups").Where(goqu.C("application_id").Eq(appID)).Select(goqu.L("count(*)")).ToSQL()
    if err != nil {
      return 0, err
    }
    count := 0
    err = api.db.QueryRow(query).Scan(&count)

    if err != nil {
      return 0, err
    }
    return count, nil
  }
```
Each one could be reduced to:
```golang
  func (api *API) GetGroupsCount(appID string) (int, error) {
    sql, _, err := goqu.From("groups").Where(goqu.C("application_id").Eq(appID)).Select(goqu.L("count(*)")).ToSQL()

    return GetCountSQL(sql, err);
  }
```

backend/pkg/middleware/auth.go
```diff
@@ -0,0 +1,51 @@
+package middleware
+
+import (
+	"github.com/labstack/echo/v4"
+	"github.com/labstack/echo/v4/middleware"
+
+	"github.com/kinvolk/nebraska/backend/pkg/auth"
+)
+
+func NewAuthSkipper(auth string) middleware.Skipper {
```
illume, line 10: Looks like this function skips authentication for some URLs?

backend/pkg/middleware/auth.go
```diff
@@ -0,0 +1,51 @@
...
+	return func(c echo.Context) bool {
+		switch auth {
+		case "oidc":
+			paths := []string{"/login", "/config", "/*", "/login/cb", "/login/token", "/v1/update"}
```
illume, line 14: Is /* supposed to be in here?

backend/pkg/middleware/path.go
```diff
@@ -0,0 +1,18 @@
+package middleware
+
+import (
+	"strings"
+
+	"github.com/labstack/echo/v4"
+)
+
+func SanitizePath() echo.MiddlewareFunc {
```
illume, line 9: What does this middleware do? Docs could make this clear.

backend/pkg/sessions/echo/echo.go
```diff
@@ -0,0 +1,32 @@
...
+			c.Set("session", session)
+			defer func() {
+				s.PutSessionUse(session)
+			}()
+			return next(c)
+		}
+	}
+}
+
+func GetSession(c echo.Context) *sessions.Session {
```
illume, line 22: Seems users of GetSession need to check if the result is `nil`, but they are not.

backend/cmd/nebraska/main.go
```diff
@@ -0,0 +1,271 @@
...
+			split := strings.Split(token, " ")
+			if len(split) == 2 {
+				if split[0] != "Bearer" {
+					return errors.New("Bearer token not found in request")
+				}
+			} else {
+				return errors.New("Invalid Bearer token")
+			}
+			return nil
+		}
```
illume, line 268: Does this function need a `"github"` case?

backend/pkg/api/db/migrations/0014_add_application_product_id.sql
```diff
@@ -0,0 +1,9 @@
+-- +migrate Up
+
+alter table application add column product_id varchar(255) default null;
```
illume, line 3: I changed this here to 255, since fully qualified domain names like the appID can be that long.

backend/pkg/api/applications.go
```diff
@@ -90,17 +115,46 @@ func (api *API) AddAppCloning(app *Application, sourceAppID string) (*Applicatio
...
+	if productID.Ptr() == nil {
+		return nil
+	}
+
+	// This regex matches an ID that matches
+	// * At least two segments.
+	// * All characters must be alphanumeric, an underscore, or a dash [a-zA-Z0-9_].
+	// Each segment must start with a letter.
+	// Each segment must not end with an underscore or dash.
+	matches, err := regexp.MatchString("^[a-zA-Z]+([a-zA-Z0-9_\\-]*[a-zA-Z0-9])*(\\.[a-zA-Z]+([a-zA-Z0-1_\\-]*[a-zA-Z0-9])*)+$", *productID.Ptr())
```
illume, line 134: FQDN should be 255 make, with each part being 63 max. 
Hostnames can start with a letter or a number.
According to Omaha v3, this is also a Mac bundle ID, which has to be unique amongst other Mac bundle IDs.

backend/pkg/api/db/migrations/0014_add_application_product_id.sql
```diff
@@ -0,0 +1,9 @@
...
+alter table application add constraint application_unique_product_id unique(product_id);
```
illume, line 4: This also has to have a unique constraint against id.

backend/pkg/api/applications.go
```diff
@@ -222,13 +303,85 @@ func (api *API) GetApps(teamID string, page, perPage uint64) ([]*Application, er
...
+			} else {
+				logger.Error().Err(err).Msg("Failed to get apps")
+			}
+
+			cachedAppsRef = cachedAppIDs
+		}
+		cachedAppsIDsLock.Unlock()
+	}
+
+	// Trim space and the {} that may surround the ID
```
illume, line 365: I'm not sure where handling the `{}` requirement comes from? I can't see anything about this in the Omaha spec.

frontend/src/js/components/Applications/EditDialog.tsx
```diff
@@ -127,6 +137,16 @@ function EditDialog(props: { create?: any; data: any; show: boolean; onHide: ()
...
       .max(50, t('applications|Must be less than 50 characters'))
       .required('Required'),
+    product_id: Yup.string()
+      // This regex matches an ID that matches
+      // * At least two segments.
+      // * All characters must be alphanumeric, an underscore, or a dash [a-zA-Z0-9_].
+      // Each segment must start with a letter.
+      // Each segment must not end with an underscore or dash.
+      .matches(
+        /^[a-zA-Z]+([a-zA-Z0-9_\-]*[a-zA-Z0-9])*(\.[a-zA-Z]+([a-zA-Z0-9_\-]*[a-zA-Z0-9])*)+$/,
```
illume, line 147: Made some comments about this in the golang validation function. I think this validation also needs to be changed to match the spec better.

frontend/src/js/components/Applications/EditDialog.tsx
```diff
@@ -74,6 +75,15 @@ function EditDialog(props: { create?: any; data: any; show: boolean; onHide: ()
             fullWidth
             required
           />
+          <Field
+            name="product_id"
+            component={TextField}
+            margin="dense"
+            label={t('frequent|Product ID')}
```
illume, line 82: Where did the naming for product id come from? In the Omaha v3 spec it's called appid.

backend/pkg/api/applications.go
joaquimrocha: I deliberately left the product-id as a simple entry point, rather than a full alias within the logic, to avoid complexity and eventual introduction of bugs. So it should only be used for interaction with the user.

backend/pkg/api/applications.go
joaquimrocha: But we don't care about FQDN here, do we?
So the only thing wrong is that we're allowing uppercase?

backend/pkg/api/db/migrations/0014_add_application_product_id.sql
joaquimrocha: Doesn't it do that already because it's the unique key?

backend/pkg/api/applications.go
joaquimrocha: Seems like it may be a thing in Windows machines. See an example in: https://omaha-consulting.com/update-windows-service
We're also using this when identifying Nebraska's own sync requests, and filtering them from the instance list.

frontend/src/js/components/Applications/EditDialog.tsx
joaquimrocha: We have already an ID which belongs to an Application, so in order to distinguish from the domain name one, I came up with that one. Let me know if there are better suggestions than appid 🙂

backend/pkg/api/db/migrations/0014_add_application_product_id.sql
illume: Sorry, what I mean is that one could enter a product_id that matches an existing(or future) id.

backend/pkg/handler/instances.go
```diff
@@ -0,0 +1,60 @@
...
+			return ctx.NoContent(http.StatusNotFound)
+		}
+		logger.Error().Err(err).Str("appID", appID).Str("instanceID", instanceID).Msg("getInstance - getting instance")
+		return ctx.NoContent(http.StatusInternalServerError)
+	}
+	return ctx.JSON(http.StatusOK, instance)
+}
+
+func (h *Handler) GetInstanceStatusHistory(ctx echo.Context, appID string, groupID string, instanceID string, params codegen.GetInstanceStatusHistoryParams) error {
+	var defaultLimit int
```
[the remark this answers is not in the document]
yolossn: The value was not initialized here because the underlying `h.db.GetInstanceStatusHistory` defaults to `20` if the value is `0`. But makes sense to set the value to `20` here itself 👍🏽

backend/pkg/middleware/auth.go
yolossn: Yes, the endpoints which are not defined in the spec have `/*` as the path, and the frontend path values are not defined in the spec, to serve the frontend files without authentication `/*` is supposed to be in there.

backend/pkg/api/applications.go
illume: Yeah, you're right... it's not a FQDN. It is a reverse dns format which inherits this stuff (like underscores in hostnames but not domain names, numbers at the start of host names). But maybe we should keep it as strict as what apple allows.

I propose these changes:
- underscores not allowed (as per apple)
- capital letters not allowed (as per omaha v3 spec)
- 155 max characters (as per apple)


https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleidentifier
> A bundle ID uniquely identifies a single app throughout the system. The bundle ID string must contain only alphanumeric characters (A–Z, a–z, and 0–9), hyphens (-), and periods (.). Typically, you use a reverse-DNS format for bundle ID strings. Bundle IDs are case-sensitive.

(and just to be sure the docs are still valid... it's specified here too:)
https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW5
> The bundle identifier string identifies your application to the system. This string must be a uniform type identifier (UTI) that contains only alphanumeric (A-Z,a-z,0-9), hyphen (-), and period (.) characters. The string should also be in reverse-DNS format. For example, if your company’s domain is Ajax.com and you create an application named Hello, you could assign the string com.Ajax.Hello as your application’s bundle identifier.The bundle identifier is used in validating the application signature.

There's no explicit documentation of the min/max length. However, the regex and the max length used by their form is visible as 155 characters long.
https://stackoverflow.com/questions/49001779/what-is-maximum-allowed-length-of-ios-bundleid

backend/pkg/api/applications.go
illume: Thanks.

frontend/src/js/components/Applications/EditDialog.tsx
illume: I'm wondering about the direction of using one application.id in the table instead to hold both formats. 

In the database it would be a var char 155, with constraints on the format (either uuid4 or bundle ID). Reading https://github.com/google/omaha/blob/master/doc/ServerProtocolV3.md#guids I don't see it talking about having two different identifiers. 

cons
- var char is not as performant as using a uuid4 field
- when an ID is updated, it needs to be changed in all of the tables that use foreign keys as well
- bit more complicated migration script required

pros 
- it just has one ID, not two
- app code is simplified, no locking or caching
- db queries are simplified, just use the id

backend/pkg/middleware/auth.go
illume: Thanks.

backend/pkg/middleware/auth.go
```diff
@@ -0,0 +1,51 @@
...
+			for _, path := range paths {
+				if c.Path() == path {
+					return true
+				}
+			}
+		case "github":
+			paths := []string{"/v1/update", "/login/cb", "/login/webhook"}
```
illume, line 21: @yolossn why is this list different to the oidc one?

backend/pkg/handler/instances.go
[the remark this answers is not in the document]
illume: Thanks. I'll make the change here so that it's 20.

backend/pkg/api/groups.go
illume: @yolossn if it is ok with you, I will try and make this change?

backend/pkg/auth/oidc.go
```diff
@@ -132,60 +137,56 @@ func NewOIDCAuthenticator(config *OIDCAuthConfig) Authenticator {
...
 		httpError(c, http.StatusUnauthorized)
-		return
+		return nil
 	}
 
 	// If refresh token is not available in the session
 	// mark the request as unauthorized so that the session
 	// can be recreated with refresh_token
-	session := ginsessions.GetSession(c)
+	session := echosessions.GetSession(c)
```
illume, line 161: Here is an example of where `GetSession` is not checked for errors. `ginsessions.GetSession(c)` would always return something. But that is not the case for `echosessions.GetSession(c)`, as it can return nil in an error condition.

@yolossn  Should I update all users of `echosessions.GetSession` to check for errors?

backend/pkg/middleware/path.go
illume: Ah... removes trailing slash / from paths.

backend/pkg/api/groups.go
yolossn: cool 👍🏽

backend/cmd/nebraska/main.go
yolossn: Good catch missed this when adding the `github` support will fix.

backend/pkg/handler/app.go
```diff
@@ -0,0 +1,176 @@
...
+		return ctx.NoContent(http.StatusInternalServerError)
+	}
+
+	app, err = h.db.GetApp(appID)
+	if err != nil {
+		logger.Error().Err(err).Str("appID", appID).Msg("updateApp - getting updated app")
+		return ctx.NoContent(http.StatusInternalServerError)
+	}
+
+	// TODO: Confirm if old and new values should be logged
```
illume, line 120: TODO? Who should this be confirmed with? Seems ok that we are logging the old and the new.

@yolossn can I remove this TODO comment?

backend/cmd/nebraska/main.go
illume: Thanks. Do you want to add it as a new commit to this branch? They can be squashed back into previous commits later... but I think it makes collaborating easier if we do commits here instead of rebasing.

backend/pkg/middleware/auth.go
yolossn: This is because the `github` auth mode was designed in a way that backend will initiate the login if the session was not set, but in `oidc` we decided that the client should initiate the login based on `/config` repsonse.

backend/pkg/middleware/path.go
yolossn: Yes, removes the trailing slash. This was added for backward compatibility.

backend/pkg/handler/app.go
yolossn: Yes, the comment can be removed.

backend/cmd/nebraska/main.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+const serviceName = "nebraska"
+
+var (
+	logger            = util.NewLogger("nebraska")
+	middlewareSkipper = func(c echo.Context) bool {
+		requestPath := c.Path()
+
+		paths := []string{"/health", "/metrics", "/config", "/v1/update", "/*"}
```
illume, line 44: I see we are also skipping paths for authentication here. I wonder if this was intended? NewAuthSkipper also skips paths, but has a different list of paths to skip.

backend/pkg/api/db/migrations/0014_add_application_product_id.sql
illume: Scratch that, I'm mistaken. This won't be the case because both IDs have different formats, so it's not possible for them to clash.

backend/pkg/omaha/omaha_test.go
```diff
@@ -246,6 +246,32 @@ func TestFlatcarGroupNamesConversionToIds(t *testing.T) {
...
+	defer a.Close()
+	h := NewHandler(a)
+
+	tTeam, _ := a.AddTeam(&api.Team{Name: "test_team"})
+	tApp, _ := a.AddApp(&api.Application{Name: "test_app", Description: "Test app", TeamID: tTeam.ID, ProductID: null.StringFrom("io.kinvolk.MyApp")})
+	tPkg, _ := a.AddPackage(&api.Package{Type: api.PkgTypeFlatcar, URL: "http://sample.url/pkg", Version: "640.0.0", ApplicationID: tApp.ID, Arch: api.ArchAMD64})
+	tChannel, _ := a.AddChannel(&api.Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID), Arch: api.ArchAMD64})
+	tGroup, _ := a.AddGroup(&api.Group{Name: "test_group", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: false, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 4, PolicyUpdateTimeout: "60 minutes"})
+
+	validAppProductID := "io.kinvolk.MyApp"
```
illume, line 260: This needs updating from "io.kinvolk.MyApp" to "io.kinvolk.myapp" if we use the lowercase requirement from the omaha v3 spec.

backend/pkg/auth/oidc.go
illume: I think the GetSession interface should also return the error rather than returning nil. 

```golang
func GetSession(c echo.Context) (*sessions.Session, err) {
	session, ok := c.Get("session").(*sessions.Session)
	if !ok {
		return nil, "Failed getting session."
	}
	return session, nil
}
```

yolossn: The session will never be nil because it is being set in the sessions middleware. In `ginsessions` also it can return nil but it is left unhandled. Ideally we should move to a more robust package like  https://github.com/gorilla/sessions but didn't want to combine many major changes.

Refer:
https://github.com/kinvolk/nebraska/blob/7ecb9fcc6c1ff7f8358da5c04308b67b8eb9587f/backend/pkg/sessions/gin/gin.go#L24
https://github.com/kinvolk/nebraska/blob/7ecb9fcc6c1ff7f8358da5c04308b67b8eb9587f/backend/pkg/sessions/context.go#L16

backend/cmd/nebraska/main.go
yolossn: this is a generic skipper for middlewares like `Gzip`, `OapiRequestValidatorWithOptions ` etc.

backend/pkg/auth/oidc.go
illume: Thanks for the explanation.

backend/pkg/middleware/path.go
illume: If it's ok, I'll remove this and use the RemoveTrailingSlash middleware that comes with echo? 
https://github.com/labstack/echo/blob/8da8e161380fd926d4341721f0328f1e94d6d0a2/middleware/slash.go#L79

yolossn: Cool, makes sense

backend/cmd/nebraska/main.go
yolossn: shall we add fixup commits for now and then squash it later ?

illume: Ok, sounds like a plan!

backend/pkg/api/applications.go
joaquimrocha: I remember why I added the uppercase and not starting with a number. I checked Apple and Android at the time. Here is the case for: https://developer.android.com/studio/build/configure-app-module

I think we shouldn't stop supporting underscores just because Apple does not.
The omaha protocol does say lowercase, though I am not sure whether this is being enforced anywhere. Version 3.1 says the IDs are case-insensitive, so if we start supporting that, maybe we have to lowercase the incoming product IDs (something for the future): https://chromium.googlesource.com/chromium/src.git/+/refs/heads/main/docs/updater/protocol_3_1.md#Application-IDs

I am fine to keep the underscore but making everything lowercase, as well as ignoring the case as 3.1 does.
Keep in mind that the frontend is also using the regex for validation.

frontend/src/js/components/Applications/EditDialog.tsx
joaquimrocha: I considered this when doing the changes, but as you say, this has DB implications. Another issue is that I wanted to support both for legacy reasons. E.g. an old flatcar instance that communicates using the UUID, and a new one that uses the domain one.
Another reason is that, if later we want to drop support for the UUID (e.g. when moving to protocol 3.1), then we can also migrate the database to iterative indexes, which are more AFAIU performant than UUIDs.
Last but not least, if we keep the product ID editable in the UI (even if not a great idea for existing updates), then having it as not being the key does make things much simpler in the DB changes.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1860 @@
...
+          schema:
+            type: string
+            format: date-time
+        - in: query
+          name: page
+          required: false
+          schema:
+            type: integer
+        - in: query
+          name: perpage
```
illume, line 1064: This (and page) must have a minimum of 0. 

Probably should have a maximum as well?

backend/pkg/api/activity.go
```diff
@@ -52,6 +52,66 @@ type ActivityQueryParams struct {
 	PerPage    uint64    `json:"perpage"`
 }
 
+func (api *API) GetActivityCount(teamID string, p ActivityQueryParams) (int, error) {
```
illume, line 55: Seems this ignores the `PerPage` and Page attributes. @yolossn should it?

frontend/src/js/components/Applications/EditDialog.tsx
illume: > E.g. an old flatcar instance that communicates using the UUID, and a new one that uses the domain one.

Interesting. Does this mean that we should not allow updating the ID? (Well, maybe the product_id... but only once?). Otherwise will old flatcar instances not be broken?

backend/pkg/api/applications.go
illume: > I think we shouldn't stop supporting underscores just because Apple does not.
I just mention apple bundle ID because Omaha v3 specifies the GUID as this (and lowercase). 

The main issue I see with underscores is that they don't work as URLs (without encoding?).

Maybe we should consider being more relaxed to make the appid compatible with Android/Java/Apple/Omaha v3/v3.1? On the other hand if we make it more strict, then people wanting to use the appid in other systems won't have a problem (like eg. if they try to use an underscore in one that doesn't support it).

If we change to case insensitive, then we need to change the GetAppID.

This is where I'm leaning...
- case insensitive (to match Omaha 3.1)
- no underscores to match Omaha v3/3.1 (via the apple bundle ID spec)
- 155 to match the apple bundle ID max length (via the apple bundle ID spec)

Can you make the call?

backend/pkg/api/activity.go
yolossn: Yes, this function returns the `totalCount` so `PerPage` and `Page` attributes should be ignored.

backend/api/spec.yaml
yolossn: I think minimum should do.

Makefile
```diff
@@ -127,6 +132,17 @@ image:
 		-t "$(DOCKER_REPO)/$(DOCKER_IMAGE_NEBRASKA):latest" \
 		-f Dockerfile .
 
+.PHONY: backend/tools/codegen
+backend/tools/codegen:
+	go get github.com/deepmap/oapi-codegen/cmd/oapi-codegen
```
yolossn, line 137: this fetches the latest version everytime leading to CI failing.Locking it at v1.90

backend/api/spec.yaml
illume: Hrmm. I can't tell if this generates any checks for the minimum values for us. Does it handle the minimum checks?

Makefile
illume: Good idea!

backend/pkg/sessions/echo/echo.go
illume: This is already addressed, the return can't be nil here, because sessions middleware is always there.

backend/pkg/api/activity.go
illume: Thanks

backend/pkg/api/db/migrations/0014_add_application_product_id.sql
illume: (maybe this needs to change to 155, because bundle IDs can be that long)

backend/api/spec.yaml
yolossn: yes it does add the checks

yolossn: the swagger spec is used by the `OapiRequestValidatorWithOptions` to validate the request.

illume: Cool, thanks :)

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+      x-oapi-codegen-extra-tags:
+        json: application
+
+    instance:
+      type: object
+      required:
+        - id
+        - ip
+        - createdTs
+        - application
```
illume, line 1690: Should application, and alias be required? Since they are marked omit empty in golang, and can be empty in the db.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+      responses:
+        "200":
+          description: login token response      
+
+  /login/cb: 
+    get:
+      operationId: loginCb
+      description: login callback
+      responses:
+        "200":
```
illume, line 38: This can return status codes StatusUnauthorized, and StatusInternalServerError

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              $ref: "#/components/schemas/loginInfo"
+      responses:
+        "200":
+          description: login token response      
+
+  /login/cb: 
+    get:
```
illume, line 34: It looks like this also needs various different params specified depending on which auth backend is used. For example oidc uses query params like state, and code.

backend/pkg/auth/oidc.go
```diff
@@ -132,60 +137,56 @@ func NewOIDCAuthenticator(config *OIDCAuthConfig) Authenticator {
...
-	oidcRouter.GET("/cb", oa.loginCb)
-	oidcRouter.GET("/", oa.login)
-	oidcRouter.GET("/validate_token", oa.validateToken)
+func (oa *oidcAuth) SetupRouter(router *echo.Echo) {
+	router.Use(echosessions.SessionsMiddleware(oa.sessionStore, "oidc"))
 }
 
-func (oa *oidcAuth) validateToken(c *gin.Context) {
-	ctx := c.Request.Context()
+func (oa *oidcAuth) ValidateToken(c echo.Context) error {
```
illume, line 144: Should this use the codegen params?

backend/pkg/auth/oidc.go
```diff
@@ -283,16 +283,63 @@ func (oa *oidcAuth) login(c *gin.Context) {
 
 	// Redirect to generated AuthURL
 	redirectTo(c, authURL.String())
+	return nil
+}
+
+func (oa *oidcAuth) LoginToken(c echo.Context) error {
```
illume, line 289: Should this use the codegen params?

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        description: payload for login token
+        required: true
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              $ref: "#/components/schemas/loginInfo"
+      responses:
+        "200":
+          description: login token response      
+
```
illume, line 32: This can also return StatusNotImplemented and StatusInternalServerError.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+    name: Kinvolk
+    email: hello@kinvolk.io
+    url: http://kinvolk.io
+paths:
+  /login:
+    get:
+      operationId: login
+      description: login browser flow
+      responses:
+        "302":
```
illume, line 16: This can also return StatusBadRequest, and StatusInternalServerError.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+  title: Nebraska
+  description: Nebraska is an update manager.
+  contact:
+    name: Kinvolk
+    email: hello@kinvolk.io
+    url: http://kinvolk.io
+paths:
+  /login:
+    get:
+      operationId: login
```
illume, line 13: Takes a login_redirect_url query paramater.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+  /login/validate_token:
+    get:
+      operationId: validateToken
+      description: validate token
+      responses:
+        "200":
+          description: validate token response
+  /login/webhook:
+    post:
+      operationId: loginWebhook
```
illume, line 49: Takes a `X-Hub-Signature` and `X-Github-Event` headers. Does a lot through the golang.org/x/oauth2/github library.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+      security:
+        - githubCookieAuth: [] 
+      responses:
+        "200":
+          description: get config response
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/config"
+  /v1/update:
```
illume, line 67: This has a GET `/v1/update/mysecret` as well.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+      description: omaha endpoint
+      requestBody:
+        description: payload for omaha endpoint
+        required: true
+        content:
+          text/xml:
+            schema:
+              $ref: "#/components/schemas/omahaRequest"
+      responses:
+        "200":
```
illume, line 79: Can also return StatusBadRequest if the size of request is too large.

backend/pkg/handler/omaha.go
```diff
@@ -0,0 +1,33 @@
...
+		logger.Error().Err(err).Msg("process omaha request")
+		if uerr := errors.Unwrap(err); uerr != nil && uerr.Error() == "http: request body too large" {
+			return ctx.NoContent(http.StatusBadRequest)
+		}
+	}
+	return ctx.XMLBlob(http.StatusOK, responseBuffer.Bytes())
+}
+
+func getRequestIP(r *http.Request) string {
+	ips := strings.Split(r.Header.Get("X-FORWARDED-FOR"), ",")
```
illume, line 27: Is nebraska always behind a proxy that rewrites and adds this header? If not, then perhaps X-FORWARDED-FOR can be set by anyone.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+      description: login web hook
+      responses:
+        "200":
+          description: login webhook response
+  /config:
+    get:
+      operationId: getConfig
+      description: get app config
+      security:
+        - githubCookieAuth: []
```
illume, line 59: I wonder why only githubCookieAuth is here and not the oidc ones?

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        totalCount:
+          type: integer
+        count:
+          type: integer
+        activities:
+          type: array
+          items:
+            $ref: "#/components/schemas/activity"
+
+  securitySchemes:
```
illume, line 1877: Should `noop` be described here too?

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+
+  /login/cb: 
+    get:
+      operationId: loginCb
+      description: login callback
+      responses:
+        "200":
+          description: login cb success response
+  /login/validate_token:
+    get:
```
illume, line 41: Probably all the endpoints that don't require auth should mark that explicitly (with `security: []`).

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+          description: Omaha Response
+  /api/apps:
+    get:
+      description: get Apps
+      operationId: paginateApps
+      security:
+        - oidcBearerAuth: []
+        - oidcCookieAuth: []
+        - githubCookieAuth: []
+      parameters:
```
illume, line 89: It uses the team_id supplied by the auth as well.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+          schema:
+            type: integer
+            minimum: 10
+      responses:
+        "200":
+          description: Get App success response
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/appsPage"
```
illume, line 108: It's not clear to me if either count, or totalCount are used by the frontend.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        password:
+          type: string
+
+    appConfig:
+      type: object
+      required:
+        - name
+      properties:
+        name:
+          type: string
```
illume, line 1144: frontend has maxLength: 50

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        description:
```
illume, line 1145: frontend has maxLength: 250. The database doesn't specify a maximum length.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+          type: string
+        product_id:
```
illume, line 1147: There should probably be a "pattern:" regex here to match the product_id format.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        - oidcCookieAuth: []
+        - githubCookieAuth: []
+      parameters:
+        - in: path
+          name: appID
+          required: true
+          schema:
+            type: string
+      responses:
+        "204":
```
illume, line 215: In app.go@DeleteApp I see it returns StatusOK(200), not 204.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+          name: appID
+          required: true
+          schema:
+            type: string
+        - in: query
+          name: page
+          required: false
+          schema:
+            type: integer
+            minimum: 0
```
illume, line 239: This should have a `default: 1`.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        - in: query
+          name: perpage
+          required: false
+          schema:
+            type: integer
```
illume, line 244: This has a `default: 500`.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        description:
+          type: string
+        product_id:
+          type: string
+
+    groupConfig:
+      type: object
+      required:
+        - name
+        - description
```
illume, line 1154: Description is not required in the frontend. Neither are other fields (like track etc).

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        - policy_period_interval
+        - policy_timezone
+        - policy_office_hours
+        - policy_safe_mode
+        - policy_update_timeout
+        - policy_updates_enabled
+        - track
+      properties:
+        name:
+          type: string
```
illume, line 1165: `maxLength: 50`

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        description:
+          type: string
```
illume, line 1167: `maxLength: 250`

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        policy_timezone:
+          type: string
+        policy_period_interval:
+          type: string
+        policy_max_updates_per_period:
+          type: integer
+        policy_update_timeout:
+          type: string
+        track:
+          type: string
```
illume, line 1185: `maxLength: 256` in the frontend

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        policyTimezone:
+          type: string
+          x-oapi-codegen-extra-tags:
+            json: policy_timezone
+        policyPeriodInterval:
+          type: string
+          x-oapi-codegen-extra-tags:
+            json: policy_period_interval
+        policyMaxUpdatesPerPeriod:
+          type: integer
```
illume, line 1507: Frontend has this as minimum: 0

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+        - color
+        - createdTs
+        - applicationID
+        - packageID
+        - arch
+      properties:
+        id:
+          type: string
+        name:
+          type: string
```
illume, line 1421: frontend has maximum: 50 for this

backend/api/spec.yaml
```diff
@@ -0,0 +1,1889 @@
...
+      properties:
+        application_id:
+          type: string
+        arch:
+          type: integer
+          x-go-type: uint
+        color:
+          type: string
+        name:
+          type: string
```
illume, line 1263: max 50

backend/api/spec.yaml
yolossn: The request validation will fail here as differnet params are required for different auth_modes, so didn't list
all the params here.

backend/pkg/auth/oidc.go
yolossn: Cautiously avoided the codegen params here as the `Authenticator` interface will get messed up with OIDC specific params.

backend/pkg/auth/oidc.go
yolossn: Cautiously avoided the codegen params here as the `Authenticator` interface will get messed up with OIDC specific params.

backend/api/spec.yaml
yolossn: In OIDC auth mode the endpoint is not behind auth.

backend/api/spec.yaml
yolossn: This is not possible as the `secret` is dynamically set by the user.

backend/api/spec.yaml
yolossn: In OIDC auth mode the endpoint is not behind auth.

backend/api/spec.yaml
yolossn: Yes, but it is not something that is requested explicitly.

backend/api/spec.yaml
yolossn: Probably should be used, what if there are more than 10 applications (default value for PerPage param).

backend/api/spec.yaml
yolossn: The frontend doesn't send the `page` and `perpage` query params so it is marked `required: false` and Internally have handled the value to default to `1` if `page` is not passed.

backend/api/spec.yaml
yolossn: Where did you find this value from? 500 is a lot for `perpage`.

backend/api/spec.yaml
yolossn: From the request made by the frontend, it looks like only `track` is not required, other values have defaults being sent.

backend/api/spec.yaml
yolossn: Usually, something like this should be solved by having a reverse proxy (like nginx) that rewrites the request to `/v1/update/mysecret` to `/v1/update`.

backend/pkg/handler/omaha.go
```diff
@@ -0,0 +1,33 @@
...
+	"strings"
+
+	"github.com/labstack/echo/v4"
+)
+
+func (h *Handler) Omaha(ctx echo.Context) error {
+	responseBuffer := new(bytes.Buffer)
+	ctx.Response().Writer.Header().Set("Content-Type", "text/xml")
+	ctx.Request().Body = http.MaxBytesReader(ctx.Response().Writer, ctx.Request().Body, UpdateMaxRequestSize)
+	if err := h.omahaHandler.Handle(ctx.Request().Body, responseBuffer, getRequestIP(ctx.Request())); err != nil {
```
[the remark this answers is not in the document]
yolossn: AFAIU the response status code is not a part of `OmahaResponse` and the Omaha errors are handled by the internal response structures.

backend/pkg/handler/omaha.go
yolossn: I think this was done this way to handle the case when nebraska is run behind a proxy, I understand this can be set by someone intentionally. Is there an alternate way to handle both cases?

backend/pkg/auth/oidc.go
yolossn: Intentionally avoided the codegen params here as the `Authenticator` interface will get messed up with OIDC specific params.

backend/pkg/auth/oidc.go
yolossn: Intentionally avoided the codegen params here as the `Authenticator` interface will get messed up with OIDC specific params.

backend/api/spec.yaml
yolossn: I don't think openapi spec supports `noop` authentication scheme.
https://swagger.io/docs/specification/authentication/

backend/api/spec.yaml
yolossn: Any specific reason for this?

backend/api/spec.yaml
illume: Ok, thanks. I see now why the spec isn't accurate here, but I wonder if we can make it accurate?

- Can we encode the different paramater requirements in such a way that the requirements for each auth mode are listed in here, and will validate correctly depending on the auth mode?
- Could the URLs be different depending on the auth mode used?

backend/api/spec.yaml
illume: helpers.go `defaultPerPage uint64 = 500`

backend/pkg/middleware/auth.go
yolossn: Yes, certain endpoints dont need authentication based on the auth mode.

backend/api/spec.yaml
illume: From what you say, it sounds like this defaults to 1? Which means we should mark it in the spec as default 1... unless I'm not understanding something.

backend/api/spec.yaml
yolossn: Having different URLs for different auth mode can be an option but the catch is in `github` auth mode the backend initiates the login and incase on `oidc` the client (frontend) initiates the login so this can be quite bigger of a change than we expect. Also, we didn't change this earlier when `oidc` was introduced because then we were discussing of removing the support for `github` auth.

backend/api/spec.yaml
illume: Mainly so we know if authentication is required, or if it was accidentally left out. As a reader, without explicitly marking each I can't be sure if it's an accidental omission or if it's really supposed to not use auth.

I think it's possible to make auth/security required by default, and then `security: []` would be required.

backend/api/spec.yaml
yolossn: That is in the DB level and in the handler level it is handled to default to `10`. IMHO 500 is a lot for `perpage`.

backend/api/spec.yaml
illume: If we prefix all the oidc parts with "oidc" like `/oidc/login/cb` do you think that would be ok? Or do you think that will not work... or require a very large change?

backend/api/spec.yaml
yolossn: Nebraska is packaged in a way that even the frontend files are served by the backend and we cant make security required by default. Certain endpoints like `/login`,`/login/cb`, `/config` incase of `oidc` auth mode need to be accessible without auth, ill better add `security: []`  to the ones were auth is not required.

backend/api/spec.yaml
yolossn: The generated server code doesn't default the value to 1 if it is not provided in the request that is why it was handled this way manually.

backend/api/spec.yaml
yolossn: If we want to use `/oidc/login/cb` we can only document for `oidc` and that will only be a workaround for documenting only the `oidc` callback endpoint.

illume: Sorry, I meant to ask about moving all URLs that have OIDC specific things to /oidc/ prefixed endpoints? This would allow each interface to be specified, and for us to use the generated code for validation/etc.

The OIDC endpoints would change, rather than the github ones because those are the new ones.

backend/api/spec.yaml
illume: Probably because it's part of the interface it should be specified here.

I don't think the mysecret stuff is documented anywhere, but if it is I'd like to link to there (and if it's not it needs documentation). I guess it's a simple (but sort of flawed) way to authentication.

backend/pkg/handler/omaha.go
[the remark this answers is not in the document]
illume: I'll check this, and write in the docs about how the errors are passed through the response.

backend/pkg/handler/omaha.go
illume: I'll look into this more.

backend/api/spec.yaml
illume: Ah yes. I don't think there's any reason why it should be 500 in that validation function. I'll change that to 10 as well.

backend/api/spec.yaml
illume: Thanks. I'll look into this. Maybe we need to raise an issue about it not setting the default.

backend/api/spec.yaml
illume: Yeah, ok. I'll look into using them in the frontend.

backend/api/spec.yaml
yolossn: This change will be a breaking change and the users will have to reconfigure the callback URLs in `OIDC` provider and their `github` applications. The two downsides are that the list of query params is not documented as they vary depending on the auth mode and the validation is done manually instead of open-api spec based validation.

backend/api/spec.yaml
yolossn: the secret is defined dynamically by the user at runtime, are you asking to add a comment with a link to the `mysecret` authentication explanation?

backend/api/spec.yaml
illume: If we keep the github URLs the same, then I guess that's ok. Because we haven't released OIDC support to anyone yet, it means we could change the OIDC routes.

joaquimrocha: We do have OIDC published, @illume . I like your suggestion of having the auth endpoints prefixed with their domain, but we need to come up with a solution for the OIDC callback issue as I don't want to eventually break people's deployments.
Maybe we move all but the callback URL, and put out a deprecation note?

illume: > We do have OIDC published

Ah ok. My mistake.

> Maybe we move all but the callback URL, and put out a deprecation note?

Yeah, I think that's a good option. Sounds like it won't break anything, and we can fix all except the callback route.

I'll look into this one.

backend/pkg/handler/omaha.go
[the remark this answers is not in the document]
illume: Seems nebraska/backend/pkg/omaha/omaha.go@Handle can return two errors here we are not handling. The previous handler implementation also had this issue.

backend/pkg/handler/omaha.go
illume: It wasn't clear where this code came from because of how the commit(s) are done. It is an existing function from controller.go that also had the same issue. I think this should only use X-FORWARDED-FOR if an option is given to use that. Something like "I am behind proxy".

illume: Because the issue exists previous to this PR, I added a new issue: https://github.com/kinvolk/nebraska/issues/528

backend/pkg/handler/omaha.go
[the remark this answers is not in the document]
illume: Since this issue was existing before this PR I made an issue: https://github.com/kinvolk/nebraska/issues/529 so it can be addressed outside of this PR.

backend/api/spec.yaml
illume: I think having it as an optional path argument is fine. I'm not aware of a way to define a dynamic spec in our setup. I'll add something in the description about `api-endpoint-suffix` and mysecret until we have a doc to link to<sup>1</sup>.

- <sup>1</sup> Note: the config.go `api-endpoint-suffix` documentation isn't anywhere else AFAIK. I created an issue to document the config.go stuff here: https://github.com/kinvolk/nebraska/issues/530

backend/pkg/handler/groups.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+		logger.Error().Err(err).Str("groupID", groupID).Msg("getGroup - getting group")
+		return ctx.NoContent(http.StatusInternalServerError)
+	}
+
+	return ctx.JSON(http.StatusOK, group)
+}
+
+func (h *Handler) UpdateGroup(ctx echo.Context, appID string, groupID string) error {
+	logger := loggerWithUsername(logger, ctx)
```
illume, line 84: Needs to return 404 when the required appID does not exist?

backend/pkg/handler/groups.go
```diff
@@ -0,0 +1,288 @@
...
+		if err == sql.ErrNoRows {
+			return ctx.NoContent(http.StatusNotFound)
+		}
+		logger.Error().Err(err).Str("groupID", groupID).Msg("updateGroup - getting old group to update")
+		return ctx.NoContent(http.StatusInternalServerError)
+	}
+
+	group := groupFromRequest(request.Name, request.Description, request.PolicyMaxUpdatesPerPeriod, request.PolicyOfficeHours, request.PolicyPeriodInterval, request.PolicySafeMode, request.PolicyTimezone, request.PolicyUpdateTimeout, request.PolicyUpdatesEnabled, request.ChannelId, request.Track, groupID, appID)
+
+	err = h.db.UpdateGroup(group)
```
illume, line 104: db.UpdateGroup doesn't update the ApplicationID... should it? Needs docs on if it does/does not.

backend/api/spec.yaml
```diff
@@ -0,0 +1,1932 @@
...
+      security:
+        - oidcBearerAuth: []
+        - oidcCookieAuth: []
+        - githubCookieAuth: []
+      parameters:
+        - in: path
+          name: appID
+          required: true
+          schema:
+            type: string
```
illume, line 198: Even though uuid v4 format we are using is not defined by swagger, it is allowed to be specified to help with documentation <sup>1</sup>. Additionally, to help avoid issues the field should be validated with a minLength/maxLength.

```
type: string
format: uuidv4
minLength: 36
maxLength: 36
```

This should be done for every field that is a UUID.

- <sup>1</sup> https://swagger.io/specification/#format

illume: This could be done later IMHO, because this functionality wasn't there previous to this PR. I raised an issue for this here: https://github.com/kinvolk/nebraska/issues/531

backend/api/spec.yaml
illume: Moved this into an issue and out of this PR instead: https://github.com/kinvolk/nebraska/issues/535

backend/api/spec.yaml
illume: Moved this into an issue and out of this PR instead: https://github.com/kinvolk/nebraska/issues/535

backend/api/spec.yaml
illume: Moved this into an issue and out of this PR instead: https://github.com/kinvolk/nebraska/issues/535

backend/api/spec.yaml
illume: Resolving this because it was an issue previous to this branch, and there's an issue for it now.

backend/pkg/api/db/migrations/0014_add_application_product_id.sql
joaquimrocha: It's better to make it smaller now than later. So we can change to 155. cc/ @yolossn

backend/pkg/api/applications.go
joaquimrocha: > On the other hand if we make it more strict, then people wanting to use the appid in other systems won't have a problem (like eg. if they try to use an underscore in one that doesn't support it).

I don't think people will have a problem if we support underscores and they don't use them.
But this is a decision we can take later, as being more restrictive now is better than trying to do it later and breaking the use-case for folks using already the underscore or whatever non-omaha rules we apply.

So my call is to go with those 3 rules you're suggesting. Let's do that.

frontend/src/js/components/Applications/EditDialog.tsx
joaquimrocha: Older instances will not be broken because we support both product-id and UUID at the same time. When we add the product-id, it will be in addition to the UUID.
About changing an existing product-id, we should just have a warning in the UI that it may break clients who are using it.

joaquimrocha: Hi again. I was in a call with @yolossn and even though we'd gain something by explicitly validating the endpoint's parameters in the spec, it's also true that it's unnecessary work for frontend/clients to be explicit about the auth backend when the auth backend should be a setup detail.

So the solution to supporting both backend-prefixed and generic login routes would be to have the backend specific ones, and redirect the generic ones (without params in the spec) to them.

However, this brings the question of whether it's worth introducing new endpoints just for the automatic param validation, as it could lead to having clients use those routes directly, which means they'd have to change their configs if later we deprecate an endpoint/backend.

Thus, the price of keeping only the generic endpoints ATM at the expense of having a bit less well defined docs and manual validation is much less than introducing things we may have to remove later.

backend/pkg/api/applications.go
illume: Great. I'm working on this.

backend/pkg/omaha/omaha_test.go
illume: We're going with case insensitive, so it doesn't need changing.

backend/api/spec.yaml
illume: Moved this comment from another unrelated thread, where it was put mistakenly.

frontend/src/js/components/Applications/EditDialog.tsx
illume: This last comment is about another topic, so I pasted it into here: https://github.com/kinvolk/nebraska/pull/513#discussion_r776525150

frontend/src/js/components/Applications/EditDialog.tsx
illume: This was changed.

frontend/src/js/components/Applications/EditDialog.tsx
illume: Still remaining is to add the UI warning/confirmation about changing the product-id.

backend/pkg/api/applications.go
illume: Changed the validation functions to reject underscore and have a max length of 155. 

Still remaining to do is make the GET/comparison functions case insensitive (so lowercase is the same as LOWERCASE).

backend/api/spec.yaml
illume: This was changed to 10.

backend/pkg/api/applications.go
illume: There's a check now for case insensitive duplicate product-ids.
Also, applications.go@GetAppID is case insensitive.

illume:
I'm making a new squashed/rebased branch `proof-of-concept3`, and leaving this one around for history. 

There's still a number of open conversations on here that I'll move into issues.

frontend/src/js/components/Applications/EditDialog.tsx
illume: Opened an issue here for the product-id warning: https://github.com/kinvolk/nebraska/issues/547

backend/api/spec.yaml
illume: Added an issue for this here: https://github.com/kinvolk/nebraska/issues/548

backend/api/spec.yaml
illume: Raised an issue about this here: https://github.com/kinvolk/nebraska/issues/549

backend/pkg/auth/oidc.go
illume: Raised an issue about this here: https://github.com/kinvolk/nebraska/issues/549

backend/pkg/auth/oidc.go
illume: Raised an issue about this here: https://github.com/kinvolk/nebraska/issues/549

illume:
New PR is here (commits squashed and rebased against main) https://github.com/kinvolk/nebraska/pull/546
I added issues for all the open conversations.

Closing this one now.

PR #517, opened 2021-11-23
frontend: fix small typo to default applications message

goshlanguage:
# fix small typo to default applications message

This is a typo fix for the default page, when no Applications are present. This usually isn't shown, but I'm seeing it with an issue with my roles, so I thought I would submit I PR. Hope this is helpful.

Screen Shot 2021-11-24 at 11 26 58 AM

## How to use

Shows up when no applications exist, or the user has insufficient roles.

## Testing done

Not sure what the stub is to run this, happy to test if necessary. Not sure how to use the typescript tool chain to run this locally.


yolossn:
Thanks for the contribution.

yolossn (approved):

Issue #518, opened 2021-11-23
chart: nebraskaUrl only populated when config.hostFlatcarPackages.enabled


goshlanguage:
Hey team, 

Thanks for your hard work!

It seems that the `-nebraska-url` flag is only populated if the `.Values.config.hostFlatcarPackages.enabled` chart value is toggled on. This can be seen in the [deployment.yaml](https://github.com/kinvolk/nebraska/blob/main/charts/nebraska/templates/deployment.yaml#L70-L79).

In the event that a user is _not_ running this locally (since the default is `http://localhost:8000/login/cb`), and `config.hostFlatcarPackages.enabled` is _not_ set, the `-nebraska-url` flag is not appended to the args list, meaning that any links and redirects in the app become broken if the ingress is set, because the application will use it's default, `http://localhost:8000`. In my case, I enabled SSO, which is broken by this logic since it tries to auth through `localhost`, and not my given (functional) ingress.

I believe this to be erroneous. While this setting is embedded in this portion of the values file, without this, the interface becomes broken for things such as authentication when `config.hostFlatcarPackages.enabled` is not set.

PR #519, opened 2021-11-23
chart: decouple nebraska-url flag from host packages toggle

goshlanguage:
# decouple nebraska-url flag from host packages toggle

The need for this PR is outlined in #518.

At current, if I enable the ingress without enabling the host package, all links and redirects default to `localhost`, which breaks nebraska.

## How to use

The ingress by default is passed to the application through the `-nebraska-url` flag. If `.Values.config.hostFlatcarPackages.nebraskaURL` is set, it takes precedence.

## Testing done

I have templated locally with 

`helm template  --output-dir test-output/ . --debug`

outcome: `deployment.yaml` appends `- "-nebraska-url=http://flatcar.example.com"` to args

`helm template  --output-dir test-output/ . --set ingress.enabled=true --set ingress.hosts\[0\]="https://kinvolk.rocks"`

outcome: `deployment.yaml` appends `- "-nebraska-url=http://flatcar.example.com"`

I have installed the chart in this PR to a local cluster as well.

I am partially concerned with the default now being that `-nebraska-url` is always set, because the default `ingress.enabled` is true. For local installations, this should not be the case, but I believe that the default ingress should be disabled if that's the desired outcome.

### DIFF
--- a/charts/nebraska/templates/deployment.yaml
+++ b/charts/nebraska/templates/deployment.yaml
@@ -71,11 +71,13 @@ spec:
             {{- if .Values.config.hostFlatcarPackages.enabled }}
             - "-host-flatcar-packages"
             - "-flatcar-packages-path={{ required "A valid 'packagesPath' is required when hosting Flatcar packages." .Values.config.hostFlatcarPackages.packagesPath }}"
-              {{- if .Values.config.hostFlatcarPackages.nebraskaURL }}
+            {{- end }}
+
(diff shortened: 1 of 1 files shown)


yolossn:
Thanks for the PR, Please bump the chart version for the CI to pass.

yolossn:
Hey, looks like there is a duplicate commit for bumping the chart version, otherwise LGTM.

Issue #520, opened 2021-11-23
chart: the extraArgs value is no longer passed to binary


goshlanguage:
## Description

Hello team,

[The values file documentation](https://github.com/kinvolk/nebraska/tree/main/charts/nebraska) states:

| Parameter                               | Description                                                                                                                              | Default                               |
|-|-|-|
| `extraArgs`                                           | Extra arguments to pass to Nebraska binary                                                                                           | `[]`                                                                    |

However, this has been removed entirely from the [`deployment.yaml`](https://github.com/kinvolk/nebraska/blob/main/charts/nebraska/templates/deployment.yaml#L44-L118) manifest it seems, so this no longer works.

## Impact

I'm not able to work around another bug with regards to the `-nebraska-url` flag by passing it through `extraArgs`

## Environment and steps to reproduce
 
1. **Set-up**: deploy nebraska helm chart, setting extraArgs to _anything_
2. **Task**: check the templated output, ie: `helm template  --output-dir test-output --set extraArgs[\0\]="anything"` 
3. **Action(s)**: `grep "anything" test-output/nebraska/templates/deployment.yaml`
4. **Error**: `anything` isn't populated in the output, because the stub is removed.

## Expected behavior

`extraArgs` should be appended to the args portion of the deployment spec.

PR #521, opened 2021-11-23
chart: add extraArgs functionality

goshlanguage:
# add extraArgs functionality

Outlined in #520, while the `README.md` outlines that an `extraArgs` list is available to pass additional flags to the `nebraska` binary, it doesn't actually exist in the `deployment.yaml` manifest. This PR resolves that.

## How to use

Populate `extraArgs`, and get the output templated in the `deployment.yaml` manifest as intended.

## Testing done

Borrowing from the example listed in [`values.yaml`](https://github.com/kinvolk/nebraska/blob/main/charts/nebraska/values.yaml#L79-L81), I tested the default chart first, which results in templates being rendered without `extraArgs` since the default is `[]`, then using the first example with:

```
helm template  --output-dir test-output/ . --set extraArgs\[0\]="-http-log"
```

resulting in the following deployment snippet being rendered:

```
          args:
            - "-client-title=Kinvolk Update Service"
            - "-client-logo=/nebraska/assets/kinvolk-logo.svg"
            - "-client-header-style=dark"
            - "-http-static-dir=/nebraska/static"
            - "-enable-syncer"
            - "-auth-mode=noop"
            - "-http-log"
```

This is also tested by setting the `values.yaml` to the first example, and getting the same output as above.

`values.yaml`:

```
extraArgs:
  - "-http-log"
```

rendered `deployment.yaml`:

```
          args:
            - "-client-title=Kinvolk Update Service"
            - "-client-logo=/nebraska/assets/kinvolk-logo.svg"
            - "-client-header-style=dark"
            - "-http-static-dir=/nebraska/static"
            - "-enable-syncer"
            - "-auth-mode=noop"
            - "-http-log"
```

### DIFF
--- a/charts/nebraska/templates/deployment.yaml
+++ b/charts/nebraska/templates/deployment.yaml
@@ -116,6 +116,13 @@ spec:
             - "-oidc-scopes={{ . }}"
               {{- end }}
             {{- end }}
+
+            {{- /* --- Extra Args --- */ -}}
+            {{- if .Values.extraArgs }}
+            {{- range .Values.extraArgs }}
+            - "{{ . }}"
+            {{- end }}
+            {{- end }}
           env:
             - name: DB_PASSWORD
               valueFrom:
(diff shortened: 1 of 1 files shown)


yolossn:
Please bump the chart version in `Chart.yaml`
Similar to https://github.com/kinvolk/nebraska/pull/522#issuecomment-979082559

yolossn (approved):
LGTM, Please bump the chart version to fix the CI.
https://github.com/kinvolk/nebraska/pull/521#issuecomment-979088649

PR #523, opened 2021-11-30
charts: Don't verify the version has been bumped when linting in CI

joaquimrocha:
So we don't have to bump the chart version every time we do changes, just for the lint to successfully run.

### DIFF
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -1,3 +1,4 @@
 chart-repos:
   - bitnami=https://charts.bitnami.com/bitnami
 target-branch: main
+check-version-increment: false


.github/workflows/helm-chart-lint-test.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
           fi
 
       - name: Run chart-testing (lint)
-        run: ct lint --config .github/ct.yaml
+        run: ct lint --check-version-increment=false --config .github/ct.yaml
```
yolossn, line 39: since we use the `--config` flag already this config has to be added to `.github/ct.yaml`

Issue #533, opened 2021-12-13
oidc auth login redirect bug breaks auth flow


goshlanguage:
## Redirects in the app break when oidc is enabled

When using github auth, it seems that there is no check of the auth redirect. However, we recently switched from github to oidc, and when on any page except for the page I've set in the `--validRedirectURLs` flag, the redirect is broken, likely because [this check in `oidcauth.go`](https://github.com/kinvolk/nebraska/blob/1b9e39c8324717bc7a444f32ac6ed9c1bc49f8ee/backend/cmd/nebraska/auth/oidcauth.go#L247-L252) is only passable on said page.

Practically, what this means is, when I go to the `nebraska-url`, I get redirected to the login. If I then click on one of my groups, then wait long enough or refresh, this auth is checked again, but fails, because the redirect url is then set to `login/?login_redirect_url=https://nebraska.mydomain.com/apps/myuuid/groups/groupuuid`

The glob check fails because the `login_redirect_url` query parameter != `validRedirectUrls`.

You can see a working example that emulates the current codebase [here](https://go.dev/play/p/_PZMalCIZDJ)

## Impact

oidc auth is broken, and breaks using the app on anything except for the home page

## Environment and steps to reproduce

1. **Set-up**: deploy Nebraska with oidc. 
2. **Task**: Go to any page not set in --validRedirectURLs, eg: any groups 
3. **Action(s)**: refresh that page
4. **Error**: Invalid login_redirect_url

I am using the helm chart specifically, and my oidc config looks something like this:
```
      oidc:
        clientID: clientID
        clientSecret: clientSecret
        issuerURL: https://my.provider.com/
        validRedirectURLs: https://nebraska.mydomain.com
        managementURL:
        logoutURL: https://my.provider.com/v2/logout
        adminRoles: admin
        viewerRoles: viewer
        rolesPath: http://kinvolk\\.io/roles
        scopes:
        sessionAuthKey:
        sessionCryptKey:
```

## Expected behavior

working auth flow and a lack of interruption/errors

yolossn:
Hey, @goshlanguage The`validRedirectURLs` value was designed to accept strings with `wildcard (*)` in the provided values. I can see that there is a gap in the documentation will update it.

goshlanguage:
I would ask, what is the benefit to not appending a glob at the end of the given redirectUrl? 

I don't see how it makes any sense to only allow auth redirection on one route, if all of the defined routes will redirect to themselves through how the auth middleware works.

[...]

PR #534, opened 2021-12-13
backend: add wildcard to glob validations for validRedirectURL

goshlanguage:
# Add wildcard to glob validations for validRedirectURL

Please see #533 for a detailed description. The way that the loginRedirectURL query parameter is validated against entries in `validRedirectURLs` doesn't account for routes in `validRedirectURLs`. While an argument could be made that the user should append '*' to this flag, this isn't called out in any of the supporting documentation at current, nor am I sure it makes sense given how the check is setup.

## How to use

You would need to deploy nebraska with a valid oidc configuration and follow the repro in #533 

## Testing done

Results can be compared from [the changes in this pr](https://go.dev/play/p/VcTZJ4e3lzT) vs [the code as it stands](https://go.dev/play/p/_PZMalCIZDJ)


joaquimrocha:
@yolossn or @ashu8912 , can you take a look?

PR #541, opened 2021-12-20
backend: Update oidc-valid-redirect-urls help string

yolossn:
This patch improves the help string of
oidc-valid-redirect-urls by adding explanation
for wildcard support.


joaquimrocha (changes requested):
I think this is good to be clarified, yet the string can be improved as it's more suggestive than it should, IMO. How about:
`OIDC valid Redirect URLs; accepts comma separated values and supports wildcards (*), for example http://nebraska.example.io/*`

backend/cmd/nebraska/nebraska.go
```diff
@@ -62,15 +62,15 @@ var (
 	oidcClientID          = flag.String("oidc-client-id", "", "OIDC client ID used for authentication")
 	oidcClientSecret      = flag.String("oidc-client-secret", "", fmt.Sprintf("OIDC client Secret used for authentication; can be taken from %s env var too", oidcClientIDEnvName))
 	oidcIssuerURL         = flag.String("oidc-issuer-url", "", fmt.Sprintf("OIDC issuer URL used for authentication;can be taken from %s env var too", oidcClientSecretEnvName))
-	oidcValidRedirectURLs = flag.String("oidc-valid-redirect-urls", "http://localhost:8000/*", "OIDC valid Redirect URLs accepts comma separated values with wildcard *, for example if nebraska is hosted at http://nebraska.kinvolk.io the value should be http://nebraska.kinvolk.io/*")
+	oidcValidRedirectURLs = flag.String("oidc-valid-redirect-urls", fmt.Sprintf("%s/*", *nebraskaURL), "OIDC valid Redirect URLs; accepts comma separated values and supports wildcards (*), for example http://nebraska.example.io/*. If not set defaults to <nebraska-url>/*")
```
joaquimrocha, line 65: I think it's easier if you set the oidcValidRedirectURLs as `""` because this way you know already that it's not been set.
The docs make it explicit what it does if the user doesn't override it.

backend/cmd/nebraska/nebraska.go
```diff
@@ -165,6 +165,16 @@ func mainWithError() error {
 
 		url.Path = "/login/cb"
 
+		if (*nebraskaURL != "http://localhost:8000") && (*oidcValidRedirectURLs == "http://localhost:8000/*") {
```
joaquimrocha, line 168: With the `oidcValidRedirectURLs` as an empty string by default, this check becomes much simpler: `if *oidcValidRedirectURLs == ""`

joaquimrocha (changes requested):
Left a small comment which I think will improve the code.

joaquimrocha (approved):

PR #551, opened 2022-01-04
Add API integration tests

yolossn:
TODO:

- [x] Test Omaha Endpoint.
- [x] Find a way to test the timeline endpoints properly (Refer TODO statements in `stats_test.go`).
- [x] Update CI/corresponding make functions to use the `docker-compose.test.yaml` file.

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -30,19 +30,18 @@ coverage.out: check-code-coverage
 check-code-coverage:
 	go test -p 1 -coverprofile=coverage.out ./...
 
-container_id:
-	./tools/setup_local_db.sh \
-		--id-file container_id.tmp \
-		--db-name nebraska_tests \
-		--password nebraska \
-		--pg-version 13.3
-	mv container_id.tmp container_id
+.PHONY: test-service-up
+test-service-up:
+	docker compose -f ./docker-compose.test.yaml up -d
+
+.PHONY: test-service-down
+test-service-down:
+	docker compose -f ./docker-compose.test.yaml down
 
 .PHONY: check-backend-with-container
-check-backend-with-container: container_id
-	set -e; \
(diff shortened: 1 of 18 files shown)


backend/Makefile
```diff
@@ -30,19 +30,18 @@ coverage.out: check-code-coverage
...
 
-container_id:
-	./tools/setup_local_db.sh \
-		--id-file container_id.tmp \
-		--db-name nebraska_tests \
-		--password nebraska \
-		--pg-version 13.3
-	mv container_id.tmp container_id
+.PHONY: service-up
+service-up:
```
illume, line 34: Maybe `test-service-up` ? To make it more clear (when just looking at target names with `make tab`) that these are tests rather than the service.

backend/test/api/stats_test.go
```diff
@@ -0,0 +1,211 @@
...
+		httpDo(t, url, method, nil, http.StatusOK, "json", &timelineResponse)
+
+		// get group version timeline from DB
+		timelineDB, _, err := db.GetGroupVersionCountTimeline(appWithInstance.Groups[0].ID, "1d")
+		require.NoError(t, err)
+		require.NotNil(t, timelineDB)
+
+		assert.Equal(t, len(timelineDB), len(timelineResponse))
+
+		// Since the response from DB and API doesn't match as they are relative to the time at which
```
illume, line 45: 👍

illume (approved):
🎉 Looking good. Thanks for doing this.

I left a little comment about the make target name. Apart from that looks good to me. 

I'm not sure if we should be testing the auth related stuff here? But I understand testing auth is out of scope for this PR. Additionally, testing error handling is out of scope of this PR.

In case anyone else is reading, the reason this was done as a http level integration test rather than the usual way with testing handlers... is because the proof-of-concept branch rewrites the backend, so that isn't possible to test the handlers easily. This way the tests can be used on the main branch and then also in the proof of concept branch.

joaquimrocha:
Thanks @yolossn for such a massive effort!

PR #558, opened 2022-01-21
backend: Add e2e test for OIDC auth mode

yolossn:
This patch adds end to end test for OIDC auth mode.


yolossn:
I think the CI is failing because of golang version mismatch,Ill fix this. I have raised an issue for the same https://github.com/kinvolk/nebraska/pull/559.

joaquimrocha (approved):
Wow. This looks very comprehensive! Thanks for sticking to it, @yolossn !

illume (approved):
🎉 👍 wonderful, thanks :)

backend/test/auth/oidc/auth_test.go
```diff
@@ -0,0 +1,62 @@
...
+	OidcClientSecret:    clientSecret,
+	OidcIssuerURL:       issuerURL,
+	OidcAdminRoles:      "nebraska-admin",
+	OidcViewerRoles:     "nebraska-member",
+	OidcRolesPath:       "groups",
+	OidcScopes:          "openid,profile,email,groups",
+}
+
+func TestMain(m *testing.M) {
+	if os.Getenv("NEBRASKA_SKIP_TESTS") != "" {
```
illume, line 44: Is NEBRASKA_SKIP_TESTS used for the login_test.go as well?

illume:
When I run tests locally with: `cd backend;make check-backend-with-container` I get a bunch of errors, including:
```
--- FAIL: TestLogin (3.99s)
    --- FAIL: TestLogin/invalid_login_redirect_url (0.34s)
        login_test.go:122: <nil>
        login_test.go:134:
                Error Trace:    login_test.go:134
                Error:          Received unexpected error:
                                Get "http://localhost:6000/login?login_redirect_url=http://localhost:9000": dial tcp 127.0.0.1:6000: connect: connection refused
                Test:           TestLogin/invalid_login_redirect_url
```

illume (changes requested):
I was testing it locally and had a couple of issues mentioned in previous comments.

backend/test/auth/oidc/auth_test.go
yolossn: This pattern is used across tests. It makes it easy to skip all the tests by setting the `NEBRASKA_SKIP_TESTS ` value, so I followed the same.

yolossn:
I am not able to reproduce it, by any chance is port 6000 used by some other application in your local machine?

illume:
No, there's nothing listening on 6000. I'll try and debug it.

backend/test/auth/oidc/auth_test.go
illume: Yeah. Sorry I wasn't clear by putting the question here. 

Should it be used in "login_test.go" as well?

illume: Yeah, they should also do the skip. If you run `cd backend;make check` you'll see them fail if you don't have a db running.

illume:
Ok, quite a bit of messing about, and found a few things but not any fixes yet...

---

Seems we now have some sort of build issue with docker images not being rebuilt when things change. After pruning my containers and images I got things working in the last commit on proof-of-concept3 branch. I see this in the proof-of-concept3 branch, but haven't gone back to see when the issue was introduced yet.

However, even after pruning, I can't get this branch to pass tests (with `make check` or `make check-backend-with-container`). 

- Make check needs the skipping via env added.
- For testing with the container, seems a server is staying open after a test... which causes the next test to fail because the port is open(I think). Maybe a bit more error checking/logging and deferring the shutdown so it runs on a test failure would help here.

yolossn:
Oh got the reason for the docker images not building. I think we should add `--build` tag. I'll look into the the `make check` and the server not shutting down properly.

backend/test/auth/oidc/login_test.go
```diff
@@ -73,6 +74,35 @@ func TestOIDCAuthModeSetup(t *testing.T) {
 	})
 }
 
+var ErrOutOfRetries = errors.New("test: out of retries")
+
+func waitServerReady() (bool, error) {
+	retries := 5
+	for i := 0; i < retries; i++ {
+		if i != 0 {
+			time.Sleep(100 * time.Millisecond)
```
joaquimrocha, line 83: This could be simplified by moving to line 87, before the continue.

backend/test/auth/oidc/login_test.go
```diff
@@ -73,6 +74,35 @@ func TestOIDCAuthModeSetup(t *testing.T) {
...
+		}
+		req, err := http.NewRequest("GET", fmt.Sprintf("%s/health", testServerURL), nil)
```
joaquimrocha, line 85: I wonder if checking for the /config would achieve the same without adding a new endpoint. I still understand this endpoint is fairly innocent in any case.

backend/test/auth/oidc/login_test.go
```diff
@@ -127,11 +159,14 @@ func TestLogin(t *testing.T) {
 		//nolint:errcheck
 		go server.Start(serverPortStr)
 
-		time.Sleep(100 * time.Millisecond)
+		// time.Sleep(100 * time.Millisecond)
```
joaquimrocha, line 162: This should be removed when adding te waitServerReady to the rest of the code.

backend/test/auth/oidc/login_test.go
illume: The idea here is that it will only sleep if there is a failure to GET. So the first time through the loop there is no sleep. It should also sleep if there is any sort of error.

backend/test/auth/oidc/login_test.go
illume: Good idea. The only downsides I can see are it generates json and has a separate purpose.

backend/test/auth/oidc/login_test.go
joaquimrocha: Isn't that what the suggestion I made would accomplish?

illume: If it's inside the first error check if on line 87 then it would not sleep if the error happens after that in one of the other error check blocks.

backend/test/auth/oidc/login_test.go
illume: @yolossn if we keep /health we need to add it to the spec. Should I do this?

backend/test/auth/oidc/login_test.go
joaquimrocha: Okay. The logic still looks like it could be simplified IMO, but it's not grave how it is.
Let's move on with it.

backend/test/auth/oidc/login_test.go
joaquimrocha: I'll let you folks decide.
IMO, health is just checking if the server is up but not a full reflection that the service is running as the name/concept may suggest. So I'd use /config for that purpose and not add the new endpoint (unless we make it more about health than about testing if the server responds).

joaquimrocha:
@yolossn , once the final touches are done, and since you've mentioned it now also works for you, I think the commits should be squashed together onto your original one and merged.

PR #561, opened 2022-01-27
chart: Update postgres dependency 12.5.x -> 12.9.x

mkilchhofer:
# Update postgres dependency 12.5.x -> 12.9.x

Let's upgrade the postgres dependency to be in line with the current version from Bitnami.

Do you have a recommended postgres version to use together with nebraska? We could also use PG 13 or 14 if you want.

## How to use

nothing special

## Testing done

- [x] Manual testing in my own k8s environment.
- [x] Tests from pipeline are green

### DIFF
--- a/charts/nebraska/Chart.lock
+++ b/charts/nebraska/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 10.3.8
-digest: sha256:1e05bf878e384a1b34cef8c93c6e98b6bfc68f8ef31f8a06c220ba49c8e627ca
-generated: "2021-03-03T14:06:51.357213+01:00"
+  version: 10.16.2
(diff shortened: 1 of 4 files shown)


illume:
Both 13.x and 14.x have significant performance increases and reduced resource usage. We've been testing mostly with 13.3.
I think only Azure is offering 14.x so far (maybe some others now?), and not AWS etc yet.

Maybe 13.5 after some testing? What do you think @yolossn ?

mkilchhofer:
Just a note to myself, if we want to update the chart to use 13.5 or so, we need a major version bump as users need to migrate PG data manually:

```log
postgresql 18:15:20.07 INFO  ==> ** Starting PostgreSQL **
2
2022-01-27 18:15:20.177 GMT [1] FATAL:  database files are incompatible with server
1
2022-01-27 18:15:20.177 GMT [1] DETAIL:  The data directory was initialized by PostgreSQL version 12, which is not compatible with this version 13.5.
```

yolossn (approved):

yolossn:
Let's skip the pg version upgrade for now.

PR #537, opened 2021-12-17
Improve Docker image

guilhem:
# Improve Docker image

- [x] :chart_with_upwards_trend: caching
  - [x] separate front / back
  - [x] node_module caching
  - [x] go.mod caching  
- [x] :chart_with_downwards_trend: build time
  - [x] multistep build can be //
- [ ] :chart_with_downwards_trend: final size
  - [x] CGO_ENABLED=0
  - [ ] before 42M
- [x] :chart_with_upwards_trend: security
  - [x] run as user (not root)
- [x] :chart_with_upwards_trend: usability
  - [x] using `ENTRYPOINT`, only have to add some `ARGS` to customize run 

## How to use

- :warning: depends on #503 
- :warning:  Dockerkit should be installed (or any builder that accept `syntax = docker/dockerfile:1.3`)

PR #562, opened 2022-01-27
Integration test improvements

illume:
# Integration test improvements

- [x] Skip server integration tests by default with `make check` like with `make code-check`. So like before, make check only requires a running test db.
- [x] Start on docs related to backend testing. Document testing make targets, and document backend testing related environment variables.
- [x] Fix test-service-up to build image before starting, to prevent stale code being tested. This is not great at the moment, because of how the dockerfile is setup. A change in backend code, triggers frontend code to be rebuilt and vice versa. The PR https://github.com/kinvolk/nebraska/pull/537 should improve DX quite a bit when doing the checks in the container because it separates front end and backend well.
- [x] The localhost:8000 testing URL was the same as is used for the development server. I'm changing this to its own port, so that tests can be run with the container whilst the dev server is running.
- [x] With the NEBRASKA_DB_URL the test service starts with the same port as our docker run postgres. Meaning we can't run it at the same time. Bind for 0.0.0.0:5432 failed: port is already allocated So I'm probably going to change this to a separate port too. So make check-backend-with-container can run separately.  
- [x] Stops test services first, so that they do not keep running keeping state from previous test runs. So that other code changes and tests can not interfere with the latest test run.

# How to test

You can run the dev server, and have a test db at the same time as running `make check-backend-with-container`.

1) run a test db (in docker perhaps as per dev guide `docker run --rm -d --name nebraska-postgres-dev -p 5432:5432 -e POSTGRES_PASSWORD=nebraska postgres`)
2) `make run-backend` (to show that the tests can run whilst this is running and occupying a port)
3) Now `cd backend;make check-backend-with-container` should work, and it should run the integration tests. Before there would be errors about the shared ports being used.
4) `cd backend;make check` should work as long as you have a test db running (as before). It skips the server integration tests.
5) Should be able to run a single test separately. `cd backend && make test-service-down test-service-up && cd test/api && NEBRASKA_RUN_SERVER_TESTS=1 NEBRASKA_DB_URL="postgres://postgres:nebraska@127.0.0.1:8001/nebraska_tests?sslmode=disable&connect_timeout=10" go test -run TestListApp`

### DIFF
--- a/backend/test/api/api_test.go
+++ b/backend/test/api/api_test.go
@@ -14,7 +14,7 @@ const (
 )
 
 func TestMain(m *testing.M) {
-	if os.Getenv("NEBRASKA_SKIP_TESTS") != "" {
+	if os.Getenv("NEBRASKA_SKIP_TESTS") != "" || os.Getenv("NEBRASKA_RUN_SERVER_TESTS") == "" {
 		return
 	}
 
--- a/backend/test/api/api_test.go
+++ b/backend/test/api/api_test.go
@@ -9,8 +9,8 @@ import (
 )
 
 const (
-	testServerURL    = "http://localhost:8000"
-	defaultTestDbURL = "postgres://postgres:nebraska@127.0.0.1:5432/nebraska_tests?sslmode=disable&connect_timeout=10"
+	defaultTestServerURL = "http://localhost:8002"
+	defaultTestDbURL     = "postgres://postgres:nebraska@127.0.0.1:5432/nebraska_tests?sslmode=disable&connect_timeout=10"
 )
 
 func TestMain(m *testing.M) {
@@ -21,6 +21,11 @@ func TestMain(m *testing.M) {
 	if _, ok := os.LookupEnv("NEBRASKA_DB_URL"); !ok {
 		log.Printf("NEBRASKA_DB_URL not set, setting to default %q\n", defaultTestDbURL)
 		_ = os.Setenv("NEBRASKA_DB_URL", defaultTestDbURL)
+	}
+
+	if _, ok := os.LookupEnv("NEBRASKA_TEST_SERVER_URL"); !ok {
+		log.Printf("NEBRASKA_TEST_SERVER_URL not set, setting to default %q\n", defaultTestServerURL)
+		_ = os.Setenv("NEBRASKA_TEST_SERVER_URL", defaultTestServerURL)
 	}
 
 	a, err := api.New(api.OptionInitDB)
--- a/backend/test/api/app_test.go
+++ b/backend/test/api/app_test.go
@@ -3,6 +3,7 @@ package api_test
 import (
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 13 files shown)


backend/test/api/api_test.go
```diff
@@ -9,12 +9,12 @@ import (
...
 const (
-	testServerURL    = "http://localhost:8000"
-	defaultTestDbURL = "postgres://postgres:nebraska@127.0.0.1:5432/nebraska_tests?sslmode=disable&connect_timeout=10"
+	defaultTestServerURL = "http://localhost:8000"
+	defaultTestDbURL     = "postgres://postgres:nebraska@127.0.0.1:5432/nebraska_tests?sslmode=disable&connect_timeout=10"
 )
 
 func TestMain(m *testing.M) {
-	if os.Getenv("NEBRASKA_SKIP_TESTS") != "" {
+	if os.Getenv("NEBRASKA_SKIP_TESTS") != "" || os.Getenv("NEBRASKA_RUN_SERVER_TESTS") == "" {
```
yolossn, line 17: This can be solved by using golang test tags to separate the `integration` tests from the `unit` tests. We can retain the `NEBRASKA_SKIP_TESTS` and use the tags to run the corresponding tests from the make file.
Refer: https://mickey.dev/posts/go-build-tags-testing/

backend/test/api/app_test.go
```diff
@@ -26,7 +27,7 @@ func TestListApp(t *testing.T) {
 		require.NotNil(t, appsDB)
 
 		// fetch apps from the API
-		url := fmt.Sprintf("%s/api/apps", testServerURL)
+		url := fmt.Sprintf("%s/api/apps", os.Getenv("NEBRASKA_TEST_SERVER_URL"))
```
yolossn, line 30: Instead of using `os.Getenv` create a function that returns the default value if the value is not found in env. I see that the value is being set in the `TestMain` function but when we want to run a single test it will fail as the value is not set manually. 
Running a single test like `go test -run TestListApp  .` this will fail.

illume: Thanks, good idea.

illume:
I didn't have my coffee yet it seems. I'm still getting test errors on oidc_test after merging in integration-test-stuff branch. At least it's only failing on login_test.go now. I'll look into it (and do those review comments you left).

backend/test/api/app_test.go
illume: I tried this and TestListApp works...

```
cd backend && make test-service-down test-service-up && cd test/api && NEBRASKA_RUN_SERVER_TESTS=1 NEBRASKA_TEST_SERVER_URL="http://localhost:8002" NEBRASKA_DB_URL="postgres://postgres:nebraska@127.0.0.1:8001/nebraska_tests?sslmode=disable&connect_timeout=10" go test -run TestListApp
```

illume: Thanks @yolossn 

The issue was when running a single test, but NEBRASKA_TEST_SERVER_URL is not set (so the default will be used).

This is what I found:
- There was an incorrect port (pushed a fix just now for this)
- TestMain runs even if you select one of the tests to run.

This is the command that should work (note the missing NEBRASKA_TEST_SERVER_URL):
```
cd backend && make test-service-down test-service-up && cd test/api && NEBRASKA_RUN_SERVER_TESTS=1 NEBRASKA_DB_URL="postgres://postgres:nebraska@127.0.0.1:8001/nebraska_tests?sslmode=disable&connect_timeout=10" go test -run TestListApp
```

illume:
Hey hey. Should be ready to review now. There's two open conversations that I've replied to that I think could be close... but maybe they shouldn't be? Feel free to respond more in there if there's anything else to do. cheers!

backend/test/api/api_test.go
yolossn: Cool, makes sense.

PR #569, opened 2022-02-10
fix regression issues from openapi codegen changes

yolossn:
Fixes:
- API key validation for github auth mode.
- Package search not working when editing the channel.


yolossn:
Working on the pagination search bug in the same branch. So marked the PR as a draft.

joaquimrocha:
I have added the frontend bits. Also the "fixup or delete" commit is because the tests are using GetPackage and I thought maybe we keep this function and add the new ones as separate (they will need testing in both cases).

The only other nitpick is that `searchTerm` in the query seems a bit quirky to me, but ATM I cannot think of a much better one.

yolossn:
@joaquimrocha let's update the tests instead of creating a new function. I was also thinking about the `searchTerm` keyword, should it be `searchVersion` ?

yolossn:
@joaquimrocha When the search term is removed the packages are not updating


https://user-images.githubusercontent.com/22731013/154635479-afc23fa1-d9b2-4729-9241-c7757a0e8241.mov

joaquimrocha:
> @joaquimrocha let's update the tests instead of creating a new function. I was also thinking about the `searchTerm` keyword, should it be `searchVersion` ?

Let's go with matchVersion if it just matches the version.

PR #571, opened 2022-02-11
docs: update Flatcar usage section

pothos:
Nowadays Nebraska supports group aliases and the UUID is not needed
anymore. While the whitespace in the machine alias works for
update_engine there are some other parts in Flatcar that try to source
the update.conf file and fail when it has unqouted whitespace. Also,
the restart of update-engine.service is not really needed and it's
probably good to point the user to the Flatcar docs for more info.

### DIFF
--- a/docs/managing-updates.md
+++ b/docs/managing-updates.md
@@ -25,7 +25,7 @@ storage:
           SERVER=http://your.nebraska.host:port/v1/update/
 ```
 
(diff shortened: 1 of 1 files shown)


illume (approved):
Removing group id stuff which is not needed anymore also makes this cleaner. 🎈
Using correct quoting seems a good thing to do. 🎉 

Thanks.

PR #577, opened 2022-02-23
Multistage docker build

yolossn:
This PR is based on the inital work by @guilhem on https://github.com/kinvolk/nebraska/pull/537. Instead of docker buildkit the docker layer caching is used for caching the dependency download on each build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-# Backend build
-FROM golang:1.17 as backend-build
+FROM golang:1.17 as base-build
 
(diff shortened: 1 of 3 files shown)


yolossn:
@illume Can you please take a look at the frontend build part ? Will we need the intermediate step that you added in headlamp here as well ?

illume:
Added some commits to make the backend and frontend not rebuild unless needed.

guilhem:
🚀

PR #578, opened 2022-02-23
backend: Fix flatcar packages path bug

yolossn:
This patch fixes a bug that redirected the
flatcar packages request to the index.html
file.

### DIFF
--- a/backend/pkg/server/server.go
+++ b/backend/pkg/server/server.go
@@ -36,8 +36,7 @@ var (
 	logger            = util.NewLogger("nebraska")
 	middlewareSkipper = func(c echo.Context) bool {
 		requestPath := c.Path()
-
-		paths := []string{"/health", "/metrics", "/config", "/v1/update", "/*"}
+		paths := []string{"/health", "/metrics", "/config", "/v1/update", "/flatcar/*", "/*"}
 		for _, path := range paths {
 			if requestPath == path {
 				return true
@@ -108,12 +107,18 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 
 	e.Static("/", conf.HTTPStaticDir)
 
+	if conf.HostFlatcarPackages && conf.FlatcarPackagesPath != "" {
+		e.Static("/flatcar/", conf.FlatcarPackagesPath)
+	}
+
 	e.HTTPErrorHandler = func(err error, c echo.Context) {
 		code := http.StatusNotFound
 		if he, ok := err.(*echo.HTTPError); ok {
 			if code == he.Code {
 				fileErr := c.File(path.Join(conf.HTTPStaticDir, "index.html"))
-				logger.Err(fileErr).Msg("Error serving index.html")
+				if fileErr != nil {
+					logger.Err(fileErr).Msg("Error serving index.html")
+				}
 				return
 			}
 		}
--- a/backend/test/api/flatcar_package_test.go
+++ b/backend/test/api/flatcar_package_test.go
@@ -1,0 +1,97 @@
+package api_test
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/kinvolk/nebraska/backend/pkg/config"
+	"github.com/kinvolk/nebraska/backend/pkg/server"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHostFlatcarPackage(t *testing.T) {
+	currentDir, err := os.Getwd()
+	require.NoError(t, err)
+
+	serverPort := uint(6000)
+	serverPortStr := fmt.Sprintf(":%d", serverPort)
+
+	conf := &config.Config{
+		HostFlatcarPackages: true,
+		FlatcarPackagesPath: currentDir,
+		AuthMode:            "noop",
+		ServerPort:          serverPort,
+	}
+
+	db := newDBForTest(t)
+
+	t.Run("file_exists", func(t *testing.T) {
+		server, err := server.New(conf, db)
+		require.NotNil(t, server)
+		require.NoError(t, err)
+
+		//nolint:errcheck
+		go server.Start(serverPortStr)
+
+		//nolint:errcheck
+		defer server.Shutdown(context.Background())
+
+		// create a temp file
+		fileName := fmt.Sprintf("%s.txt", uuid.NewString())
+		file, err := os.Create(path.Join(currentDir, fileName))
+		require.NoError(t, err)
+
+		fileString := "This is a test"
+		_, err = file.WriteString(fileString)
+		require.NoError(t, err)
+		err = file.Close()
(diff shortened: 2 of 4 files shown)


backend/pkg/server/server.go
```diff
@@ -112,8 +112,22 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 		code := http.StatusNotFound
 		if he, ok := err.(*echo.HTTPError); ok {
 			if code == he.Code {
+				dir, file := path.Split(c.Request().URL.Path)
+				if dir == "/flatcar/" {
```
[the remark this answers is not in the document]
jepio: Also this needs to be conditional on -host-flatcar-packages.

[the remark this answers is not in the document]
yolossn: Hey @jepio, This is because these endpoints are dynamically created and we follow Open API documentation to generate handlers for the defined set of endpoints. So these dynamic endpoints are considered "Not found (404)" by default and that is why it is handled in the error handler.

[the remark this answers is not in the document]
yolossn: >But the error handler only resolves the index.html in case of 404 file, other static assets are being served via e.Static(), or am I wrong? The flatcar packages are also static assets so it would seem cleaner to serve them the same way. Does open API not make a distinction between API endpoints and static assets?

There are dynamic endpoints in the frontend like "nebraska.com/application/:appId" etc in this case it is redirected to the `index.html`
> conf.FlatcarPackagesPath defaults to "", so it would serve arbitrary files from the working directory of nebraska right? That doesn't sound safe.

Aah, good catch. I thought anyways the file wont be available and 404 will be returned. Will add another level of check.

backend/pkg/server/server.go
```diff
@@ -112,8 +112,22 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 		code := http.StatusNotFound
 		if he, ok := err.(*echo.HTTPError); ok {
 			if code == he.Code {
+				if conf.HostFlatcarPackages {
+					dir, file := path.Split(c.Request().URL.Path)
+					if dir == "/flatcar/" {
```
joaquimrocha, line 117: The 2 occurrences of `/flatcar/` should be used from a const, to avoid future problems.

backend/test/api/flatcar_package_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	db := newDBForTest(t)
+
+	t.Run("file_exists", func(t *testing.T) {
+		server, err := server.New(conf, db)
+		require.NotNil(t, server)
+		require.NoError(t, err)
+
+		//nolint:errcheck
+		go server.Start(serverPortStr)
```
joaquimrocha, line 41: Don't you have to check for whether the server is already up, otherwise we may have a race issue like before.

joaquimrocha (changes requested):

backend/pkg/server/server.go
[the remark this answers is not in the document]
illume: I think we need to consider the topic of dynamic endpoints and the schema in depth a bit. It would be good to represent dynamic endpoints in the schema somehow. I can't remember if we looked into this already for the other dynamic end points?

teizz:
Sorry to bump in like this. Reading the thread I didn't really find the answer to the question by @jepio why this couldn't be solved with a `e.Static`. Handling serving files from the error handler doesn't seem like the most elegant solution imho.

I'm not at all familiar with the code, or very experienced with Echo, but having added the following works for me:
```
diff --git a/backend/pkg/server/server.go b/backend/pkg/server/server.go
index 9b7c57b..81e1d9a 100644
--- a/backend/pkg/server/server.go
+++ b/backend/pkg/server/server.go
@@ -37,7 +37,7 @@ var (
        middlewareSkipper = func(c echo.Context) bool {
                requestPath := c.Path()
 
-               paths := []string{"/health", "/metrics", "/config", "/v1/update", "/*"}
+               paths := []string{"/health", "/metrics", "/config", "/v1/update", "/flatcar/*", "/*"}
                for _, path := range paths {
                        if requestPath == path {
                                return true
@@ -106,6 +106,9 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
                return nil, fmt.Errorf("Error setting up handlers: %w", err)
        }
 
+       if conf.HostFlatcarPackages && conf.FlatcarPackagesPath != "" {
+               e.Static("/flatcar/", conf.FlatcarPackagesPath)
+       }
        e.Static("/", conf.HTTPStaticDir)
 
        e.HTTPErrorHandler = func(err error, c echo.Context) {
```

yolossn:
@teizz I didn't think of adding the `/flatcar/*` path to the skipper and the OpenAPI validation middleware was returning "Status 404" as it is not defined in the specification YAML. So I handled it using the `ErrorHandler` like we previously handled the dynamic endpoints from the frontend that had to be redirected to `index.html`. Sorry my bad, Thanks for pointing it out. I'll update the PR.

joaquimrocha (approved):

joaquimrocha:
LGTM, but maybe @illume wants to take an extra look before merging.

illume:
Hrmm. I can't get `make check-backend-with-container` working for me locally. Looking into it.

illume:
Added waitServerReady to the second test and it worked...

```golang
(flatcar-path-fix)$ git diff
diff --git a/backend/test/api/flatcar_package_test.go b/backend/test/api/flatcar_package_test.go
index bde48cb..3bd21bd 100644
--- a/backend/test/api/flatcar_package_test.go
+++ b/backend/test/api/flatcar_package_test.go
@@ -85,6 +85,9 @@ func TestHostFlatcarPackage(t *testing.T) {
                //nolint:errcheck
                defer server.Shutdown(context.Background())

+               _, err = waitServerReady(fmt.Sprintf("http://localhost:%d", serverPort))
+               require.NoError(t, err)
+
                resp, err := http.Get(fmt.Sprintf("http://localhost:%d/flatcar/%s", serverPort, fileName))
                assert.NoError(t, err)
                assert.NotNil(t, resp)
```

backend/test/api/flatcar_package_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+func TestHostFlatcarPackage(t *testing.T) {
+	currentDir, err := os.Getwd()
+	require.NoError(t, err)
+
+	serverPort := uint(6000)
+	serverPortStr := fmt.Sprintf(":%d", serverPort)
+
+	conf := &config.Config{
+		HostFlatcarPackages: true,
```
illume, line 27: I don't know if it's important to define what `HostFlatcarPackages: false` does?

illume (approved):
Thanks for fixing this, and for adding the tests! 👍 🎉 

(Looks good to me when the wait patch is applied)

backend/test/api/flatcar_package_test.go
yolossn: Do you mean add test cases for `HostFlatcarPackages: false`?

illume: I personally don't think it matters...

PR #579, opened 2022-02-25
Add multi package support

### LINKED ISSUE #473: [RFE] Ship multiple files with one update, let client pick which files to download
## Current situation

From a Nebraska client perspective, any given version of an application relates to a single file. This implies that updates to an application must consist of a single file only (commonly referred to as the "update payload").

## Impact

It's not possible to ship multiple files with the same update, resulting in the need to bundle all files in a single archive even if the client only requires a subset of files to fully apply the update.
More specifically, this prevents the Flatcar Container Linux project from efficiently shipping OEM partition updates ("vendor tools") alongside OS image updates.

## Ideal future situation

Instead of a single file per version update providers *optionally* can provide a list of files. All files are cached by Nebraska (stand-alone use case for e.g. air-gapped environments) and a list of files is sent to clients querying for new versions. Clients will then decide which files to download and to install in order to fully update to the new version.

It is desirable to retain the current behaviour for single-file updates for backwards compatibility.

## Additional information

The Flatcar feature is tracked in https://github.com/flatcar-linux/Flatcar/issues/60 .

joaquimrocha:
- [X] Add DB and backend support
- [X] Add UI
- [X] Add base testing
- [X] Ensure the files' positions are kept as originally added, after edition
- [X] Make sure the syncer can retrieve also more than one file

fixes #473


frontend/src/js/components/Common/MoreMenu.tsx
```diff
@@ -7,7 +7,15 @@ import { useTranslation } from 'react-i18next';
 
 let menuCount = 0;
 
-export default function MoreMenu(props: { options: { label: string; action: () => void }[] }) {
+interface MoreMenuProps {
+  options: {
+    label: string;
+    action: () => void;
+  }[];
+  iconButtonProps?: { [key: string]: any };
```
illume, line 15: Something like this to have the props typed (and not use any)?
```tsx
type MoreMenuProps {
  options: {
    label: string;
    action: () => void;
  }[];
} & React.ComponentProps<typeof IconButton>;
```

backend/pkg/api/packages.go
```diff
@@ -33,6 +33,20 @@ var (
...
 )
 
+type File struct {
+	ID        int64       `db:"id" json:"id"`
+	PackageID string      `db:"package_id" json:"package_id"`
+	Name      null.String `db:"name" json:"name"`
+	Size      null.String `db:"size" json:"size"`
+	Hash      null.String `db:"hash" json:"hash"`
+	CreatedTs time.Time   `db:"created_ts" json:"created_ts"`
+	Position  int         `db:"position" json:"position"`
```
pothos, line 43: What should the Omaha `required` value be for these? I think it makes sense to set them to false by default to distinguish it from the first entry or make it configurable - for Flatcar this could be interesting when not only OEMs but also other additional payloads get added.

pothos: Ok, looks like this should be false already, only for the first package it seems to be set: https://github.com/kinvolk/nebraska/blob/dfc08d64bc1a453a4dc411db66139650222b7985/backend/pkg/omaha/omaha.go#L217

backend/pkg/api/packages.go
```diff
@@ -481,3 +551,65 @@ func (api *API) updatePackageBlacklistedChannels(tx *sqlx.Tx, pkg *Package) erro
...
 }
+
+// updatePackageFiles adds or removes as needed files to the
+// package's extra files list based on the new entries provided in the updated
+// package entry.
+//
+// This method is part of the transaction that updates a package and when it's
+// called, the package has already been updated except for the files, that may
+// happen here if needed.
+func (api *API) updatePackageFiles(tx *sqlx.Tx, pkg *Package, oldPkg *Package) error {
```
yolossn, line 562: The logic to maintain order is a bit complex and can lead to long transactions. When there are n files and the first file is deleted there will `n-1` inserts and `n-1` delete operations. 

There can be 3 cases when the order has to be changed.
1. Reorder the files. (Not possible right now with the UI).
2. Delete a file.
3. Create a file. (Right now there is no provision in the frontend, for a user to create a file in between the existing files, the user will have to delete the files after the positon to add and then recreate those deleted files).

There can be two ways in which we approach this for better efficiency. 
1. Store the order of the child (files) in the parent(package) as an array of fileIds and the position in the array becomes the position of the file. 
2. Instead of storing the absolute position in the files, store the previousid (like a linked list).

I think storing the position in package based on array position is a better idea. As the files are altered only in package update/create flow.

- can change the order with one update query.
- inserting/deleting a file is also 2 queries insert/delete in files table and order can be updated in one query.

yolossn: Created a sample to check how this will work

```SQL
-- create package
create table package (
	id VARCHAR,
	name VARCHAR,
	file VARCHAR[]
)

-- insert package
INSERT INTO "package" ("id", "name", "file") VALUES ('xys', 'package 1', '{}');

-- create file
create table file(
	id VARCHAR,
	name VARCHAR,
	hash VARCHAR
)

-- insert files
insert into "file" ("id","name","hash") values('a','file a','qwerty'),('b','file b','asdfg'),('c','file c','zxcvb');

-- set file order in package
update "package" set "file"='{"a","b","c"}' where "id"='xys'

-- update order in package
update "package" set "file"='{"b","a","c"}' where "id"='xys'

-- select file in order based on array position
select file.* from file join unnest('{"b","a","c"}'::VARCHAR[]) WITH ORDINALITY t(id,ord) using(id) ORDER by t.ord
```

backend/pkg/omaha/omaha.go
```diff
@@ -216,6 +216,22 @@ func (h *Handler) prepareUpdateCheck(appResp *omahaSpec.AppResponse, pkg *api.Pa
 	}
 	mpkg.Required = true
 
+	for _, pkgFile := range pkg.ExtraFiles {
+		fpkg := manifest.AddPackage()
```
yolossn, line 220: Every extra file adds another package in the Omaha response. Shouldn't this be multiple files under a single package?

yolossn: If this is intentional to treat extra files as packages in the Omaha response, should we allow the user to set separate URLs for the extra files?

backend/pkg/api/packages.go
pothos: @joaquimrocha Do we need a certain order? I think it's also fine to just rely on the order the DB spits out, or?

backend/pkg/omaha/omaha.go
pothos: It's multiple packages entries, a package can only have one file in Omaha.

frontend/src/js/components/Common/Tabs.tsx
```diff
@@ -0,0 +1,114 @@
...
+  React.useEffect(
+    () => {
+      if (defaultIndex === null) {
+        setTabIndex(false);
+        return;
+      }
+      setTabIndex(defaultIndex);
+    },
+    // eslint-disable-next-line
+    [defaultIndex]
```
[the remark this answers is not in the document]
joaquimrocha: I copied the file from Headlamp as stated in the commit. I think if we are improving this file, we should do it in a different PR.

frontend/src/js/components/Common/MoreMenu.tsx
```diff
@@ -33,6 +41,7 @@ export default function MoreMenu(props: { options: { label: string; action: () =
         aria-label={t('common|Open menu')}
         onClick={handleClick}
         data-testid="more-menu-open-button"
```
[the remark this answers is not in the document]
joaquimrocha: You're right I don't find it anywhere. But it should be fixed in a different PR (by tackling any other unused data-testid as well).

backend/pkg/api/packages.go
joaquimrocha: Right. I chose not to set it nor support changing it in Nebraska on purpose for now (so it takes the default value).

backend/pkg/omaha/omaha.go
joaquimrocha: Right. Nebraska's DB and Omaha do not match 100% (and they don't have to).

backend/pkg/syncer/syncer.go
```diff
@@ -354,7 +384,7 @@ func getArchString(arch api.Arch) string {
 // downloadPackage downloads and verifies the package payload referenced in the
 // update provided. The downloaded package payload is stored in packagesPath
 // using the filename provided.
-func (s *Syncer) downloadPackage(update *omaha.UpdateResponse, filename string) error {
+func (s *Syncer) downloadPackage(update *omaha.UpdateResponse, pkgName, sha256Checksum, filename string) error {
```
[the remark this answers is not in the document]
joaquimrocha: Hmm, it is used in the function. To rename the downloaded package, line 424, right?

joaquimrocha:
> It's not clear to me why you didn't make extraFiles just files, and get rid of the existing file fields? Seems like it would make it less code and edge cases. I guess because it's quite an invasive change to do that?

Yeah, I considered that, but it'd require a migration, and since we require at least one file to be associated with the package (the reason why the legacy code-base had the file as part of the package), I thought we may as well keep it like this.
I think the whole DB will require a re-design at some point, and these changes shouldn't make it more complicated than it is now to migrate at some point.

backend/pkg/syncer/syncer.go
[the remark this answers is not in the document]
yolossn: My bad sorry

backend/pkg/api/packages.go
yolossn: If the position doesn't make anychanges in the omaha response should there be order in the UI ? I was thinking that the order will used by the omaha response too. 

> @yolossn , I think that if we storing the position in an array, we may as well just store the whole files as a json column in the package. So we always have single table transactions.

I agree lets go with the existing flow itself considering that the frequency of such change is very less, but given that there will only be `write` and `read` on extra_file `json` can also be used here.

joaquimrocha: The order is reflected in the package list of the omaha response. Now, whether a client uses it or thinks it's important, is a different matter.

illume: @yolossn way sounds good, or you can do a single statement insert and delete perhaps. Not sure how to do that with goqu off the top of my head (though it's possible with SQL)!

backend/pkg/omaha/omaha.go
```diff
@@ -216,6 +216,22 @@ func (h *Handler) prepareUpdateCheck(appResp *omahaSpec.AppResponse, pkg *api.Pa
...
+		fpkg.Name = pkgFile.Name.String
+		if pkgFile.Hash.Valid {
+			fpkg.SHA1 = pkgFile.Hash.String
```
illume, line 223: Does sha256 need to be handled here?

backend/pkg/api/packages_test.go
```diff
@@ -305,3 +305,112 @@ func TestGetPackages(t *testing.T) {
...
+
+	pkg, err = a.AddPackage(pkg)
+	assert.NoError(t, err)
+	assert.Nil(t, pkg.ExtraFiles)
+
+	pkg.ExtraFiles = []File{
+		{
+			Name: null.StringFrom("myfile1.txt"),
+			Size: null.StringFrom("1234"),
+			Hash: null.StringFrom("abcd"),
```
illume, line 335: I'm sure if we need it... but do we need to test sha1/sha256?

backend/pkg/syncer/syncer.go
```diff
@@ -291,12 +291,41 @@ func (s *Syncer) processUpdate(descriptor channelDescriptor, update *omaha.Updat
...
+					Size: null.StringFrom(size),
+					Hash: null.StringFrom(hash),
+				}
+			}
+		}
+
 		if s.hostPackages {
 			filename = fmt.Sprintf("flatcar-%s-%s.gz", getArchString(descriptor.arch), update.Manifest.Version)
-			if err := s.downloadPackage(update, filename); err != nil {
+			if err := s.downloadPackage(update, update.Manifest.Packages[0].Name, update.Manifest.Actions[0].SHA256, filename); err != nil {
```
illume, line 317: Another sha256/sha1 question goes here.

backend/pkg/omaha/omaha.go
joaquimrocha: I noticed that it's SHA1 that's being used in the other cases... Probably we should assign it to both or check for sha256: prefixes.

backend/pkg/api/packages_test.go
joaquimrocha: The checksums checks will be dependent on the client's implementation, but I agree we should test the values in the tests here.

backend/pkg/syncer/syncer.go
joaquimrocha: I will update.

frontend/src/js/components/Packages/EditDialog.tsx
```diff
@@ -156,123 +175,163 @@ function EditDialog(props: { data: any; show: boolean; create?: boolean; onHide:
...
+                                        ? t('packages|channel pointing to this package')
+                                        : null
+                                    }
+                                  />
+                                </MenuItem>
+                              );
+                            })}
+                        </Field>
+                        <FormHelperText>
+                          Blacklisted channels cannot point to this package.
```
illume, line 306: They weren't translated before... but these should be translated at some point.

backend/pkg/syncer/syncer.go
joaquimrocha: This is for flatcar, so I have used what was being used before. But I have updated the part that was just setting up SHA1 now, so it should be good to go.

backend/pkg/api/packages_test.go
joaquimrocha: Done.

backend/pkg/omaha/omaha.go
joaquimrocha: I have added both SHA1 and SHA256 assignments now. I think it's enough because only one shall be used anyway.

backend/pkg/api/packages.go
illume: It was decided to do query optimization work later using the relational tables as it is, so resolving this.

illume:
When I run `make backend` the lint fix changes backend/test/api/flatcar_package_test.go

```diff
22:05:54-rene~/dev/nebraska (multi-pkg)$ git diff
diff --git a/backend/test/api/flatcar_package_test.go b/backend/test/api/flatcar_package_test.go
index 3bd21bd..5fdc1aa 100644
--- a/backend/test/api/flatcar_package_test.go
+++ b/backend/test/api/flatcar_package_test.go
@@ -10,10 +10,11 @@ import (
        "testing"

        "github.com/google/uuid"
-       "github.com/kinvolk/nebraska/backend/pkg/config"
-       "github.com/kinvolk/nebraska/backend/pkg/server"
        "github.com/stretchr/testify/assert"
        "github.com/stretchr/testify/require"
+
+       "github.com/kinvolk/nebraska/backend/pkg/config"
+       "github.com/kinvolk/nebraska/backend/pkg/server"
 )

 func TestHostFlatcarPackage(t *testing.T) {
```

backend/pkg/api/packages.go
```diff
@@ -391,6 +427,32 @@ func (api *API) getFlatcarAction(packageID string) (*FlatcarAction, error) {
 	return &flatcarAction, nil
 }
 
+func (api *API) getExtraFiles(packageID string) ([]File, error) {
+	query, _, err := goqu.From("package_file").Where(goqu.C("package_id").Eq(packageID)).Order(goqu.C("id").Asc()).ToSQL()
```
yolossn, line 431: @joaquimrocha I see that the `id` field is `serial` and is used to `Order` here. Is this intentional? I understood that order is removed completely from the `package_files`, am I missing something here?

yolossn: Just FYI, I see that we use the serial value of `id` to store the position. We will not be able to support reordering of the files in the future. Every time the user will have to delete the files and then create them in the order that they intend to.

joaquimrocha: Yes, but we also don't allow reordering right now. So if we introduce reordering, we can make the selection depend on a new mechanism (like the array you proposed before). Until then, all this code guarantees is that when you modify a file, it'll still be in the same order as originally.

pothos:
Thanks everyone!

PR #581, opened 2022-03-10
Fix package version search

joaquimrocha:


illume:
Could you please write a, how you tested, section in the PR description?

illume:
The packages are listed in there, they are all dynamically loaded, and when I save it is persisted.

There's an odd problem when I select a package then click Cancel... in that it shows the package I selected in the channel edit dialog, but when I then click Save in the channel edit dialog it does not persist the selected package. IMHO I would not classify this as a release blocking problem, because the user did click Cancel... but it's still a problem. It should show the previous selected package in the channel edit dialog if the user presses Cancel in the package chooser dialog.

PR #582, opened 2022-03-17
Dockerfile backend: Fix VERSION not in container image

illume:
# Version was not in container image

`make image` should now echo the VERSION that is used.

## Testing done

- [x] `make image` shows version
- [x] change a backend .go file, `make image` should not rebuild frontend
- [x] change a frontend file, `make image` should not rebuild backend
- [x] change a frontend file, `make image` should not reinstall frontend npm packages
- [x] passing version via build-arg works and is echoed during build `--build-arg NEBRASKA_VERSION=meow`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,13 +9,25 @@ ENV GOPATH=/go \
 	GOOS=linux 
 
 # Backend build
-FROM base-build as backend-build
+FROM base-build as version-build
+ARG NEBRASKA_VERSION=""
+
+WORKDIR /app/
+COPY ./.git ./.git
+COPY ./backend ./backend
+
 # We optionally allow to set the version to display for the image.
 # This is mainly used because when copying the source dir, docker will
 # ignore the files we requested it to, and thus produce a "dirty" build
 # as git status returns changes (when effectively for the built source
 # there's none).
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -9,13 +9,22 @@ ENV GOPATH=/go \
...
 # This is mainly used because when copying the source dir, docker will
 # ignore the files we requested it to, and thus produce a "dirty" build
 # as git status returns changes (when effectively for the built source
 # there's none).
 ENV VERSION=${NEBRASKA_VERSION}
+
+FROM version-build as backend-build
+
+# Use the existing VERSION, or get it from make version if not there.
+RUN export VERSION=`make -f backend/Makefile version | tail -1` && echo "VERSION:$VERSION"
```
joaquimrocha, line 26: I am puzzled as to why this won't just override the VERSION from line 21. If it's been tested and it doesn't, then carry on.

joaquimrocha (approved):
Approved with a comment.

Dockerfile
illume: Yeah, it's a bit magical. I updated the comment to hopefully make it more clear?

> make version uses the existing VERSION if set, otherwise gets it from git

PR #593, opened 2022-04-06
backend: Fix Version Breakdown graph

yolossn:
This patch breaks the version breakdown query
into multiple sub queries and runs them in
parallel using go routines for better
performance.

### DIFF
--- a/backend/pkg/api/groups.go
+++ b/backend/pkg/api/groups.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/doug-martin/goqu/v9"
 	"github.com/google/uuid"
-	"github.com/lib/pq"
+	"golang.org/x/sync/errgroup"
 	"gopkg.in/guregu/null.v4"
 )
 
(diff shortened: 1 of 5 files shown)


yolossn:
The query is broken into 3 different parts and then the final output is computed, and the queries are run in parallel for performance.
- Instances without any instance status history (ie instances that have not been updated), In this case, the latest version in instance_application table is used.
- Instances that have updated in the timespan (30 days/7days etc), then fetch the instance status histories and compute the timeline.
- Instances that have been updated but not in the timespan. In this case, consider the latest instance status history version for all the spans in the timespan.

yolossn:
ToDo:

- [ ] Check if new indexes can be added 
- [ ]  Remove the indexes that were introduced for the old query.

yolossn:
The response time for 280000 instances.
image

backend/pkg/api/groups.go
```diff
@@ -648,81 +669,108 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
...
-	}
-	defer rows.Close()
-	for rows.Next() {
-		var timelineEntryEntity VersionCountTimelineEntry
-		err := rows.StructScan(&timelineEntryEntity)
+		instancesWithoutStatusQuery := fmt.Sprintf(`
+		with time_series as ( select * from generate_series( now() - interval '%[2]s', now(), interval '%[3]s' ) as ts ), instances as ( select ia.instance_id, case when last_update_granted_ts is not null then last_update_granted_ts else ia.created_ts end, ia."version" from instance_application ia left join ( select * from instance_status_history where group_id = '%[1]s' and created_ts >= now() - interval '%[4]s' and status = 4 ) ish on ia.instance_id = ish.instance_id where (ia."group_id" = '%[1]s') and last_check_for_updates >= now() - interval '%[2]s' and ( ia.instance_id is null or ia.instance_id not like '{________-____-____-____-____________}' ) and ia.version not like '%%nightly%%' and ish.instance_id is null ) select ts, ( case when version is null then '' else version end ), sum( case when version is not null then 1 else 0 end ) total from ( select * from time_series left join ( select * from instances ) _ on created_ts <= time_series.ts ) as _ group by 1, 2 order by ts desc;
```
joaquimrocha, line 687: Can we break this line for better readability?

backend/pkg/api/groups.go
```diff
@@ -753,6 +801,40 @@ func (api *API) GetGroupVersionCountTimeline(groupID string, duration string) (m
...
+		}
+	}
+
+	// post query processing for active instances without instance_status_history in the interval
+	for _, vc := range versionCounts {
+		if isNightlyVersion(vc.Version) {
+			continue
+		}
+		for _, span := range spans {
+			_, ok := timelineCount[span][vc.Version]
```
joaquimrocha, line 829: Nit: I'd get the result here, and if "not ok", I'd set up the result as 0. So below you'd have only to assign the value to the result + unit64(vc.Count) once.

joaquimrocha (commented):
Thanks for all this work. I know things were already very difficult to understand before, but it was one query. Now it's several steps, so I'd like to ask for a bit more detailed comments in each one, so next time we have to touch it, we can understand it quicker.

joaquimrocha (changes requested):
I think you meant to add the comment to the previous commit, not the indexes one, right?
The comments, LGTM, so after this tidy up, it should be good to merge.

joaquimrocha (approved):

joaquimrocha:
Awesome. Thanks.

PR #597, opened 2022-04-19
frontend: Add 404 page

yolossn:


illume (changes requested):
Thanks for adding this.

Should use appropriate headers (h1 and h2) and change the size to match(Check the dev console for accessibility errors/warnings). The text should use translation.

PR #602, opened 2022-04-29
Backend: Allow accessing resources via API using ProductID

yolossn:


backend/pkg/handler/channels.go
```diff
@@ -65,6 +75,10 @@ func (h *Handler) CreateChannel(ctx echo.Context, appID string) error {
 }
 
 func (h *Handler) GetChannel(ctx echo.Context, appID string, channelID string) error {
+	appID, err := h.db.GetAppID(appID)
+	if err != nil {
+		return appNotFoundResponse(ctx, appID)
+	}
```
joaquimrocha, line 81: We aren't using the appID here, AFAICS, maybe we don't need the extra query?

joaquimrocha (commented):
I have avoided using the productID beyond the initial request so we didn't have to change the whole codebase to include it.

I see you preferred to move the check to the handler functions, rather than the API ones.
This may be more efficient since we can perform API functions without having to check for the ID, but maybe it's more error prone?

I'd also verify the places that receive the appID as a parameter but we're not using it for much (or can get it from the channel/package/etc.). There's at least a comment I left about that.

backend/pkg/handler/channels.go
yolossn: I moved the check to handler function because of two reasons, 
1. There are handler functions where the appId is used in more than one `api` functions like `PaginateChannels`, So it made sense to find the appID once and reuse it.
2. If the `appID` is not found for the `productID` the response will be a `App not found error` instead of Internal Server Error, we can bubble up the error but keeping it this way makes it more clear both at handler level and api level, At handler level it helps for better error handling and at api level the functions do one specific thing.

yolossn:
> I see you preferred to move the check to the handler functions, rather than the API ones. This may be more efficient since we can perform API functions without having to check for the ID, but maybe it's more error prone?

I moved the check to handler function because of two reasons, 
1. There are handler functions where the appId is used in more than one `api` functions like `PaginateChannels`, So it made sense to find the appID once and reuse it.
2. If the `appID` is not found for the `productID` the response will be a `App not found error` instead of Internal Server Error, we can bubble up the error but keeping it this way makes it more clear both at handler level and api level, At handler level it helps for better error handling and at api level the functions do one specific thing (also there are api functions that expect params as a struct and there can be chances where the `GetAppID` could be missed).

backend/pkg/handler/channels.go
yolossn: Ill remove these extra ones.

joaquimrocha (approved):

PR #604, opened 2022-05-02
backend: Add migration down support

yolossn:


yolossn:
The `migrate-down` command should be used to remove the last migrations before reverting to an old version of Nebraska. One question still remains, how will the admin know how many migrations should be reverted ? Should we start adding the migrations introduced along with the releases going on ? @joaquimrocha  thoughts ?

joaquimrocha:
> The `migrate-down` command should be used to remove the last migrations before reverting to an old version of Nebraska. One question still remains, how will the admin know how many migrations should be reverted ? Should we start adding the migrations introduced along with the releases going on ? @joaquimrocha thoughts ?

I think maybe we can an argument support to migrate-down, so users can do: `--migrate-down=004` (exclusive, i.e. down to 004 which will be the last migration applied applied).

Users can use git to get the last migration for a past version:
`git show 2.4.0:./backend/pkg/api/db/migrations | tail -n1`

Alternatively we could add a command called `db-migrate` with the following options:
`nebraska db-migrate --down-to=004`: this migrations down till a migration starting with 004 (this one is not migrated down, i.e. it's exclusive)
`nebraska db-migrate --show`: this lists the current migrations for the current version.

(but this command would only be available in new versions, so maybe it's not so useful when users have git)

WDYT?

yolossn:
Makes sense to use the migration id to migrate down, I think we should add documentation for the same.

yolossn:
Makes sense to use the migration id for migrate down, I will add the documentation for the same.

yolossn:
Added docs for migrating to an older version of Nebraska.

illume:
> Added test 👍🏽

Awesome. Thanks!

docs/migrate-to-older-version.md
```diff
@@ -0,0 +1,21 @@
...
+---
+ 
+
+When migrating down to an older version of Nebraska one might have to rollback database migrations that were applied as a part of the release. 
+
+First get the migrations that were introduced as a part of the last release using the following command.
+
+> git show <NEBRASKA-VERSION>:./backend/pkg/api/db/migrations | tail -n1
+
+For example to find the migrations that were introduced as part of `2.4.0` release clone the Nebraska repo and run.
```
joaquimrocha, line 13: I think this article is a bit confusing. The command associated with this phrase doesn't necessarily show the newly added files in this folder for this tag, it just lists the contents of that dir, for that tag (so it is the same result for 2.4.1).

It should instead say that if users want to migrate to a version, like 2.4.0, they should run the command you wrote (gets them the last migration considered in 2.4.0) and run the `--migrate-down-to 0013`.

joaquimrocha (changes requested):
Found an issue in the docs.
BTW, `--migrate-down-to` sounds a bit complicated and is unrelated to DBs. Should we rename it as `--migrate-db-to` and support migrating up as well? (maybe in the beginning we can just support migrating down, but it's important that the command name makes sense)

backend/pkg/config/config.go
```diff
@@ -103,6 +104,7 @@ func Parse() (*Config, error) {
 	var config Config
 
 	f := flag.NewFlagSet("config", flag.ContinueOnError)
+	f.String("rollback-db-to", "", "Rollback db migration to the provided id, eg:0003")
```
joaquimrocha, line 107: As I tried to explain before, I think we can use this command later for migrating up too. So we should not go with a name that hints migrating down, which both this + the previous version do.
I think something like `migrate-db-to 0013` is more adequate but could also be `set-db-migration 0013`.

BTW, if you touch this line, please also make `db` -> `DB`.

joaquimrocha (changes requested):
Left a comment about the command's name. Also, I don't think the docs were updated.

joaquimrocha (approved):

PR #608, opened 2022-06-02
frontend: Get a high number of packages until we add real pagination

joaquimrocha:
Due to how the backend changed how the items can get fetched (making
the pagination mandatory), the list of packages gotten by default was
just 10.

We should fix this by applying a real pagination in the packages'
list, but until then, this patch requests a high number of packages
so the UX is a bit better until we fix the pagination in the UI.


illume (approved):
👍 thanks

PR #612, opened 2022-06-09
frontend: Prevent sending an empty version when fetching instances

joaquimrocha:
Sending an empty version is not allowed in the current API. That should be ideally prevented in the backend, but until then, this
patch removes an empty version from the query params in the respective API call.

@yolossn , I noticed that we could use an `allowEmptyValue` param for the version param and possibly others too, but the docs in swagger are not clear whether that option is going away: https://swagger.io/specification/

### DIFF
--- a/frontend/src/api/API.ts
+++ b/frontend/src/api/API.ts
@@ -195,12 +195,19 @@ export default class API {
   static getInstances(
     applicationID: string,
     groupID: string,
-    queryOptions = {}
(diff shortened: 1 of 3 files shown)


joaquimrocha:
I have added an unrelated commit because we need to get the threshold higher for tests to pass but it's complicated (or adequate) to add the stories for `API.ts`.

illume (approved):
👍

PR #613, opened 2022-06-09
Do not consider a slash for the endpoint secret

joaquimrocha:
This is a quick PR to try to fix the issue where we can only access the update endpoint using a slash when that's being used as the endpoint secret suffix.

This PR is however incomplete: it should prevent the use of the / earlier, by letting the user know it's not allowed, and it's missing tests.
Maybe @yolossn can help complete it.

### DIFF
--- a/backend/pkg/server/server.go
+++ b/backend/pkg/server/server.go
@@ -82,10 +82,17 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 	}
 
 	// setup middlewares
-	if conf.APIEndpointSuffix != "" {
-		e.Pre(custommiddleware.OmahaSecret(conf.APIEndpointSuffix))
-	}
 	e.Pre(middleware.RemoveTrailingSlash())
+
+	// remove trailing slash from the endpoint secret
+	endpointSuffix := strings.TrimSuffix(conf.APIEndpointSuffix, "/")
+	if endpointSuffix != "" {
+		// if endpoint secret doesn't start with slash prepend it
+		if !strings.HasPrefix(endpointSuffix, "/") {
+			endpointSuffix = fmt.Sprintf("/%s", endpointSuffix)
+		}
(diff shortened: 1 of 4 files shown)


backend/pkg/server/server.go
```diff
@@ -82,10 +82,10 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 	}
 
 	// setup middlewares
-	if conf.APIEndpointSuffix != "" {
+	e.Pre(middleware.RemoveTrailingSlash())
+	if conf.APIEndpointSuffix != "" && conf.APIEndpointSuffix != "/" {
```
joaquimrocha, line 86: I think this is unfortunately not enough. Have you tested with a secret that ends with a trailing slash but has more to it? E.g. `/mysecret/`. In this case, attempts to use `/mysecret` should fail (unless I misunderstood the logic in the custommiddleware.OmahaSecret)

The right solution IMO should be to edit the mentioned secret checker function and consider them to be equal if the discarded secret prefix is `""` or `"/"`.
I'd add more cases to the tests, to cover what I mentioned + cases like `no-pre-slash-secret`, `/using-two/trailing-slices//`, etc.

joaquimrocha (commented):
Thanks for updating this PR @yolossn . I think it may be missing the case I mentioned in the comment (but I haven't tested it).

backend/pkg/server/server.go
yolossn: Good catch, Since the logic of `custommiddleware.OmahaSecret` is to check if the extra string after removing the  "/v1/update" prefix matches the provided secret, I added checks to ensure that the API secret begins with a '/' and the trailing '/' in the secret is removed to accept both "/v1/update/secret" and "/v1/update/secret/" as valid.  Also added the mentioned test cases, PTAL.

joaquimrocha (commented):
LGTM.

PR #617, opened 2022-08-08
frontend: Fix selecting a package for a channel

joaquimrocha:
It was not getting correctly selected due to the "setSelected" function
not being applied to the correct value in the AutoCompletePicker.


ashu8912 (approved):
Tested on local, LGTM!!

illume (approved):
👍

PR #620, opened 2022-09-01
frontend: Fix undefined appId breaking group edit

yolossn:
## How to use

1. Click on an application.
2. Click on a group.
3. Editing the group should work.

### DIFF
--- a/frontend/src/components/Groups/GroupEditDialog/GroupEditDialog.tsx
+++ b/frontend/src/components/Groups/GroupEditDialog/GroupEditDialog.tsx
@@ -10,6 +10,7 @@ import Tabs from '@material-ui/core/Tabs';
 import { Form, Formik } from 'formik';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
 import * as Yup from 'yup';
 import { Group } from '../../../api/apiDataTypes';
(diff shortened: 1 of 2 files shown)


illume (approved):
👍

frontend/src/components/Groups/GroupEditDialog/GroupEditDialog.tsx
```diff
@@ -42,6 +43,8 @@ export default function GroupEditDialog(props: GroupEditDialogProps) {
   const [groupEditActiveTab, setGroupEditActiveTab] = React.useState(0);
   const { t } = useTranslation();
   const theme = useTheme();
+  const params = useParams<{ appID: string }>();
```
ashu8912, line 46: small nitpick, 
`const { appID } = useParams<{ appID: string }>();`

yolossn: When initiating edit from group page `useParams` returns an object with both `appID` and `groupID`.

ashu8912: yupp and this way we ask for the appID on the same line, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

PR #622, opened 2022-09-24
chart: Upgrade postgres dependency to latest major version

mkilchhofer:
# Upgrade to latest postgresql dependency

Bitnami released a new major version of the postgresql dependency. As major chart upgrades are sometimes tricky, let's upgrade that dependency before additional majors arrive.

## How to use

Nothing special.

## Testing done

Pushed the helm chart to my own helm repository (chartmuseum) and upgraded my nebraska instance.

### DIFF
--- a/charts/nebraska/values.yaml
+++ b/charts/nebraska/values.yaml
@@ -70,13 +70,13 @@ config:
   database:
     host: ""
     port: 5432
-    dbname: '{{ .Values.postgresql.postgresqlDatabase }}'
-    username: '{{ .Values.postgresql.postgresqlUsername }}'
+    dbname: '{{ .Values.postgresql.auth.database }}'
+    username: '{{ .Values.postgresql.auth.username }}'
     password: ""
     passwordExistingSecret:
       enabled: true
       name: '{{ .Release.Name }}-postgresql'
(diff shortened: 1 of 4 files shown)


joaquimrocha:
And thanks a lot @mkilchhofer

charts/nebraska/values.yaml
```diff
@@ -70,13 +70,13 @@ config:
   database:
     host: ""
     port: 5432
-    dbname: '{{ .Values.postgresql.postgresqlDatabase }}'
-    username: '{{ .Values.postgresql.postgresqlUsername }}'
+    dbname: '{{ .Values.postgresql.auth.database }}'
+    username: postgres
```
yolossn, line 74: Lets make the username configurable like before for compatibility purpose.

yolossn (commented):
Overall Looks good, just a small change.

charts/nebraska/values.yaml
mkilchhofer: Can have a look into it. New dependency chart is a bit different.

yolossn: Made the changes required for this and tested.

PR #627, opened 2022-11-24
backend/docker-compose: remove race condition and improve interactive dev experience

t-lo:
This change adds a dependency to the (nebraska) "server" service on the "postgres" service in docker-compose.test.yaml. Motivation of the change is to prevent a race condition where the "server" container starts faster than "postgres", fails to connect to the database, and shuts down.

Additionally, an optional persistent storage for postgres is added. DB contents can be persisted across docker-compose runs to aid manual, interactive testing. Usage instructions have been added to the YAML files as comments. The local directory used for persiting postgres data has been added to .gitignore to prevent accidental commits.

Tested locally with:
- `docker-compose --file docker-compose.test.yaml build`  (works as expected)
- `docker-compose --file docker-compose.test.yaml up` (works as expected)
- `docker-compose --file docker-compose.test.yaml --file docker-compose.pg-persist.yaml up`
   - create a new package, assign to a channel
   - `docker-compose --file docker-compose.test.yaml --file docker-compose.pg-persist.yaml down`
   - `docker-compose --file docker-compose.test.yaml --file docker-compose.pg-persist.yaml up`
   - Package still present in web interface

### DIFF
--- a/backend/docker-compose.test.yaml
+++ b/backend/docker-compose.test.yaml
@@ -1,3 +1,15 @@
+# Docker-compose file for interactive and for automated testing.
+#
+# Build docker image from local sources:
+#   docker-compose --file docker-compose.test.yaml build
+#
+# Run:
+#   docker-compose --file docker-compose.test.yaml run
+# Run with persistent storage:
+#   docker-compose --file docker-compose.test.yaml --file docker-compose.test-db-persist.yaml up
+#
+# See docker-compose.pg-persist.yaml for more informatino on persistent storage.
+
 version: "3.9"
 
 services:
@@ -23,6 +35,10 @@ services:
       network: host
     ports:
       - "8002:8000"
+    # The "long form" depends_on silently fails in some scenarios (we've seen this with podman)
+    #  so the short form is added for compatibility.
+    depends_on:
+      - postgres
     depends_on:
       postgres:
         condition: service_healthy
--- a/backend/docker-compose.test.yaml
+++ b/backend/docker-compose.test.yaml
@@ -35,10 +35,8 @@ services:
       network: host
     ports:
       - "8002:8000"
-    # The "long form" depends_on silently fails in some scenarios (we've seen this with podman)
-    #  so the short form is added for compatibility.
-    depends_on:
-      - postgres
+    # Workaround for missing support for depends_on conditions in podman-compose:
+    restart: always
     depends_on:
       postgres:
         condition: service_healthy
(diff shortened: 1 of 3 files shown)


backend/docker-compose.test.yaml
```diff
@@ -21,6 +33,8 @@ services:
       context: ../
       dockerfile: Dockerfile
       network: host
+    depends_on:
+      - postgres
     ports:
       - "8002:8000"
     depends_on:
```
illume, line 40: There's already a depends_on postgres... or am I missing something?

illume (commented):
Cool. Thanks. 

I wonder if it makes sense to add a make target for the persistence. For documentation/ease of use.

backend/docker-compose.test.yaml
t-lo: Hmmm, interesting. We've seen this fail in at least one scenario resulting in the whole "depends" being silently ignored.
I've added a comment to clarify.

t-lo: (It broke in @pothos 's set-up, maybe Kai could tune in...)

pothos: Could it be related to the `service_health` condition, not sure… how about solving this race with an automatic restart of the nebraska container?

pothos: The port is not reachable in my case and the startup is <1s delayed, maybe the health check doesn't say whether the port is reachable?

pothos: Seems podman-compose [is the problem](https://github.com/containers/podman-compose/pull/453) because it started the container while the other one was still in state `(starting)`, I could try docker-compose as root on top of the Podman socket or we add `restart: always` which mitigates the problem for me.

backend/docker-compose.test.yaml
```diff
@@ -23,6 +35,10 @@ services:
       network: host
     ports:
       - "8002:8000"
+    # The "long form" depends_on silently fails in some scenarios (we've seen this with podman)
+    #  so the short form is added for compatibility.
+    depends_on:
+      - postgres
```
pothos, line 41:
```suggestion
    # Workaround for missing support for depends_on conditions in podman-compose:
    restart: always
```

t-lo: Why though? Short notation of `depends_on` works fine for podman, and having 2 `depends_on` statements in there does not create problems with either podman or docker. It feels like the cleaner approach to use dependencies?

t-lo: ping @pothos ^

pothos: I used the development version and should have the fix.

pothos (approved):

PR #628, opened 2022-11-24
Support Podman in docker-compose setup

pothos:
The "docker compose" subcommand to run docker-compose does not work when "docker" is a symlink to "podman" and "docker-compose" is a symlink to "podman-compose".
Use "docker-compose" instead of the subcommand to support running with Podman and give a hint in the docs on podman-compose.

## How to use

With the mentioned symlinks.

## Testing done

Tested `make test-service-up`

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -55,16 +55,16 @@ opened browser page should automatically refresh.
 
 ## Docker compose V2
 
-You may be missing the `docker compose` plugin, which is required to run some tests.
+You may be missing `docker-compose`, which is required to run some tests.
 
 ```bash
-docker compose version
+docker-compose version
 
 cd backend
 make check-backend-with-container
 ```
 
-If so, see https://docs.docker.com/compose/cli-command/#install-on-linux
+If so, see https://docs.docker.com/compose/install/other/ or https://github.com/containers/podman-compose (with setting up a `docker-compose` symlink)
 
 # Development Concepts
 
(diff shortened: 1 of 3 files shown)


docs/development.md
```diff
@@ -55,16 +55,16 @@ opened browser page should automatically refresh.
...
 ```bash
-docker compose version
+docker-compose version
 
 cd backend
 make check-backend-with-container
 ```
 
-If so, see https://docs.docker.com/compose/cli-command/#install-on-linux
+If so, see https://docs.docker.com/compose/cli-command/#install-on-linux or https://github.com/containers/podman-compose (with a symlink)
```
illume, line 67: I think we should then link to the docker-compose standalone install page. https://docs.docker.com/compose/install/other/

illume (changes requested):
👍thanks

As well as supporting podman...  I like that removing the compose plugin step as is done here makes it easier for people. I think `docker compose` hasn't really caught on, and docker-compose is still a bit more widespread in scripts.

IMHO the docs should now change to link to the docker-compose standalone docs rather than the sub command plugin docs https://docs.docker.com/compose/install/other/

t-lo:
For podman-compose compatibility please also take this into account: https://github.com/kinvolk/nebraska/pull/627/files#diff-91315b4cf15177fd76e06ff8c63ccad1721d8ec2d518a82252aef2bda6b95f49R38-R39

docs/development.md
pothos: Ok, done

yolossn (approved):
LGTM

PR #643, opened 2023-03-10
vbatts/backend update

vbatts:
# linters and golang updates

fix inconsistencies in reported linting, and differences in the versions of golang used.

I'm testing this locally with `go1.19.7` now, since the `backend/go.mod` says go1.19.
Though the github workflow was using go1.17 currently, but now is also go1.19.

This also aligns an issue I found where I saw errors by running `make` (i.e. `make all`) locally, but did not see them in `make ci` locally or on the github action. Now that is cleaned up.

## Testing done

```shell
cd backend
make all ci
```
and now the github actions output

Closes: #638


vbatts:
ok. Ready for review.

.github/workflows/backend.yml
```diff
@@ -31,5 +31,5 @@ jobs:
 
     - name: Build, Lint & Check
       run: |
-        make ci
+        make all ci
```
yolossn, line 34: The `all` recipe is for building and running both `frontend` and `backend`. Just `make ci` should do here.

yolossn (changes requested):

.github/workflows/backend.yml
vbatts: ah, this should `cd backend` to run the backend test. The reason I added this at all, is that I was seeing failures locally, that were not getting caught in the github action.

vbatts: oh, it is already in the backend directory: `working-directory: backend`
This is not running the frontend build.

yolossn: We have a separate action for the frontend.

vbatts: correct. This is not running any of the frontend make targets. It's running `make all ci` from within the `./backend/` directory

PR #647, opened 2023-03-27
Minimal frontend updates, drop support for old node versions (12.x, 14.x), new Ubuntu LTS on CI

illume:
# Some frontend updates

More updates could be done, but they would take longer and require more intensive testing. This would be an intermediate update of frontend dependencies.

- Drop node 12.x, and 14.x support to ease maintenance. Node 16+ supports the new lockfile version 2/3. Supporting node 14.x would be possible but would harder to keep dependencies up to date since much of the ecosystem has moved on.
- Npm audit fix was run. Some vulnerabilities remain listed, but less than before. From "79 vulnerabilities (3 low, 6 moderate, 55 high, 15 critical)" to "51 vulnerabilities (1 low, 1 moderate, 40 high, 9 critical)".
- Supporting node 18 requires either using an old openssl or upgrading some packages. Upgrading webpack via react-scripts update would be a larger upgrade requiring more changes. In order to reduce the amount of changes in this release and reduce the time required for testing. Node 18 isn't supported so far. Continuing to not support it is not a regression. Not supporting node 18 seems to be the better of the options for now, with future upgrades to support node 18 with more extensive changes.

## How to use

It should work as before.

## Testing done

- [x] locally confirmed tests pass
- [x] ci passes


yolossn (approved):
Thanks for quickly taking a look into this.

PR #656, opened 2023-06-27
Add instance statistics table

skoeva:
A by-product of the Omaha update protocol employed by Nebraska delivers statistics on version spread, but accumulating and presenting this data suffers from a number of limitations:
-	The chart only represents all nodes’ current OS versions. We currently do not have historic data of the nodes’ previous software versions, and we would like to track of this data.
-	While historic data can be calculated from update events, this is computationally intense and unfeasible in practice. We currently cannot calculate or store this historic data efficiently, and we would like to be able to have quick and easy access to this data.
-	To be able to back-fill the live Nebraska server using historic data available in CSV format, we need to use a third-party tool. We currently cannot accomplish this using our existing setup, and we would like to implement a mechanism to do this natively.

We implement the `instance_stats` table (based on instance fields from `instance_application`) to store data on current version spread in an efficient format optimized for querying. To accomplish this, there are five data points that we associate each instance with, serving as the main fields of this table:
-	Timestamp: Necessary for keeping track of when update checks are performed
-	Channel name: Necessary to identify an individual instance
-	Architecture: Necessary to identify an individual instance
-	Version: Necessary to identify an individual instance
-	Instance count: Necessary for grouping instances by the previous four fields

New entries to the table will be generated using a background job scheduled periodically at a specified interval.

### DIFF
--- a/backend/pkg/api/db/migrations/0019_add_instance_stats_table.sql
+++ b/backend/pkg/api/db/migrations/0019_add_instance_stats_table.sql
@@ -1,0 +1,14 @@
+-- +migrate Up
+
+create table if not exists instance_stats (
+    timestamp timestamptz not null,
+    channel_name varchar(25) not null,
+    arch varchar(7) not null,
+    version varchar(255) not null,
+    instances int not null check (instances >= 0),
+    unique(timestamp, channel_name, arch, version)
+);
+
+-- +migrate Down
+
+drop table if exists instance_stats;
--- a/backend/pkg/api/instances.go
+++ b/backend/pkg/api/instances.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/doug-martin/goqu/v9"
@@ -48,6 +49,7 @@ const (
 
 const (
 	validityInterval postgresDuration = "1 days"
+	defaultInterval  time.Duration    = 2 * time.Hour
 )
 
 // Instance represents an instance running one or more applications for which
@@ -106,6 +108,14 @@ type InstancesQueryParams struct {
 	SortOrder     string `json:"sort_order"`
 	SearchFilter  string `json:"search_filter"`
 	SearchValue   string `json:"search_value"`
+}
+
+type InstanceStats struct {
+	Timestamp   time.Time `db:"timestamp" json:"timestamp"`
+	ChannelName string    `db:"channel_name" json:"channel_name"`
+	Arch        string    `db:"arch" json:"arch"`
+	Version     string    `db:"version" json:"version"`
+	Instances   int       `db:"instances" json:"instances"`
 }
 
 type instanceFilterItem int
@@ -630,4 +640,157 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
 		Where(goqu.C("group_id").Eq(groupID)).
 		Order(goqu.C("created_ts").Desc()).
 		Limit(uint(limit))
+}
+
+// instanceStatsQuery returns a SelectDataset prepared to return all instances
+// that have been checked in during a given duration from a given time.
+func (api *API) instanceStatsQuery(t *time.Time, duration *time.Duration) *goqu.SelectDataset {
+	if t == nil {
+		now := time.Now()
+		t = &now
+	}
+
+	if duration == nil {
+		d := defaultInterval
+		duration = &d
+	}
+
+	// Helper function to convert duration to PostgreSQL interval string
+	durationToInterval := func(d time.Duration) string {
+		if d <= 0 {
+			d = time.Microsecond
+		}
+
+		parts := []string{}
+
+		hours := int(d.Hours())
+		if hours != 0 {
+			parts = append(parts, fmt.Sprintf("%d hours", hours))
+		}
+
+		remainder := d - time.Duration(hours)*time.Hour
+		minutes := int(remainder.Minutes())
+		if minutes != 0 {
+			parts = append(parts, fmt.Sprintf("%d minutes", minutes))
+		}
+
+		remainder -= time.Duration(minutes) * time.Minute
+		seconds := int(remainder.Seconds())
+		if seconds != 0 {
+			parts = append(parts, fmt.Sprintf("%d seconds", seconds))
+		}
+
+		remainder -= time.Duration(seconds) * time.Second
+		microseconds := remainder.Microseconds()
+		if microseconds != 0 {
+			parts = append(parts, fmt.Sprintf("%d microseconds", microseconds))
+		}
+
+		return strings.Join(parts, " ")
+	}
+
+	interval := durationToInterval(*duration)
+	timestamp := goqu.L("timestamp ?", goqu.V(t.Format("2006-01-02T15:04:05.999999Z07:00")))
+	timestampMinusDuration := goqu.L("timestamp ? - interval ?", goqu.V(t.Format("2006-01-02T15:04:05.999999Z07:00")), interval)
+
+	query := goqu.From(goqu.T("instance_application")).
+		Select(
+			timestamp,
+			goqu.T("channel").Col("name").As("channel_name"),
+			goqu.Case().
+				When(goqu.T("channel").Col("arch").Eq(1), "AMD64").
+				When(goqu.T("channel").Col("arch").Eq(2), "ARM").
+				Else("").
+				As("arch"),
+			goqu.C("version").As("version"),
+			goqu.COUNT("*").As("instances")).
+		Join(goqu.T("groups"), goqu.On(goqu.C("group_id").Eq(goqu.T("groups").Col("id")))).
+		Join(goqu.T("channel"), goqu.On(goqu.T("groups").Col("channel_id").Eq(goqu.T("channel").Col("id")))).
+		Where(
+			goqu.C("last_check_for_updates").Gt(timestampMinusDuration),
+			goqu.C("last_check_for_updates").Lte(timestamp)).
+		GroupBy(timestamp,
+			goqu.T("channel").Col("name"),
+			goqu.T("channel").Col("arch"),
+			goqu.C("version")).
+		Order(timestamp.Asc())
+
+	return query
+}
+
+// GetInstanceStats returns an InstanceStats table with all instances that have
+// been previously been checked in.
+func (api *API) GetInstanceStats() ([]InstanceStats, error) {
+	query, _, err := goqu.From("instance_stats").
+		Order(goqu.C("timestamp").Asc()).ToSQL()
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := api.db.Queryx(query)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var instances []InstanceStats
+	for rows.Next() {
+		var instance InstanceStats
+		err = rows.StructScan(&instance)
+		if err != nil {
+			return nil, err
+		}
+		instances = append(instances, instance)
+	}
+
+	return instances, nil
+}
+
+// GetInstanceStatsByTimestamp returns an InstanceStats array of instances matching a
+// given timestamp value, ordered by version.
+func (api *API) GetInstanceStatsByTimestamp(t time.Time) ([]InstanceStats, error) {
+	timestamp := goqu.L("timestamp ?", goqu.V(t.Format("2006-01-02T15:04:05.999999Z07:00")))
+
+	query, _, err := goqu.From("instance_stats").
+		Where(goqu.C("timestamp").Eq(timestamp)).
+		Order(goqu.C("version").Asc()).ToSQL()
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := api.db.Queryx(query)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var instances []InstanceStats
+	for rows.Next() {
+		var instance InstanceStats
+		err = rows.StructScan(&instance)
+		if err != nil {
+			return nil, err
+		}
+		instances = append(instances, instance)
+	}
+
+	return instances, nil
+}
+
+// updateInstanceStats updates the instance_stats table with instances checked
+// in during a given duration from a given time.
+func (api *API) updateInstanceStats(t *time.Time, duration *time.Duration) error {
+	insertQuery, _, err := goqu.Insert(goqu.T("instance_stats")).
+		Cols("timestamp", "channel_name", "arch", "version", "instances").
+		FromQuery(api.instanceStatsQuery(t, duration)).
+		ToSQL()
+	if err != nil {
+		return err
+	}
+
+	_, err = api.db.Exec(insertQuery)
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/backend/pkg/api/instances_test.go
+++ b/backend/pkg/api/instances_test.go
@@ -3,6 +3,7 @@ package api
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
@@ -293,4 +294,105 @@ func TestGetInstanceStatusHistory(t *testing.T) {
 	assert.Equal(t, history[2].Version, "1.0.1")
 	assert.Equal(t, history[3].Status, InstanceStatusUpdateGranted)
 	assert.Equal(t, history[3].Version, "1.0.1")
+}
+
+func TestUpdateInstanceStats(t *testing.T) {
+	a := newForTest(t)
+	defer a.Close()
+
+	instances, err := a.GetInstanceStats()
+	assert.NoError(t, err)
+	assert.Equal(t, 0, len(instances))
+
+	// First test case: Create tInstance1, tInstance2, and tInstance3; check tInstance1 twice; switch tInstance2 version
+	start := time.Now().UTC()
+
+	tTeam, _ := a.AddTeam(&Team{Name: "test_team"})
+	tApp, _ := a.AddApp(&Application{Name: "test_app", TeamID: tTeam.ID})
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID, Arch: ArchAMD64})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID), Arch: ArchAMD64})
+	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: false, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+	tInstance1, _ := a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	tInstance2, _ := a.RegisterInstance(uuid.New().String(), "", "10.0.0.2", "1.0.0", tApp.ID, tGroup.ID)
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.3", "1.0.1", tApp.ID, tGroup.ID)
+
+	_, err = a.GetUpdatePackage(tInstance1.ID, "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	assert.NoError(t, err)
+
+	_, err = a.GetUpdatePackage(tInstance2.ID, "", "10.0.0.2", "1.0.1", tApp.ID, tGroup.ID)
+	assert.NoError(t, err)
+
+	ts := time.Now().UTC()
+	elapsed := ts.Sub(start)
+
+	err = a.updateInstanceStats(&ts, &elapsed)
+	assert.NoError(t, err)
+
+	instances, err = a.GetInstanceStats()
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(instances))
+
+	instanceStats, err := a.GetInstanceStatsByTimestamp(ts)
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(instanceStats))
+	assert.Equal(t, "1.0.0", instanceStats[0].Version)
+	assert.Equal(t, 1, instanceStats[0].Instances)
+	assert.Equal(t, "1.0.1", instanceStats[1].Version)
+	assert.Equal(t, 2, instanceStats[1].Instances)
+
+	// Next test case: Switch tInstance1 and tInstance2 versions to workaround the 5-minutes-rate-limiting of the check-in time and add new instance
+	ts2 := time.Now().UTC()
+
+	_, err = a.GetUpdatePackage(tInstance1.ID, "", "10.0.0.1", "1.0.3", tApp.ID, tGroup.ID)
+	assert.NoError(t, err)
+
+	_, err = a.GetUpdatePackage(tInstance2.ID, "", "10.0.0.2", "1.0.4", tApp.ID, tGroup.ID)
+	assert.NoError(t, err)
+
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.4", "1.0.5", tApp.ID, tGroup.ID)
+
+	ts3 := time.Now().UTC()
+	elapsed = ts3.Sub(ts2)
+
+	err = a.updateInstanceStats(&ts3, &elapsed)
+	assert.NoError(t, err)
+
+	instances, err = a.GetInstanceStats()
+	assert.NoError(t, err)
+	assert.Equal(t, 5, len(instances))
+
+	instanceStats, err = a.GetInstanceStatsByTimestamp(ts3)
+	assert.NoError(t, err)
+	assert.Equal(t, 3, len(instanceStats))
+	assert.Equal(t, "1.0.3", instanceStats[0].Version)
+	assert.Equal(t, 1, instanceStats[0].Instances)
+	assert.Equal(t, "1.0.4", instanceStats[1].Version)
+	assert.Equal(t, 1, instanceStats[1].Instances)
+	assert.Equal(t, "1.0.5", instanceStats[2].Version)
+	assert.Equal(t, 1, instanceStats[2].Instances)
+}
+
+func TestUpdateInstanceStatsNoArch(t *testing.T) {
+	a := newForTest(t)
+	defer a.Close()
+
+	tTeam, _ := a.AddTeam(&Team{Name: "test_team"})
+	tApp, _ := a.AddApp(&Application{Name: "test_app", TeamID: tTeam.ID})
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID)})
+	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: false, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+
+	ts := time.Now().UTC()
+	// Use large duration to have some test coverage for durationToInterval
+	elapsed := 3*time.Hour + 45*time.Minute + 30*time.Second + 1000*time.Microsecond
+
+	err := a.updateInstanceStats(&ts, &elapsed)
+	assert.NoError(t, err)
+
+	instanceStats, err := a.GetInstanceStatsByTimestamp(ts)
+	assert.NoError(t, err)
+	assert.Equal(t, "", instanceStats[0].Arch)
+	assert.Equal(t, "1.0.0", instanceStats[0].Version)
+	assert.Equal(t, 1, instanceStats[0].Instances)
 }
--- a/backend/pkg/api/bindata.go
+++ b/backend/pkg/api/bindata.go
@@ -20,6 +20,7 @@
 // db/migrations/0016_add_version_breakdown_indexes.sql (734B)
 // db/migrations/0017_drop_unused_indexes.sql (297B)
 // db/migrations/0018_add_sha256_file_field.sql (173B)
+// db/migrations/0019_add_instance_stats_table.sql (361B)
 
 package api
 
@@ -487,6 +488,26 @@ func dbMigrations0018_add_sha256_file_fieldSql() (*asset, error) {
 	return a, nil
 }
 
+var _dbMigrations0019_add_instance_stats_tableSql = []byte("\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\xff\x6c\x90\xcd\x4a\xc6\x30\x10\x45\xf7\xf3\x14\xb3\x4c\x30\x05\x11\xc4\x45\xd1\x95\xaf\xe0\xba\x8c\x71\xb4\xc1\x64\x52\x93\x69\x15\x9f\x5e\x5a\xe9\x8f\xf5\xdb\x0d\x9c\x7b\x93\xc3\x6d\x1a\xbc\x4a\xe1\xad\x90\x32\x3e\x0d\x00\xbe\xf0\x7c\x2a\x3d\x47\xc6\xf0\x8a\x92\x15\xf9\x2b\x54\xad\x18\xa4\x2a\x89\xe7\xae\x2a\x69\x45\x03\x88\x88\x1a\x12\x57\xa5\x34\xec\x97\x7e\x2f\x2d\x19\x63\x74\x4b\xc6\xf7\x24\xc2\xb1\x13\x4a\x8c\x13\x15\xdf\x53\x31\x37\xb7\xf6\x14\x9b\xc1\x86\xef\xce\x74\xe2\x52\x43\x96\x43\xff\xdf\x03\xab\xe0\xac\xaa\x1b\x43\xdf\xb3\x7f\x47\xb3\xd3\x87\x7b\xbc\xb6\xbf\x95\x51\xc2\xc7\xc8\x66\x73\x77\x7f\x64\xdd\xe2\xe4\xd6\xbf\x2d\xd8\x16\xe0\xb8\xd8\x63\xfe\x14\x80\x97\x92\x87\x7d\xb1\x8b\x6b\xb5\xf0\x13\x00\x00\xff\xff\xdc\x2b\xd0\xda\x69\x01\x00\x00")
+
+func dbMigrations0019_add_instance_stats_tableSqlBytes() ([]byte, error) {
+	return bindataRead(
+		_dbMigrations0019_add_instance_stats_tableSql,
+		"db/migrations/0019_add_instance_stats_table.sql",
+	)
+}
+
+func dbMigrations0019_add_instance_stats_tableSql() (*asset, error) {
+	bytes, err := dbMigrations0019_add_instance_stats_tableSqlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "db/migrations/0019_add_instance_stats_table.sql", size: 361, mode: os.FileMode(0644), modTime: time.Unix(1, 0)}
+	a := &asset{bytes: bytes, info: info, digest: [32]uint8{0x58, 0x78, 0xb3, 0x28, 0xce, 0xbb, 0xe7, 0x2, 0x5e, 0xaa, 0xdc, 0x2a, 0x88, 0xe6, 0x60, 0x28, 0x95, 0x87, 0x4a, 0x2d, 0x6e, 0x73, 0x9c, 0x5c, 0x11, 0x23, 0xb0, 0xbb, 0x3e, 0x60, 0x4c, 0x76}}
+	return a, nil
+}
+
 // Asset loads and returns the asset for the given name.
 // It returns an error if the asset could not be found or
 // could not be loaded.
@@ -598,6 +619,7 @@ var _bindata = map[string]func() (*asset, error){
 	"db/migrations/0016_add_version_breakdown_indexes.sql": dbMigrations0016_add_version_breakdown_indexesSql,
 	"db/migrations/0017_drop_unused_indexes.sql":           dbMigrations0017_drop_unused_indexesSql,
 	"db/migrations/0018_add_sha256_file_field.sql":         dbMigrations0018_add_sha256_file_fieldSql,
+	"db/migrations/0019_add_instance_stats_table.sql":      dbMigrations0019_add_instance_stats_tableSql,
 }
 
 // AssetDebug is true if the assets were built with the debug flag enabled.
@@ -667,6 +689,7 @@ var _bintree = &bintree{nil, map[string]*bintree{
 			"0016_add_version_breakdown_indexes.sql": {dbMigrations0016_add_version_breakdown_indexesSql, map[string]*bintree{}},
 			"0017_drop_unused_indexes.sql": {dbMigrations0017_drop_unused_indexesSql, map[string]*bintree{}},
 			"0018_add_sha256_file_field.sql": {dbMigrations0018_add_sha256_file_fieldSql, map[string]*bintree{}},
+			"0019_add_instance_stats_table.sql": {dbMigrations0019_add_instance_stats_tableSql, map[string]*bintree{}},
 		}},
 		"sample_data.sql": {dbSample_dataSql, map[string]*bintree{}},
 	}},


backend/pkg/api/instances.go
```diff
@@ -108,6 +108,14 @@ type InstancesQueryParams struct {
 	SearchValue   string `json:"search_value"`
 }
 
+type InstanceFact struct {
+	Timestamp   time.Time `db:"timestamp" json:"timestamp"`
+	ChannelName string    `db:"channel_name" json:"channel_name"`
+	Arch        string    `db:"arch" json:"arch"`
+	Version     string    `db:"version" json:"version"`
+	Instances   int       `db:"instances" json:"instances"`
```
pothos, line 116: Currently it would have to be `null.Int` but I think we can rather forbid null entries

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,70 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
 		Order(goqu.C("created_ts").Desc()).
 		Limit(uint(limit))
 }
+
+// instanceFactQuery returns a SelectDataset prepared to return all instances
+// that have been checked in the last hour.
+func (api *API) instanceFactQuery() *goqu.SelectDataset {
```
pothos, line 645: I think it would be nice to allow to pass a custom interval because it simplifies test writing.

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,70 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		Join(goqu.T("channel"), goqu.On(goqu.C("groups.channel_id").Eq(goqu.C("channel.id")))).
+		Join(goqu.T("package"), goqu.On(goqu.C("channel.package_id").Eq(goqu.C("package.id")))).
+		Where(
+			goqu.C("last_check_for_updates").Gte(goqu.L("now() - interval '1 hour'")),
+			goqu.C("last_check_for_updates").Lte(goqu.L("now()"))).
+		GroupBy("last_check_for_updates", "c.name", "c.arch", "version")
+}
+
+func (api *API) GetInstanceFact() ([]InstanceFact, error) {
+	query, _, err := api.instanceFactQuery().ToSQL()
```
pothos, line 666: Why is this used here?  The function shouldn't work on the `instance_application` table but just on the `instance_fact` table, but `instanceFactQuery` is for calculation of a new entry.

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,70 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		Join(goqu.T("groups"), goqu.On(goqu.C("group_id").Eq(goqu.C("groups.id")))).
+		Join(goqu.T("channel"), goqu.On(goqu.C("groups.channel_id").Eq(goqu.C("channel.id")))).
+		Join(goqu.T("package"), goqu.On(goqu.C("channel.package_id").Eq(goqu.C("package.id")))).
+		Where(
+			goqu.C("last_check_for_updates").Gte(goqu.L("now() - interval '1 hour'")),
+			goqu.C("last_check_for_updates").Lte(goqu.L("now()"))).
+		GroupBy("last_check_for_updates", "c.name", "c.arch", "version")
+}
+
+func (api *API) GetInstanceFact() ([]InstanceFact, error) {
```
pothos, line 665:
```suggestion
func (api *API) GetInstanceFacts() ([]InstanceFact, error) {
```
Let's use the plural as it returns many

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +294,31 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: true, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.2", "1.0.1", tApp.ID, tGroup.ID)
+
+	err := a.updateInstanceFact()
+	assert.NoError(t, err)
+
+	instances, err := a.GetInstanceFact()
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(instances))
```
pothos, line 315: After changing `GetInstanceFact` to not calculate new entries, this would be one "fact" entry.

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,70 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+			goqu.L("COUNT(*)").As("instances")).
+		Join(goqu.T("groups"), goqu.On(goqu.C("group_id").Eq(goqu.C("groups.id")))).
+		Join(goqu.T("channel"), goqu.On(goqu.C("groups.channel_id").Eq(goqu.C("channel.id")))).
+		Join(goqu.T("package"), goqu.On(goqu.C("channel.package_id").Eq(goqu.C("package.id")))).
+		Where(
+			goqu.C("last_check_for_updates").Gte(goqu.L("now() - interval '1 hour'")),
+			goqu.C("last_check_for_updates").Lte(goqu.L("now()"))).
+		GroupBy("last_check_for_updates", "c.name", "c.arch", "version")
+}
+
```
pothos, line 664: Let's also add a new function that only returns the last saved "instance fact" based on the timestamp.
This will be useful for the test and we can also make use of it for a quick textual summary besides the planned graph.

backend/pkg/api/instances.go
skoeva: Added `not null` to instances in the migration

backend/pkg/api/instances.go
skoeva: Should we pass it in as a string (as SQL allows, i.e. "1 day") or some kind of int value representing hours/days which would probably be more applicable?

backend/pkg/api/instances.go
skoeva: Ah I mixed up these functions while writing, my apologies for the confusion and thanks for pointing this out. You're right, we should not introduce `instance_application` at all when we are simply getting results from `instance_fact`

backend/pkg/api/db/migrations/0019_add_flat_table.sql
```diff
@@ -0,0 +1,17 @@
+-- +migrate Up
+
+create table instance_fact (
+    timestamp timestamptz,
+    channel_name text,
+    arch text,
+    version text,
+    instances int
```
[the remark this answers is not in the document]
pothos:
```suggestion
    instances int not null
```
Not null would be added like this (The value can still be `0` but it has to be set).

backend/pkg/api/instances.go
pothos: The most natural way from Go would be to use https://pkg.go.dev/time#Duration

backend/pkg/api/instances_test.go
pothos: Then, once we have a way to pass a duration to `updateInstanceFact` we can use `time.Now()` for each "checkpoint" to calculate the duration and it should allow us to run it multiple times with new `RegisterInstance` calls between the "checkpoints".

backend/pkg/api/instances.go
pothos: My comment is a bit wrong, we want all "instance facts" with the same timestamp (there are many).

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +294,23 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID), Arch: 1})
+	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: true, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.2", "1.0.1", tApp.ID, tGroup.ID)
+
+	err := a.updateInstanceFact()
+	assert.NoError(t, err)
+
+	instances, err := a.GetInstanceFacts()
```
pothos, line 313:
```suggestion
	instance_facts, err := a.GetInstanceFacts()
```

backend/pkg/api/db/migrations/0019_add_flat_table.sql
```diff
@@ -0,0 +1,17 @@
+-- +migrate Up
+
+create table instance_fact (
+    timestamp timestamptz,
+    channel_name varchar(10) not null,
+    arch varchar(5) not null,
+    version varchar(20) not null,
+    instances int not null
+);
```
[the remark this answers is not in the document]
pothos: We can also do this check from Go before creating a new entry but not so important for now.

backend/pkg/api/instances.go
pothos: We could also change the type to be >= 0

pothos:
The CI also has this error `Error: pkg/api/instances.go:668:28: cannot use query (variable of type *goqu.SelectDataset) as type string in argument to api.db.Query`

backend/Makefile
```diff
@@ -4,6 +4,7 @@ export GO111MODULE
 TAG := `git describe --tags --always`
 SHELL = /bin/bash
 DOCKER_CMD ?= "docker"
+DOCKER_COMPOSE ?= "docker compose"
```
pothos, line 7: We also need to make use of it in the two lines reported by `grep docker-compose Makefile`
```
36:	docker-compose -f ./docker-compose.test.yaml up -d --build
40:	docker-compose -f ./docker-compose.test.yaml down
```

pothos: Formatting is `$(CONTAINER_NEBRASKA_DB_URL)`

backend/pkg/api/db/migrations/0019_add_flat_table.sql
```diff
@@ -0,0 +1,18 @@
+-- +migrate Up
+
+create table instance_fact (
+    timestamp timestamptz,
+    channel_name varchar(10) not null,
+    arch varchar(5) not null,
+    version varchar(20) not null,
+    instances int not null
```
pothos, line 8:
```suggestion
    instances int not null,
```

backend/pkg/api/db/migrations/0019_add_flat_table.sql
```diff
@@ -0,0 +1,18 @@
+-- +migrate Up
+
+create table instance_fact (
+    timestamp timestamptz,
```
pothos, line 4: I guess we can also set `not null` here (didn't try)?

backend/pkg/api/instances.go
pothos: Suggested the check at the DB level

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,97 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		t = &now
+	}
+
+	if duration == nil {
+		d := time.Hour
+		duration = &d
+	}
+
+	// Convert duration to string
+	interval := fmt.Sprintf("%d milliseconds", int(duration.Milliseconds()))
```
pothos, line 690: Curious, do we need the `int` cast?

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,97 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		now := time.Now()
+		t = &now
+	}
+
+	if duration == nil {
+		d := time.Hour
+		duration = &d
+	}
+
+	// Convert duration to string
```
pothos, line 689:
```suggestion
```
Thanks for thinking about comments, I would say that this comment here is not really needed but in general more high level comments would be nice

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,97 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+	interval := fmt.Sprintf("%d milliseconds", int(duration.Milliseconds()))
+
+	query := goqu.From("instance_fact").
+		Select(goqu.L("*")).
+		Where(
+			goqu.C("timestamp").Gte(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
```
pothos, line 695:
```suggestion
			goqu.C("timestamp").Gt(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
```
I think we should prevent double counting, even if not very likely

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,97 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		if err != nil {
+			return nil, err
+		}
+		instances = append(instances, instance)
+	}
+
+	return instances, nil
+}
+
+func (api *API) updateInstanceFact(t time.Time, duration time.Duration) error {
```
pothos, line 717: These should be pointers to allow omission

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,97 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+				When(goqu.C("c.arch").Eq(1), "AMD64").
+				When(goqu.C("c.arch").Eq(2), "ARM").
+				As("arch"),
+			goqu.C("version"),
+			goqu.L("COUNT(*)").As("instances")).
+		Join(goqu.T("groups"), goqu.On(goqu.C("group_id").Eq(goqu.C("groups.id")))).
+		Join(goqu.T("channel"), goqu.On(goqu.C("groups.channel_id").Eq(goqu.C("channel.id")))).
+		Join(goqu.T("package"), goqu.On(goqu.C("channel.package_id").Eq(goqu.C("package.id")))).
+		Where(
+			goqu.C("last_check_for_updates").Gte(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
```
pothos, line 673:
```suggestion
			goqu.C("last_check_for_updates").Gt(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
```
I think we should prevent double counting, even if not very likely

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,97 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		now := time.Now()
+		t = &now
+	}
+
+	if duration == nil {
+		d := time.Hour
+		duration = &d
+	}
+
+	// Convert duration to string
```
pothos, line 656:
```suggestion
```
Thanks for thinking about comments, I would say that this comment here is not really needed but in general more high level comments would be nice

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,97 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		Join(goqu.T("groups"), goqu.On(goqu.C("group_id").Eq(goqu.C("groups.id")))).
+		Join(goqu.T("channel"), goqu.On(goqu.C("groups.channel_id").Eq(goqu.C("channel.id")))).
+		Join(goqu.T("package"), goqu.On(goqu.C("channel.package_id").Eq(goqu.C("package.id")))).
+		Where(
+			goqu.C("last_check_for_updates").Gte(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
+			goqu.C("last_check_for_updates").Lte(t.Format(time.RFC3339))).
+		GroupBy("last_check_for_updates", "c.name", "c.arch", "version")
+}
+
+func (api *API) GetInstanceFacts(t *time.Time, duration *time.Duration) ([]InstanceFact, error) {
```
pothos, line 678:
```suggestion
func (api *API) GetInstanceFacts(t time.Time) ([]InstanceFact, error) {
```
Actually, this function should rather use exact matching to only get the timestamp events with the given value, no duration or Gt/Lt comparisons.

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +296,36 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.2", "1.0.1", tApp.ID, tGroup.ID)
+
+	ts := time.Now()
+	elapsed := ts.Sub(start)
+
+	err := a.updateInstanceFact(ts, elapsed)
+	assert.NoError(t, err)
+
+	instanceFacts, err := a.GetInstanceFacts()
```
pothos, line 320:
```suggestion
	instanceFacts, err := a.GetInstanceFacts(ts)
```
We should only get the newly calculated entries here

backend/pkg/api/instances.go
skoeva: Is there ever a case where we can pass in `nil` for these parameters (given the `time.*` types)? Should we remove these `nil` checks?

backend/pkg/api/instances.go
skoeva: Based on your most recent comment on `GetInstanceFacts` (matching events with just time instead of time + duration), this new proposed function would be unnecessary, right?

pothos: There are two cases, one is that we want to retrieve the newly calculated statistics (by timestamp, for the test case), one is that we want to retrieve all statistics (for the graph).

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,92 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		GroupBy("last_check_for_updates", "c.name", "c.arch", "version")
+}
+
+// GetInstanceFacts returns an InstanceFact table of instances matching a
+// given timestamp value.
+func (api *API) GetInstanceFacts(t *time.Time) ([]InstanceFact, error) {
+	if t == nil {
+		now := time.Now()
+		t = &now
+	}
```
pothos, line 684: This doesn't make sense, it's almost impossible that there exist an event for the current time. Thus, the time should be mandatory and not a pointer.
The rest of the function here is good and behaves like the `GetInstanceFactsByTimestamp` we talked about, so let's give it that name.
We need another one `GetInstanceFacts` that has no parameter and returns all entries.

backend/pkg/api/instances.go
pothos: Yes, the explicit setting is only really needed for tests, and the normal production usage will be `nil` to use the defaults.

pothos:
```suggestion
func (api *API) updateInstanceFact(t *time.Time, duration *time.Duration) error {
```

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,112 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		return err
+	}
+
+	_, err = api.db.Exec(insertQuery)
+	if err != nil {
+		return err
+	}
+
+	// Run GetInstanceFact() as a sanity check
+	_, err = api.GetInstanceFacts(&t)
```
pothos, line 748: ByTimestamp or we can also remove this line since we have some test coverage now

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,108 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		Join(goqu.T("package"), goqu.On(goqu.C("channel.package_id").Eq(goqu.C("package.id")))).
+		Where(
+			goqu.C("last_check_for_updates").Gt(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
+			goqu.C("last_check_for_updates").Lte(t.Format(time.RFC3339))).
+		GroupBy("last_check_for_updates", "c.name", "c.arch", "version")
+}
+
+// GetInstanceFacts returns an InstanceFact table with all instances that have
+// been previously been checked in.
+func (api *API) GetInstanceFacts() ([]InstanceFact, error) {
```
pothos, line 680: For the consumer it would be great to get the elements always sorted by timestamp (can be done efficiently in the SQL query).

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,108 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+
+	return instances, nil
+}
+
+// updateInstanceFact updates the instance_fact table with instances checked
+// in during a given duration from a given time.
+func (api *API) updateInstanceFact(t *time.Time, duration *time.Duration) error {
+	insertQuery, _, err := goqu.Insert("instance_fact").
+		Cols("timestamp", "channel_name", "arch", "version", "instances").
+		FromQuery(api.instanceFactQuery(&t, &duration)).
```
pothos, line 735:
```suggestion
		FromQuery(api.instanceFactQuery(t, duration)).
```

pothos: This is for the one build failure

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,108 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		ToSQL()
+
+	if err != nil {
+		return err
+	}
+
+	_, err = api.db.Exec(insertQuery)
+	if err != nil {
+		return err
+	}
```
pothos, line 745:
```suggestion
	if err != nil {
		return err
	}
	return nil
```
Build failure otherwise because a return is needed

skoeva: Can't believe I missed this thanks!

pothos: Pointed out by https://github.com/flatcar/nebraska/actions/runs/5416821950/jobs/9847049099?pr=656

backend/pkg/api/db/migrations/0019_add_flat_table.sql
```diff
@@ -0,0 +1,18 @@
+-- +migrate Up
+
+create table instance_fact (
```
pothos, line 3:
```suggestion
create table if not exists instance_fact (
```
See `0015_add_package_file.sql` for reference

backend/pkg/api/db/migrations/0019_add_flat_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    timestamp timestamptz not null,
+    channel_name varchar(10) not null,
+    arch varchar(5) not null,
+    version varchar(20) not null,
+    instances int not null check (instances >= 0),
+    unique(timestamp, channel_name, arch, version)
+);
+
+create index idx_instance_fact on instance_fact(timestamp, channel_name, arch, version);
```
pothos, line 12:
```suggestion
create index if not exists idx_instance_fact on instance_fact(timestamp, channel_name, arch, version);
```

pothos: Do we need the index? If yes, I think we only would require the index for the timestamp but since that's mostly for tests, I think we could even do without the index.

skoeva: Would it help with more efficient querying? I suppose we could adjust for just timestamp? I'm not quite sure of the extent of its effect on performance

pothos: The index for timestamp only would help with `GetInstanceFactsByTimestamp` but since we only plan to use it for tests where we have < 20 entries in the DB I think we don't need it.

skoeva: Right, though if we query with `GetInstanceFacts` on all the fields, and let's say we have millions of entries, would an index on those fields not help speed up queries? The overarching goal here is to reduce query time as much as we can without significant costs, and it sounds good in theory but unsure in practice.

pothos: The index is for quicker finding of a row with a specific value. Since it returns all rows, I don't think the index makes a difference. In any case a benchmark can tell best whether we have problems and a tweak helps or not.
For now we are more concerned with increasing the test coverage.

backend/pkg/api/db/migrations/0019_add_flat_table.sql
pothos: Otherwise the tests will fail: https://github.com/flatcar/nebraska/actions/runs/5417059075/jobs/9847587765?pr=656

backend/pkg/api/db/migrations/0019_add_flat_table.sql
skoeva: Writing up test cases quickly though running into an issue, see teams for details

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,110 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+			goqu.C("last_check_for_updates").Lte(t.Format(time.RFC3339))).
+		GroupBy("last_check_for_updates", "c.name", "c.arch", "version").
+		Order(goqu.C("last_check_for_updates").Desc())
+}
+
+// GetInstanceFacts returns an InstanceFact table with all instances that have
+// been previously been checked in.
+func (api *API) GetInstanceFacts() ([]InstanceFact, error) {
+	query := goqu.From("instance_fact").
+		Select(goqu.L("*"))
```
pothos, line 683:
```suggestion
		Select(goqu.L("*")).Order(goqu.C("timestamp").Asc())
```
Something like that to order

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,110 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		d := time.Hour
+		duration = &d
+	}
+
+	// Convert duration to string
+	interval := fmt.Sprintf("%d milliseconds", int(duration.Milliseconds()))
+
+	return goqu.From("instance_application").
+		Select(
+			goqu.C("last_check_for_updates").As("timestamp"),
```
pothos, line 661: Wait, that looks wrong, when we use `last_check_for_updates` we get multiple InstanceFacts entries for a version but we actually want to sum up all instances to get a single entry for how many instances are on that version.
What we should store in the final entry is `t` for the current time where we calculated the InstanceFact.

backend/pkg/api/instances.go
skoeva: I've sorted in descending order for the previous function, would the most recent timestamps not be more relevant to include at the top of the list?

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,36 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(instanceFacts))
+
+	for _, instanceFact := range instanceFacts {
+		assert.NotNil(t, instanceFact.Timestamp)
+		assert.Equal(t, "test_channel", instanceFact.ChannelName)
+		assert.Equal(t, "AMD64", instanceFact.Arch)
+		assert.Contains(t, []string{"1.0.0", "1.0.1"}, instanceFact.Version)
+		assert.Equal(t, 1, instanceFact.Instances)
+	}
```
pothos, line 329: The test here show continue by letting the same instances we created with RegisterInstance check in again. Then we take another `ts2 := time.Now()` snapshot, run updateInstanceFact, GetInstanceFactsByTimestamp(ts2), and then check again in the for loop.
Then the test should let the 2 old instances check in again, and also a new instance. Same checks (yet now there should be 3 entries).
Finally only 1 new instance check in (same checks, we expect to only see 1 instance).

backend/pkg/api/instances.go
pothos: For going through the historic data I think it's better to start from the left of the x axis but yeah, it's just my preference and for me it would be more readable in a debug output to see oldest first and then how the statistics evolve related to when our test instances check in.

backend/pkg/api/instances_test.go
pothos: We should also make the first test case stronger: Currently there are two instances with different versions, but we should add one more instance for one of these versions, and we should let one of these instance check in twice (simulate an update request), and we should let one of these instance switch the version. With all this happening in the end we should see a report that says that for one version there are two instances, and for the other version there is one (assuming that we do the switch from the version which has two instances to the version which has one in the beginning, instead of ending up with all instances on the same version). This also covers the case that we want to count the instance that updates only with the most recent version it reported.

skoeva: Awesome, working on updating the tests now

skoeva: For the instance checking in twice, do you mean we should model that as an update request to the same current version? As opposed to an update request to a different version (which we allow for another instance)?

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,70 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	tApp, _ := a.AddApp(&Application{Name: "test_app", TeamID: tTeam.ID})
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID), Arch: ArchAMD64})
+	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: true, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+	tInstance1, _ := a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	tInstance2, _ := a.RegisterInstance(uuid.New().String(), "", "10.0.0.2", "1.0.0", tApp.ID, tGroup.ID)
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.3", "1.0.1", tApp.ID, tGroup.ID)
+
+	// check tInstance1 in twice
+	err := a.grantUpdate(tInstance1, "1.0.0")
```
pothos, line 315: Maybe the high level `_, err = a.GetUpdatePackage(tInstance.ID, "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)` would be better?

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,70 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	a := newForTest(t)
+	defer a.Close()
+
+	start := time.Now()
+
+	tTeam, _ := a.AddTeam(&Team{Name: "test_team"})
+	tApp, _ := a.AddApp(&Application{Name: "test_app", TeamID: tTeam.ID})
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID), Arch: ArchAMD64})
+	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: true, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
```
pothos, line 309:
```suggestion
	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: false, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
```
I think we should disable the safe mode because it can cause problems when we don't mock the full completion of an update cycle in the test

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,70 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+
+	ts := time.Now()
+	elapsed := ts.Sub(start)
+
+	update := a.updateInstanceFact(&ts, &elapsed)
+	assert.NoError(t, update)
+
+	instanceFacts, err := a.GetInstanceFactsByTimestamp(ts)
+	assert.NoError(t, err)
+	assert.Equal(t, 3, len(instanceFacts))
```
pothos, line 330: Why do you expect 3? From reading the above I would expect 2 entries because there are only two different versions in use at the end.

pothos: I think that's because the grouping by `last_check_for_updates` is still used, but we don't want to create a new instanceFact for each instance

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,110 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+
+// GetInstanceFactsByTimestamp returns an InstanceFact table of instances matching a
+// given timestamp value.
+func (api *API) GetInstanceFactsByTimestamp(t time.Time) ([]InstanceFact, error) {
+	query := goqu.From("instance_fact").
+		Select(goqu.L("*")).
+		Where(
+			goqu.C("timestamp").Eq(t.Format(time.RFC3339)))
+
+	rows, err := api.db.Query(query.ToSQL())
```
pothos, line 712: Idea: if you sort this by version the test becomes simpler because you get a stable order and can use instaceFacts[x] to refer to a certain version and check the count.

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,70 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+		assert.NotNil(t, instanceFact.Timestamp)
+		assert.Equal(t, "test_channel", instanceFact.ChannelName)
+		assert.Equal(t, "AMD64", instanceFact.Arch)
+		assert.Contains(t, []string{"1.0.0", "1.0.1"}, instanceFact.Version)
+	}
+
+	ts2 := time.Now()
+	elapsed = ts2.Sub(ts)
+
+	update = a.updateInstanceFact(&ts2, &elapsed)
```
pothos, line 351: Not needed, nothing was done since the last updateInstancFact.

backend/pkg/api/instances_test.go
pothos: As an update request but since it's just a request, it won't do an update.

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,70 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	tInstance1, _ := a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	tInstance2, _ := a.RegisterInstance(uuid.New().String(), "", "10.0.0.2", "1.0.0", tApp.ID, tGroup.ID)
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.3", "1.0.1", tApp.ID, tGroup.ID)
+
+	// check tInstance1 in twice
+	err := a.grantUpdate(tInstance1, "1.0.0")
+	assert.NoError(t, err)
+
+	// switch tInstance2 version
+	switched := a.grantUpdate(tInstance2, "1.0.1")
```
pothos, line 319: This does not overwrite the version, it only sets the version the update was granted for.
I think you can use `_, err = a.GetUpdatePackage(tInstance.ID, "", "10.0.0.1", "1.0.1", tApp.ID, tGroup.ID)` and it will let the instance check in with a newer version.

backend/pkg/api/instances.go
pothos: I think it's still not correct because you group by `last_check_for_updates` and this creates an instanceFact entry per instance instead of summing them up.

backend/pkg/api/instances_test.go
skoeva: Should we revise how we set up additional test cases then? https://github.com/flatcar/nebraska/pull/656#discussion_r1248287515

pothos: No, I just mean that this single call here only makes sense if there was some interaction with the server before, so we need to simulate the next test case here first.

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,70 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+func TestUpdateInstanceFact(t *testing.T) {
+	a := newForTest(t)
+	defer a.Close()
+
+	start := time.Now()
+
+	tTeam, _ := a.AddTeam(&Team{Name: "test_team"})
+	tApp, _ := a.AddApp(&Application{Name: "test_app", TeamID: tTeam.ID})
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID), Arch: ArchAMD64})
```
pothos, line 308: Please add error checking, currently it seems to return `nil` for `tChannel` causing the next line to crash

skoeva: I have, it's indicating other query issues one-by-one that I'm going through

backend/pkg/api/instances.go
```diff
@@ -656,31 +656,34 @@ func (api *API) instanceFactQuery(t *time.Time, duration *time.Duration) *goqu.S
...
-		Join(goqu.T("package"), goqu.On(goqu.C("channel.package_id").Eq(goqu.C("package.id")))).
+		Join(goqu.T("groups"), goqu.On(goqu.L("group_id").Eq(goqu.L("groups.id")))).
+		Join(goqu.T("channel"), goqu.On(goqu.L("groups.channel_id").Eq(goqu.L("channel.id")))).
+		Join(goqu.T("package"), goqu.On(goqu.L("channel.package_id").Eq(goqu.L("package.id")))).
 		Where(
-			goqu.C("last_check_for_updates").Gt(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
-			goqu.C("last_check_for_updates").Lte(t.Format(time.RFC3339))).
-		GroupBy("last_check_for_updates", "c.name", "c.arch", "version").
-		Order(goqu.C("last_check_for_updates").Asc())
+			timestamp.Gt(goqu.L("?::timestamp - ?::interval", t.Format(time.RFC3339), interval)),
```
pothos, line 675: `last_check_for_updates` should be used for the check, or?

skoeva: Right, thanks for picking up on that

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,102 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+
+	update = a.updateInstanceFact(&ts2, &elapsed)
+	assert.NoError(t, update)
+
+	instanceFacts, err = a.GetInstanceFactsByTimestamp(ts2)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.NoError(t, err)
+	assert.Equal(t, 3, len(instanceFacts))
```
pothos, line 388: This doesn't make sense at the moment because since line 353 nothing happened (no client checked in). So, 0 are expected, or? Also a good test case^^

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,102 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	// switch tInstance2 version
+	_, err = a.GetUpdatePackage(tInstance2.ID, "", "10.0.0.2", "1.0.1", tApp.ID, tGroup.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ts := time.Now()
+	elapsed := ts.Sub(start)
+
+	update := a.updateInstanceFact(&ts, &elapsed)
```
pothos, line 353:
```suggestion
	err = a.updateInstanceFact(&ts, &elapsed)
	if err != nil {
		t.Fatal(err)
	}
```
same below at the other call site

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,133 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+}
+
+// GetInstanceFacts returns an InstanceFact table with all instances that have
+// been previously been checked in.
+func (api *API) GetInstanceFacts() ([]InstanceFact, error) {
+	query := goqu.From("instance_fact").
+		Select(goqu.L("*")).Order(goqu.C("timestamp").Asc()).
+		Order(goqu.C("timestamp").Asc())
+
+	rows, err := api.db.Queryx(query.ToSQL())
```
pothos, line 702: _n the function below you to the ToSQL() call at a different place, I would decide for one way and stick to it

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,133 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+	// fmt.Printf("Generated SQL: %s\n", sql)
+
+	return query
+}
+
+// GetInstanceFacts returns an InstanceFact table with all instances that have
+// been previously been checked in.
+func (api *API) GetInstanceFacts() ([]InstanceFact, error) {
+	query := goqu.From("instance_fact").
+		Select(goqu.L("*")).Order(goqu.C("timestamp").Asc()).
```
pothos, line 699:
```suggestion
		Select(goqu.L("*")).
```
Already done in the line below

backend/pkg/api/instances.go
pothos: Did you revert it for debugging? The query still uses `GroupBy(goqu.L("last_check_for_updates")` which I think is not what we want and `goqu.C("last_check_for_updates").As("timestamp"),` is not correct either, we should use the given time from `timestamp` as timestamp for the generated entry.

pothos:
The CI won't build unless the bindata Go file is generated and committed.

backend/pkg/api/instances.go
pothos: I think with the above it would be
```suggestion
			timestamp.As("timestamp"),
```
and the plain SQL query text should look like `select '$timestamp' as timestamp …` where `$timestamp` is the Go variable t in the right format (it's a constant, all returned rows will have the same value).

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,108 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+func TestUpdateInstanceFact(t *testing.T) {
+	a := newForTest(t)
+	defer a.Close()
+
+	start := time.Now()
+
+	tTeam, err := a.AddTeam(&Team{Name: "test_team"})
+	if err != nil {
+		t.Fatal(err)
+	}
```
pothos, line 308: This can be more concise with `assert.NoError(t, err)`

pothos: and I think `t.Fatal` is normally not what we want there, it should only be used by `checkDB` to prevent the other tests from running

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,136 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+
+	query := goqu.From(goqu.T("instance_application")).
+		Select(
+			timestamp.As("timestamp"),
+			goqu.L("channel.name").As("channel_name"),
+			goqu.Case().
+				When(goqu.L("channel.arch").Eq(1), "AMD64").
+				When(goqu.L("channel.arch").Eq(2), "ARM").
+				As("arch"),
+			goqu.L("package.version").As("version"),
```
pothos, line 670: `instance_application` already has a `version` field that can be used without the below join for `package`.

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,136 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+// instanceFactQuery returns a SelectDataset prepared to return all instances
+// that have been checked in during a given duration from a given time.
+func (api *API) instanceFactQuery(t *time.Time, duration *time.Duration) *goqu.SelectDataset {
+	if t == nil {
+		now := time.Now()
+		t = &now
+	}
+
+	if duration == nil {
+		d := time.Hour
```
pothos, line 652: One hour could be too often because not all instances may check in during this time frame. Two hours or maybe even 6 or 12 could also be good defaults. Anyway, we should move the default out of this function as constant to have the background job use the same value easily.

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,136 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		duration = &d
+	}
```
pothos, line 654: As sanity check we could report an error if `duration.Microseconds()` is 0

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,136 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+
+	// Convert duration to string
+	interval := fmt.Sprintf("%d microseconds", int(duration.Microseconds()))
```
pothos, line 657: I think the conversion to string can't always use microseconds because Postgres doesn't like too large values. Therefore, we should do the work for Postgres here to cut it in the various parts like "2 hours 34 minutes 23 seconds 123 milliseconds 543 microseconds" or something like that.

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,132 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
 }
+
+// instanceFactQuery returns a SelectDataset prepared to return all instances
+// that have been checked in during a given duration from a given time.
+func (api *API) instanceFactQuery(t *time.Time, duration *time.Duration) *goqu.SelectDataset {
+	if t == nil {
+		now := time.Now()
+		t = &now
+	}
+
```
pothos, line 650:
```suggestion
	if duration == nil {
		d := time.Hour
		duration = &d
	}
```
Removed for debugging but should be added back (maybe comment it out instead?)

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,94 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+
+	// switch tInstance2 version
+	_, err = a.GetUpdatePackage(tInstance2.ID, "", "10.0.0.2", "1.0.1", tApp.ID, tGroup.ID)
+	assert.NoError(t, err)
+
+	ts := time.Now()
+	elapsed := ts.Sub(start)
+
+	if elapsed == nil {
+		elapsed = time.Hour
```
pothos, line 341: Doesn't compile (You meant .microseconds() returning 0?) and I think we rather want to do the sanity check in the function called, and either return and error or fall back to 1 microsecond or so.

backend/pkg/api/instances.go
```diff
@@ -631,3 +640,184 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+	return instances, nil
+}
+
+// GetInstanceFactsByTimestamp returns an InstanceFact table of instances matching a
+// given timestamp value.
+func (api *API) GetInstanceFactsByTimestamp(t time.Time) ([]InstanceFact, error) {
+	_, _, err := goqu.From("instance_fact").
+		Select(goqu.L("*")).
+		Where(goqu.C("timestamp").Eq(goqu.V(t.Format(time.RFC3339)))).
```
pothos, line 773: If not done yet, this needs `TIMESTAMP` somehow inserted into the query to type the value

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,129 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(instanceFacts))
+
+	// Count instances per version
+	counts := make(map[string]int)
+	for _, instanceFact := range instanceFacts {
+		counts[instanceFact.Version] += instanceFact.Instances
+	}
+
+	assert.Equal(t, 1, counts["1.0.0"])
```
pothos, line 348: Since it's ordered by version, can we simplify this to `assert.Equal(t, instanceFacts[0].Version, "1.0.0")` and then `assert.Equal(t, instanceFacts[0].Instances, 1)`?

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,129 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	instanceFacts, err = a.GetInstanceFactsByTimestamp(ts2)
+	assert.NoError(t, err)
+	assert.Equal(t, 0, len(instanceFacts))
+
+	for _, instanceFact := range instanceFacts {
+		assert.NotNil(t, instanceFact.Timestamp)
+		assert.Equal(t, "test_channel", instanceFact.ChannelName)
+		assert.Equal(t, "AMD64", instanceFact.Arch)
+		assert.Contains(t, []string{"1.0.0", "1.0.1"}, instanceFact.Version)
+	}
```
pothos, line 377: Not needed if it's expected to be empty

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,129 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+
+	// Let 2 old instances check in and add new instance
+	_, err = a.GetUpdatePackage(tInstance1.ID, "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+	assert.NoError(t, err)
+
+	_, err = a.GetUpdatePackage(tInstance2.ID, "", "10.0.0.2", "1.0.1", tApp.ID, tGroup.ID)
+	assert.NoError(t, err)
```
pothos, line 384: They will only get a new last-check-for-updates entry under this condition (where recent is 5 minutes ago):
```
 191   │         updateInstanceApplication = instance.Application.LastCheckForUpdates.UTC().Before(recent) ||
 192   │             instance.Application.Version != instanceVersion || instance.Application.GroupID.String != groupID
```
I suggest you add a comment here that you change the version of each instance to ensure that they get a new timestamp entry regardless when they last checked in.

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,103 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	assert.Equal(t, 2, counts["1.0.1"])
+
+	ts2 := time.Now().UTC()
+	elapsed = ts2.Sub(ts)
+
+	if elapsed <= 0 {
+		elapsed = time.Microsecond
+	}
+
+	// Switch tInstance1 and tInstance2 versions and add new instance
```
pothos, line 364:
```suggestion
	// Switch tInstance1 and tInstance2 versions to workaround the 5-minutes-rate-limiting of the check-in time and add new instance
```

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,103 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	instances, err = a.GetInstanceStats()
+	assert.NoError(t, err)
+	assert.Equal(t, 5, len(instances))
+
+	instanceStats, err = a.GetInstanceStatsByTimestamp(ts3)
+	assert.NoError(t, err)
+	assert.Equal(t, 3, len(instanceStats))
+
```
pothos, line 390:
```suggestion
assert.Equal(t, instanceStats[0].Version, "1.0.3")
assert.Equal(t, instanceStats[0].Instances, 1)
assert.Equal(t, instanceStats[1].Version, "1.0.4")
assert.Equal(t, instanceStats[1].Instances, 1)
assert.Equal(t, instanceStats[2].Version, "1.0.5")
assert.Equal(t, instanceStats[2].Instances, 1)
```

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,103 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.4", "1.0.5", tApp.ID, tGroup.ID)
+
+	ts3 := time.Now().UTC()
+	elapsed = ts3.Sub(ts2)
+
+	if elapsed <= 0 {
+		elapsed = time.Microsecond
+	}
```
pothos, line 378: This check/replacement can be moved to `updateInstanceStats`

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,129 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+		if err != nil {
+			return nil, err
+		}
+		instances = append(instances, instance)
+	}
+
+	return instances, nil
+}
+
+// GetInstanceStatsByTimestamp returns an InstanceStats table of instances matching a
```
pothos, line 719:
```suggestion
// GetInstanceStatsByTimestamp returns an InstanceStats array of instances matching a
```

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,129 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+// given timestamp value.
```
pothos, line 720:
```suggestion
// given timestamp value, ordered by version.
```

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,129 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+	timestamp := goqu.L("timestamp ?", goqu.V(t.Format("2006-01-02T15:04:05.999999Z07:00")))
+	timestampMinusDuration := goqu.L("timestamp ? - interval ?", goqu.V(t.Format("2006-01-02T15:04:05.999999Z07:00")), interval)
+
+	query := goqu.From(goqu.T("instance_application")).
+		Select(
+			timestamp,
+			goqu.T("channel").Col("name").As("channel_name"),
+			goqu.Case().
+				When(goqu.T("channel").Col("arch").Eq(1), "AMD64").
+				When(goqu.T("channel").Col("arch").Eq(2), "ARM").
```
pothos, line 673: This needs to cover the case where another arch value is present - e.g., you can use the empty string or `unknown`

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,103 @@ func TestGetInstanceStatusHistory(t *testing.T) {
 	assert.Equal(t, history[3].Status, InstanceStatusUpdateGranted)
 	assert.Equal(t, history[3].Version, "1.0.1")
 }
+
+func TestUpdateInstanceStats(t *testing.T) {
```
pothos, line 299: Suggestion, I would add one more test function, something like ` func TestUpdateInstanceStatsNoArch(t *testing.T) {` where the architecture isn't set for the package/channel. I can be simpler and only have one instance check in.

backend/pkg/api/db/migrations/0019_add_instance_stats_table.sql
```diff
@@ -0,0 +1,14 @@
+-- +migrate Up
+
+create table if not exists instance_stats (
+    timestamp timestamptz not null,
+    channel_name varchar(50) not null,
+    arch varchar(50) not null,
+    version varchar(50) not null,
```
pothos, line 7: Why 50? I think we can look at the data types where this comes from and use similar lengths.

skoeva: This was a placeholder because of errors on field length, will adjust this and update the bindata as long as it runs

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,141 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+// instanceStatsQuery returns a SelectDataset prepared to return all instances
+// that have been checked in during a given duration from a given time.
+func (api *API) instanceStatsQuery(t *time.Time, duration *time.Duration) *goqu.SelectDataset {
+	if t == nil {
+		now := time.Now()
+		t = &now
+	}
+
+	if duration == nil {
+		d := 2 * time.Hour
```
pothos, line 652: Please define a constant for this outside of the function (similar to `postgresDuration` in this file)

backend/pkg/api/instances.go
```diff
@@ -631,3 +639,141 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+	durationToInterval := func(d time.Duration) string {
+		str := fmt.Sprintf("%d microseconds", d.Microseconds())
+		if d.Microseconds() > 999999 {
+			str = fmt.Sprintf("%d seconds", int(d.Seconds()))
+		}
+		if d.Seconds() > 99 {
+			str = fmt.Sprintf("%d minutes", int(d.Minutes()))
+		}
+		if d.Minutes() > 99 {
+			str = fmt.Sprintf("%d hours", int(d.Hours()))
```
pothos, line 666: Casting to int means we loose the precision. It could be 1.5 hours and then it gets casted to 1 which is missing quite a lot of the interval. If postgres doesn't support fractions, I suggest you do some modulo/division magic to add `minutes`, `seconds`, `microseconds` substrings.

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,103 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+
+	instances, err = a.GetInstanceStats()
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(instances))
+
+	instanceStats, err := a.GetInstanceStatsByTimestamp(ts)
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(instanceStats))
+	assert.Equal(t, "1.0.0", instanceStats[0].Version)
+	assert.Equal(t, 1, instanceStats[0].Instances)
```
pothos, line 339:
```suggestion
	assert.Equal(t, "1.0.0", instanceStats[0].Version)
	assert.Equal(t, 1, instanceStats[0].Instances)
	assert.Equal(t, "1.0.1", instanceStats[1].Version)
	assert.Equal(t, 1, instanceStats[1].Instances)
```

backend/pkg/api/db/migrations/0019_add_instance_stats_table.sql
```diff
@@ -0,0 +1,14 @@
+-- +migrate Up
+
+create table if not exists instance_stats (
+    timestamp timestamptz not null,
+    channel_name varchar(12) not null,
```
pothos, line 5: In `migrations/0001_initial.sql` I see `name varchar(25) not null check (name <> ''),`, we should use the same length here.

backend/pkg/api/instances.go
```diff
@@ -631,3 +641,156 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
+
+	return instances, nil
+}
+
+// updateInstanceStats updates the instance_stats table with instances checked
+// in during a given duration from a given time.
+func (api *API) updateInstanceStats(t *time.Time, duration *time.Duration) error {
+	if *duration <= 0 {
+		*duration = time.Microsecond
+	}
```
pothos, line 781: I think this could be moved into instanceStatsQuery to the rest of the duration handling

backend/pkg/api/instances_test.go
```diff
@@ -294,3 +295,105 @@ func TestGetInstanceStatusHistory(t *testing.T) {
...
+
+	tTeam, _ := a.AddTeam(&Team{Name: "test_team"})
+	tApp, _ := a.AddApp(&Application{Name: "test_app", TeamID: tTeam.ID})
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID)})
+	tGroup, _ := a.AddGroup(&Group{Name: "group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: false, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+	_, _ = a.RegisterInstance(uuid.New().String(), "", "10.0.0.1", "1.0.0", tApp.ID, tGroup.ID)
+
+	ts := time.Now().UTC()
+	elapsed := ts.Sub(start)
```
pothos, line 389: The durationToInterval function doesn't have much test coverage, can we use a larger duration here to test that, e.g., something like
```suggestion
	// Use large duration to have some test coverage for durationToInterval
	elapsed := 3 * time.Hour + 45 * time.Minute + 30 * time.Second + 1000 * time.Microsecond)
```

pothos (approved):

PR #663, opened 2023-07-28
Add instance_stats background job

skoeva:
Following #656 where we add the `instance_stats` table, we now implement a background task that populates this table every hour, with the purpose of committing a snapshot of live instance counts to the new database table. This task, using a Go ticker, is started at service startup and runs for the whole lifetime of the Nebraska server until it is stopped.

### DIFF
--- a/backend/pkg/server/server.go
+++ b/backend/pkg/server/server.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"time"
 
 	oapimiddleware "github.com/deepmap/oapi-codegen/pkg/middleware"
 	"github.com/getkin/kin-openapi/openapi3filter"
@@ -47,6 +48,7 @@ var (
 )
 
 // New takes the config and db connection to create the server and returns it.
+// It also starts a background job to update instance stats periodically.
 func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 	// Setup Echo Server
 	e := echo.New()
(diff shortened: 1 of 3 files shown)


backend/pkg/server/server.go
```diff
@@ -112,6 +113,19 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 		return nil, fmt.Errorf("Error setting up handlers: %w", err)
 	}
 
+	// setup background job for updating instance stats
```
pothos, line 116: Can you move this to the start or the end of the function?
It would also good to update the function description (currently it is `// New takes the config and db connection to create the server and returns it` and does not tell that it starts a background task already)

backend/pkg/server/server.go
```diff
@@ -132,6 +134,20 @@ func New(conf *config.Config, db *db.API) (*echo.Echo, error) {
 		e.DefaultHTTPErrorHandler(err, c)
 	}
 	codegen.RegisterHandlers(e, handlers)
+
+	// setup background job for updating instance stats
+	go func() {
+		ticker := time.NewTicker(db.GetDefaultInterval())
```
pothos, line 140:
```suggestion
		err := db.UpdateInstanceStats(nil, nil)
			if err != nil {
				logger.Err(err).Msg("Error updating instance stats")
			}
		ticker := time.NewTicker(db.GetDefaultInterval())
```
It would be good to do a run on startup. The ticker will not directly issue an event.

pothos: Maybe we can add a check that this is only done initially if there is no "recent" entry.

pothos: By the way, I wondered why it didn't work with a local instance registering even after adding this initial run and it turned out that the default `time.Now` is not UTC. I've filed a PR: https://github.com/flatcar/nebraska/pull/664

pothos: > Maybe we can add a check that this is only done initially if there is no "recent" entry.

On the other hand it's nice for testing to directly see something after a restart. I guess it's ok to always run it once on start.

skoeva: I've been making changes locally to test, is there a formal way to test that this works as intended (i.e. proper entries, no duplicates, etc.)?

pothos: A test case is possible, but I think it's ok how it is now.

pothos (approved):

PR #665, opened 2023-08-02
Add metrics endpoints

skoeva:


backend/api/spec.yaml
```diff
@@ -1177,6 +1177,41 @@ paths:
...
+        - githubCookieAuth: []
+      responses:
+        "200":
+          description: Get instance stats success response
+          content:
+            application/x-ndjson:
+              schema:
+                $ref: "#/components/schemas/instanceStats"
+        "500":
+          description: Get instance stats error response
```
yolossn, line 1213: Why is this a response stream? And do we want to return (via stream) all the instance stats in this endpoint? I would suggest to use some form pagination instead of stream.

backend/pkg/api/instances.go
```diff
@@ -49,7 +49,7 @@ const (
 
 const (
 	validityInterval postgresDuration = "1 days"
-	defaultInterval  time.Duration    = time.Hour
+	defaultInterval  time.Duration    = 2 * time.Hour
```
yolossn, line 52: Is this change intentional here?

backend/pkg/api/instances.go
```diff
@@ -651,7 +651,7 @@ func (api *API) GetDefaultInterval() time.Duration {
 // that have been checked in during a given duration from a given time.
 func (api *API) instanceStatsQuery(t *time.Time, duration *time.Duration) *goqu.SelectDataset {
 	if t == nil {
-		now := time.Now()
+		now := time.Now().UTC()
```
yolossn, line 654: This change is done in #664, please rebase with main to avoid redoing this part.

yolossn (changes requested):
Left few comments, overall the prometheus metrics endpoint looks good.

backend/pkg/api/metrics.go
```diff
@@ -21,6 +21,12 @@ WHERE a.id = e.application_id AND e.event_type_id = et.id AND et.result = 0 AND
 GROUP BY app_name
 ORDER BY app_name
 `, ignoreFakeInstanceCondition("e.instance_id"))
+
+	latestInstanceStatsSQL string = `
+SELECT channel_name, version, arch, timestamp, instances AS instances_count
+FROM instance_stats
+WHERE timestamp = (SELECT MAX(timestamp) FROM instance_stats)
```
pothos, line 28: Instead of doing raw SQL here, could we add a new function `GetInstanceStatsLatest` in `pkg/api/instances.go` and use that? This would also allow to add a test for this new function.

backend/api/spec.yaml
skoeva: Yeah the goal is to output all instance stats here, and I agree with the pagination suggestion

backend/pkg/api/instances.go
skoeva: Yup! This was just an issue with rebasing, see #666

backend/pkg/api/metrics.go
skoeva: I had this logic initially in `pkg/api/instances.go` but decided to follow the convention above, though I agree with this suggestion

skoeva:
I've left an update [in my project doc](https://microsoft.sharepoint.com/:w:/t/Kinvolk/EcvyG-0M7KZIvHQG4c8RboQBKxdFQHnVB-i5G0wTLwhdbw?e=d4mZ08) for you guys to go through regarding my progress as I've moved away and the started the fall semester at my university

I'm unable to make changes with my current setup but happy to follow along with the work on Nebraska!

ErvinRacz:
Relates to https://github.com/flatcar/nebraska/issues/566

PR #699, opened 2023-12-13
fix: Exclude helm chart tags from appVersion calculation

mkilchhofer:
It seems that the versioning gets confused by the helm releases also produces git tags.

Before this change:
image

After:
image

## How to use

`-`

## Testing done

See screenshots above


pothos (approved):
Thanks for catching this

PR #713, opened 2024-01-18
build(deps): bump github.com/prometheus/client_golang from 1.14.0 to 1.18.0 in /backend

mkilchhofer:
Updated PR of:
- https://github.com/flatcar/nebraska/pull/709


pothos:
Thanks!

pothos:
https://github.com/flatcar/nebraska/pull/714 is also a good idea but I guess the best would be if you can squash the two commits in of this PR here because they belong together.

mkilchhofer:
> #714 is also a good idea but I guess the best would be if you can squash the two commits in of this PR here because they belong together.

Jup, I wanted to try what fits best, so you can easily see the small change needed.
PR is updated 👍

pothos (approved):

PR #735, opened 2024-03-06
backend: Update to Go 1.21

dongsupark:
As Go 1.19 reached EOL, update the default Go version to 1.21.

As `math/rand.Read` has been deprecated since Go 1.20, it is necessary to replace `math/rand` with `crypto/rand`, to fix warnings from golangci-lint.

As `rand.Seed` has been deprecated since Go 1.20, it is necessary to replace `rand.Seed(seed)` with `rand.New(NewSource(seed))`.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -20,10 +20,10 @@ jobs:
 
     steps:
 
-    - name: Set up Go 1.19
+    - name: Set up Go 1.21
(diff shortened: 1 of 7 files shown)


pothos (approved):

pothos:
Thanks!
I've directly merged it because I otherwise needed `go mod tidy` to build.

PR #772, opened 2024-04-25
.github: Add GitHub Action to release container image

pothos:
When creating a release tag we want the image to be built and published
    as flatcar/nebraska with the version tag and also as 'latest' image when
    the tag is not the special 'staging' tag.
    The nebraska-helm chart is also released on the same repo, and thus we
    need to filter for strict semver tags.

## How to use

Maybe we need to iterate over this a bit more in the course of the new release until it publishes things correctly.

### DIFF
--- a/.github/workflows/release-container.yml
+++ b/.github/workflows/release-container.yml
@@ -1,0 +1,49 @@
+name: Build Container Image
+on:
+  push:
+    tags:
+    - '[0-9]+.[0-9]+.[0-9]+'
+    - 'staging'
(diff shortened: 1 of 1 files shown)


pothos:
The idea is to now publish new tags in `flatcar/nebraska` and add a note to the `README.md` and the release notes that the container image location has changed.

dongsupark (approved):
Looks good.

Then first we should create the `nebraska` of `ghcr.io/flatcar`.

PR #773, opened 2024-04-25
charts: Update Helm chart to work with new container image

pothos:
The newly published container image does not include the Kinvolk
branding but uses the vanilla Nebraska branding. Therefore, the Helm
chart defaults need to be changed and we need to make a new Helm chart
release 1.2.0 that uses the Nebraska 2.9.0 image.


charts/nebraska/values.yaml
```diff
@@ -19,9 +19,6 @@ fullnameOverride: ""
 
 config:
   app:
-    logoPath: /nebraska/assets/kinvolk-logo.svg
-    title: Kinvolk Update Service
-    headerStyle: dark
```
mkilchhofer, line 24: Can we set them to `""`?

```yaml
    logoPath: ""
    title: ""
    headerStyle: ""
```

Otherwise the readme is not correct.

pothos: Thanks for the review.

I'm also happy that the CI test the chart - I opened the PR a bit too early because the image is still building.

pothos:
@tormath1 Can you do a new nebraska-helm-1.2.0 release with this?

dongsupark (approved):

Issue #599, opened 2022-04-24
Frontend: All Packages are not displayed


yolossn:
Not all packages are displayed in the UI.
No of packages returned by the API
![image](https://user-images.githubusercontent.com/22731013/164984327-08e73d0f-d972-4191-98a7-1f20013af0a5.png)
No of packages displayed in the UI
![image](https://user-images.githubusercontent.com/22731013/164984401-02f1aac9-4f14-4490-81d9-dc3006717aa9.png)

ervcz:
Solved by fixing pagination:
https://github.com/flatcar/nebraska/pull/1228

PR #851, opened 2024-10-02
Fixed count display for packages

anistark:
Fixes issue #599


ErvinRacz:
Hello @anistark 
Thank you for your pull request and for taking the time to work on this issue! I'm sorry that it took a long time to get back to you with a response.  Your fix is much appreciated! :)

After the first look at #599 I thought that the issue is only with the paginator itself which cannot display the total count as your solution also suggests, however I would challenge this understanding:

There seems to be two problems with the state management of the packages. The `List` component doesn't react to the creation of a new package properly even if pagination is not required.

Here you can see that I have posted a new package. However, the 8th package gets displayed only after a browser refresh, however the 8th package should already be fetched according to the network tab:
 
Screenshot 2025-02-20 112054

PR #852, opened 2024-10-02
fixed env load for db creds in api backend

anistark:
# Fixed dotenv loading for DB and other config

DB config was not getting picked from env file. Added [dotenv](https://github.com/joho/godotenv) to load in for backend api.

## Testing done

Connects successfully to db on build


backend/pkg/api/api.go
```diff
@@ -75,6 +77,12 @@ type API struct {
 
 // New creates a new API instance, creates the underlying db connection.
 func New(options ...func(*API) error) (*API, error) {
+	// Load .env
+	envErr := godotenv.Load()
```
ErvinRacz, line 81: I suggest moving this at the beginning of the `main` function in the `main` file, as the configuration parser method also relies on `env vars`.

ErvinRacz (changes requested):
Hello @anistark , thanks for your time to create this patch. It is always good to see others interest in the project :)

PR #898, opened 2025-01-14
add custom "Refresh-Token" HTTP header support for OIDC

justdan96:
# add custom "Refresh-Token" HTTP header support for OIDC

These changes enable a user to login to Nebraska with OIDC authentication, whilst also supplying the refresh token using the custom HTTP header "Refresh-Token". This naming convention conforms to https://www.rfc-editor.org/rfc/rfc6648.

## How to use

Gain an `id_token` and `refresh_token` from OIDC, supply these in the Bearer auth header and the new `Refresh-Token` header, respectively. 

(I will take a look at creating automated tests later)

## Testing done

Used `dex` with an LDAP connector and local user authentication, with group support added (https://github.com/mjudeikis/dex/tree/mjudeikis/groups.support). Tested OIDC authentication in browser through Dex with LDAP user, confirmed working. Tested OIDC authentication in `curl` through Dex with local user, confirmed working. The `curl` commands were similar to:
```
curl -L -X POST 'http://localhost:5556/dex/token' \
-H 'Authorization: Basic REDACTED' \
-H 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'grant_type=password' \
--data-urlencode 'scope=openid profile groups offline_access' \
--data-urlencode 'username=REDACTED' \
--data-urlencode 'password=REDACTED'

curl -vvv -H 'Authorization: Bearer REDACTED' -H 'Refresh-Token: REDACTED' -H 'Content-Type: application/json' -X GET http://localhost:3000/api/apps
```

### DIFF
--- a/backend/pkg/auth/oidc.go
+++ b/backend/pkg/auth/oidc.go
@@ -162,8 +162,12 @@ func (oa *oidcAuth) ValidateToken(c echo.Context) error {
 	refreshToken := session.Get("refresh_token")
 	if refreshToken == nil {
 		logger.Debug().Str("request_id", requestID).Msg("ValidateToken, Refresh token not found in session")
-		httpError(c, http.StatusUnauthorized)
-		return nil
+		refreshToken = refreshTokenFromRequest(c)
+		if refreshToken == "" {
+			logger.Debug().Str("request_id", requestID).Msg("ValidateToken, Refresh token not found in request")
+			httpError(c, http.StatusUnauthorized)
+			return nil
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


justdan96:
I think what you are saying makes sense - I guess for that the behaviour would change to load a refresh token, if there is one, ignore an empty refresh token and then for the validation we just check if the access token has expired and there is no refresh token available?

justdan96:
I changed the behaviour to how you described and everything seems to be working, I will close this PR and raise a new one with the different code.

PR #941, opened 2025-02-13
fix(frontend:e2e): update tests and snapshots

ErvinRacz:
# Update flaky tests and snapshots due

- Removed redundant test setup for application creation dialog in `global.setup.ts`.
- Updated screenshot masking in `instance.spec.ts` to include additional elements, improving test reliability.
- Updated snapshot images for `instance-history` and `instance-info` to reflect changes.

## Testing done

Executed the e2e locally a dozen times.


ErvinRacz:
> Ok, so if I understand correctly, we're masking datetime values on the screenshot as they are not predictable when running the tests?

Yes, if we make a screenshot of the pages where the status of the instances is timestamped now, tomorrow they will be different.

Also, it is not predictable how much time it takes for the test suit to perform certain actions, for example to create entities, therefore just masking these values seems to be a more reliable solution than trying to fix the datetime.

ErvinRacz:
> Ok, so if I understand correctly, we're masking datetime values on the screenshot as they are not predictable when running the tests?

Yes, if we make a screenshot of the pages where the status of the instances is timestamped now, tomorrow they will be different.

Also, it is not predictable how much time it takes for the test suit to perform certain actions, for example to create entities, therefore just masking these values seems to be a more reliable solution than trying to fix the datetime values.

PR #1087, opened 2025-06-21
Bump golangci-lint to v2

tylerauerbeck:
# Bump golangci-lint to v2

This bumps golangci-lint from the existing v1 syntax which is no longer being developed to the now supported v2. It updates the config, tooling and lints associated with the updated linters.

## How to use

Run lints as you would normally

## Testing done

```
cd backend && golangci-lint run --config=../golangci.yml -v ./...
```


tylerauerbeck:
@tormath1 This should be cleaned up now.

PR #1096, opened 2025-06-26
add link checking to CI

tylerauerbeck:
# Add link-checking to CI

To ensure that all links stay up to date and valid, this PR adds `lychee` to check that there are no dead links within the repository

## How to use

Open pull request, watch `lychee` run and check for dead links

## Testing done

Ran this against the existing repository and added the appropriate ignore rules and addressed existing dead link.

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]


tylerauerbeck:
Alright, this should be in good shape with all the extensions you mentioned. This will probably require a little tweaking as we play around with this a little more.

The only links that I needed to outright ignore or change were ones at code.google.com and x.com because they were intermittently returning 500 errors (code.google.com) or just always returning a 400 (x.com) but only from CI.

PR #655, opened 2023-06-23
helm chart: Add extra labels and annotations to pods and k8s resources

jescarri:
# Allow users to specify extra labels and annotations to pods and other k8s resources.

This change allows users to set extra labels and annotations to pods and k8s resources.

## How to use

In the values file:

````
extraLabels:
  foo: bar
 extraAnnotations:
  foo: bar

podLabels:
  another: test
podAnnotations:
   test: test
````

The above will label all k8s resources with foo: bar and annotate all k8s resources with foo:bar

In addition pods will be labelled with extraLabels + another: test and annotated with extraAnnotations + test: test

To test:

Deploy nebraska using the helm chart + add the above values to your values file.

## Testing done

Nebraska has been deployed in our clusters using the modified helm chart, we are able to add extra labels and annotations.

- [ *] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

PR #1097, opened 2025-06-26
add ability to specify extraAnnotations and extraLabels

### LINKED ISSUE #654: [RFE] Allow helm chart to add extraLabels and extraAnnotations to k8s resources
## Current situation

The helm chart does not allow adding extra labels to the resources it creates, in our clusters we have admission control and we require all resources to be labelled.

## Impact

Helm chart cannot be used if extra labels are required.

## Ideal future situation

Users can specify a set of extra labels and annotations.

## Implementation options

Add the option in the helm charts.

tylerauerbeck:
# Add ability to specify extra annotations and labels for helm chart

This adds the ability to specify a global set of extra annotations and labels that will be applied to each Kubernetes object deployed by the helm chart. It also allows for specifying extra annotations or labels specifically for the pods or pvc that is deployed as occasionally there are labels/annotations that need to be applied specifically just to these objects for various upstream integrations.

## How to use

With the following values files, you can run `helm template nebraska .` and you can see the labels present on the appropriate objects

```
extraAnnotations:
  hello: world
extraLabels:
  flatcar: linux

podLabels:
  pod: test

podAnnotations:
  pod: annotest

config:
  hostFlatcarPackages:
    persistence:
      enabled: true
      annotations:
        pvc: test
      labels:
        another: pvctest

serviceAccount:
  create: true
```

## Testing done

Ran `helm template` with the above updated files and verified that annotations and labels appear as expected

replaces #655
fixes #654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
-helm: add ability to specify extra annotations and labels and extra labels/annotations for pods and pvcs([nebraska#1097](https://github.com/flatcar/nebraska/pull/1097/files))
+- helm: add ability to specify extra annotations and labels for pods, PVCs, ingress, deployments, and other resources ([nebraska#1097](https://github.com/flatcar/nebraska/pull/1097/files))
 
 ### Changed
 


tylerauerbeck:
@ervcz This should be in the right order now. Let me know what you think.

PR #1100, opened 2025-06-28
add SECURITY.md file

tylerauerbeck:
# Add SECURITY.md file

To improve the OSSF score for this repository, we can add a `SECURITY.md` file which points to the instructions in the main `flatcar/flatcar` repository. This feels like a better options than copying the file directly and then it becoming out of date. Also leaves a note on how to reach out to the security team with any immediate concerns to avoid adding hurdles for reaching out to the security team (because sometimes one extra link is just too far).

## How to use

N/A. Just documentation.

## Testing done

Look at the markdown file to ensure it makes sense and renders appropriately.


tylerauerbeck:
@ervcz Should be better now. Let me know what you think!

sayanchowdhury (approved):

PR #1114, opened 2025-07-02
mod: remove explicit github.com/pkg/errors dependency

tormath1:
the 'wrap' of the error is supported in 'fmt' since a while now. Let's drop this explicit dependency.

It's still required from other dependencies:
```
$ go mod why github.com/pkg/errors
github.com/kinvolk/nebraska/backend/pkg/api
github.com/jackc/pgx/v4/stdlib
github.com/jackc/pgx/v4
github.com/jackc/pgx/v4.test
github.com/cockroachdb/apd
github.com/pkg/errors
```

---

Let's start to chase down useless dependency. This does not need a change log entry.

(related to: https://github.com/flatcar/nebraska/issues/1113)

### DIFF
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"net/url"
@@ -8,7 +9,6 @@ import (
 
 	"github.com/knadh/koanf"
 	"github.com/knadh/koanf/providers/basicflag"
-	"github.com/pkg/errors"
 
 	"github.com/kinvolk/nebraska/backend/pkg/random"
 )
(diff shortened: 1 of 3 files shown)


tylerauerbeck (approved):
Obviously no power here, but this LGTM 🎉 . Just made this bump in another project and these are the steps I took there as well.

backend/pkg/config/config.go
```diff
@@ -71,12 +71,12 @@ const (
...
 	if c.HostFlatcarPackages {
 		if c.FlatcarPackagesPath == "" {
-			return errors.New("Invalid Flatcar packages path. Please ensure you provide a valid path using -flatcar-packages-path")
+			return errors.New("invalid Flatcar packages path. Please ensure you provide a valid path using -flatcar-packages-path")
 		}
 
 		tmpFile, err := os.CreateTemp(c.FlatcarPackagesPath, "")
 		if err != nil {
-			return errors.New("Invalid Flatcar packages path: " + err.Error())
+			return errors.New("invalid Flatcar packages path: " + err.Error())
```
krnowak, line 79: Should this be
```suggestion
			return fmt.Errorf("invalid Flatcar packages path: %w", err)
```
?

PR #1140, opened 2025-07-15
Address React hooks dependency warnings

ervcz:
# Address React hooks dependency warnings and optimize caching

This PR addresses multiple React hooks dependency warnings across the frontend components by:
- Adding `eslint-disable-next-line react-hooks/exhaustive-deps` comments where exhaustive deps checks are intentionally bypassed

## How to use
1. Review the eslint-disable comments to verify they're justified
2. Verify no console warnings about missing dependencies remain


krnowak (commented):
Were these warnings just noise then? Maybe we could consider disabling the specific check in eslint config?

krnowak:
Got it, thanks for the explanation. I appreciate it.

PR #1143, opened 2025-07-16
fix: log when port is in use

tylerauerbeck:
# Log when Nebraska port is in use

At some point, we stopped logging when the Nebraska port is in use. This results in the Nebraska server just silently exiting when attempting to startup and something is already using that port. As a user, it's beneficial to know the reason why the server doesn't remain online.

## How to use

Run `make run-backend` and watch Nebraska startup (and remain online)
In a separate terminal, run `make run-backend` again.
You should now see a log that looks like:
```
{"level":"fatal","error":"listen tcp :8000: bind: address already in use","time":"2025-07-16T00:41:37-04:00","context":"main","message":"Failed to start server"}
```

## Testing done

`make ci`
`make run-backend`


tylerauerbeck:
✅ Should be in good shape now.

PR #1151, opened 2025-07-24
updater: remove kinvolk

tylerauerbeck:
# Remove last `kinvolk` references from updater

This removes the final references to old `kinvolk` packages and updates them to use the new `flatcar` repos

## How to use

`make ci`

## Testing done

`make ci`


tylerauerbeck:
Looks like `ExampleUpdater()` and `ExampleUpdater_withUpdateHandler()` needed to be updated to not return `error` as we were getting a compile error stating that these function `should return nothing`. Update them to just print out the errors as they currently stood, but maybe there's something else we'd rather do with them instead.

tylerauerbeck:
@ervcz Yepp, that was it. Sorry for the delay here. This should be cleaned up now!

PR #1195, opened 2025-09-03
feat: Implement multi-step updates with floor packages

ervcz:
full context: https://github.com/flatcar/Flatcar/discussions/1831, https://github.com/flatcar/Flatcar/issues/1185

To safely roll out Flatcar updates that require mandatory
intermediate steps—such as partition migrations or
system-level configuration changes—this change introduces
floor packages: channel-scoped checkpoints that all clients
must install before progressing to any newer version.

We initially considered using prerequisite chaining, where
each package declares a specific predecessor it depends on.
While technically feasible, this model creates a
semantic mismatch: it treats a global update policy
(e.g., “all clients must run v1.5 before going beyond”) as if
it were a local dependency of individual packages.
To prevent clients from bypassing critical steps, every newly
released package would need to inherit all historical
prerequisites—even if those steps have nothing to do with the
new package itself. This forces artificial metadata into
packages and obscures the actual intent: enforcing a consistent
update path, not expressing per-package dependencies.

By shifting to floor-based semantics, we decouple policy from
package identity. Floors are managed independently, apply
universally to all future updates, and make the update
progression rules explicit and easy to audit. This approach
is simpler to understand, safer to maintain, and avoids
polluting package metadata with inherited constraints that
don’t belong there.

Core Changes
- Add floor packages (mandatory intermediate versions) support for safe multi-step updates
- Store channel-specific floors in new channel_package_floors junction table
- Reuse last_update_version field for already-granted state tracking

API Updates
- Add floor management endpoints (mark/unmark/list floor packages)
- Add AddPackageWithMetadata for atomic package creation with complete metadata
- Prevent floor/blacklist conflicts with validation rules

Update Logic
- Regular clients progress sequentially through floors (one package at a time)
- Modern syncers with multi_manifest_ok receive all packages in single response
- Legacy syncers blocked with NoUpdate when floors exist
- Fix already-granted logic to check against last_update_version not channel target

Syncer Enhancements
- Support multi-manifest processing with target detection
- Add package integrity verification for existing packages
- Extract helper functions for maintainability
- Add cleanup for failed downloads

Omaha Protocol
- Support floor attributes (IsFloor, FloorReason, IsTarget) in manifests
- Identify syncers via InstallSource="scheduler"
- Block legacy syncers without multi-package capability

Safety & Constraints
- Enforce mutual exclusion between floor and blacklist status
- Protect channel targets from blacklisting
- Abort on floor marking failures (atomic operations)

Testing
- Add comprehensive test coverage for all floor scenarios
- Test backward compatibility and edge cases
- Verify safety rules and constraints

Dependencies
- Requires enhanced go-omaha library with multi-manifest support

## Testing done

See my two comments.
 
- [x] Update the go-omaha dependency [after PR was merged](https://github.com/flatcar/go-omaha/pull/23) 


pothos:
Thanks for the work!
Using the channel as main point makes it simpler to deal with the "nesting" we would otherwise have if it's a package property.
We should keep in mind though that this requires to be careful when introducing a new floor relation. It has to be done for all channels down to LTS but always only when the package gets promoted, in a short time window directly before the new package gets selected. Adding it too early would require all clients to go through the floor version even if not required. Forgetting/adding it to late means that clients will already update to the latest version without going through the floor version.

pothos:
The commit messages tell what was done but miss the reasoning on why the change was done. I find it useful to first capture the current state formulated as problem statement as first part of the commit message and the second part then tells what was done as a solution (and might contain info on other alternatives and why they are not taken).

Even reviewing a small change as the removal of the `switch err` block I am left a bit confused on whether this changes the behavior or not and if that was intentional.

pothos:
> IMO you no longer strictly need a very short time window, because clients will go up until the channel head. The floor packages will be considered only if the channel head has a greater version then them. So, the promotion of a channel can be considered as a switch that also activates floor packages at the same time.

I think we wouldn't directly mark a new release as floor. Rather, we would mark one of the latest possible old releases as a floor when this is required as precondition to update to the new release. So the automatic exclusion doesn't really help there.

pothos:
Anyway, not a big issue, and even if it was the "nested package" approach we would still have to coordinate release progressions because a new package gets created there and would also need this info set up.

pothos:
I'm still reading all these changes, it takes a while. Especially `backend/pkg/syncer/syncer.go` seems like it was entirely rewritten and it's hard to follow the diff.

pothos:
In general most changes look ok and now we'll probably need a bit more (non-author) testing to check for problems.

The part about download cleaning sounds like it's also good to have but this would have been an ideal candidate to live in its own follow-up PR.

backend/api/spec.yaml
```diff
@@ -705,6 +705,124 @@ paths:
...
+          application/json:
+            schema:
+              type: object
+              properties:
+                floor_reason:
+                  type: string
+                  nullable: true
+                  description: Optional reason for marking this package as a floor (e.g., "Security fix", "Critical update")
+                  maxLength: 500
+                  example: "Security fix for CVE-2024-1234"
```
pothos, line 791:
```suggestion
                  example: "Introduces btrfs support in update-engine to handle new updates after this version"
```

backend/api/spec.yaml
```diff
@@ -1475,6 +1636,39 @@ components:
...
+        channel_id:
+          type: string
+        package_id:
+          type: string
+        floor_reason:
+          type: string
+          nullable: true
+          description: Optional reason for marking this package as a floor
+          maxLength: 500
+          example: "Security fix for CVE-2024-1234"
```
pothos, line 1668:
```suggestion
          example: "Introduces needed filesystem support to handle new updates after this version"
```

PR #1327, opened 2026-03-11
chore(docs): document the PostgreSQL 14+ requirement

ervcz:
**CHANGELOG.md**
- Breaking: PostgreSQL 14+ now required (via pgx/v5)
- Breaking: cascaded Nebraska syncers must run this release or newer to consume channels with floor packages
- Adds missing PR links: `--ca-file` (#1370), regex hardening (#1222), instance stats query (#1356, Thilo Fromm)

**README.md** — adds Requirements section with PG14+

**charts/nebraska/README.md** — adds PG14+ warning under "Upgrade PostgreSQL", and fixes the postgresql image tag in the parameter table

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Security
 ### Added
+### Changed
+### Removed
+### Bugfixes
 
-- **Custom CA Certificate for TLS:** Added `--ca-file` flag to trust additional CA certificates for TLS verification (e.g., internal CA, Let's Encrypt staging). Applies to the OIDC provider client and the syncer. Supports multiple PEM-encoded certs, additive to system CAs. Also exposed as `config.caFile` in the Helm chart.
+## [4.0.0] - 04/05/2026
+
+### Breaking Changes
+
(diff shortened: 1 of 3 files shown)


ervcz:
Hello @tormath1, @krnowak 
Can you please double-check the changelog edit?

tormath1 (approved):

ervcz:
Rebased onto current `main` and reworked.

4.0.0 was cut from the security fork branch, so none of this content is in it — and the next release has no date yet, so the entries now sit under `[Unreleased]` with no version heading. The `## [4.0.0]` section this PR used to add is gone: #1619 already landed the real one on `main`.

One factual fix while rebasing: the entry blamed `lib/pq` for the PostgreSQL 14 floor, but `lib/pq` is a blank import here and never opens a connection — `dbDriver` is `"pgx"` and everything goes through `pgx/v5`. The requirement is real, just for a different reason, so the entry now cites pgx's support policy instead.

Also reworded the syncer note to "this release or newer" rather than naming a version.

Separately: `lib/pq` looks removable altogether. I checked that dropping the blank import builds, vets, and passes `go test -p 1 ./pkg/api/...` against PostgreSQL 17. Keeping that out of this docs PR — happy to open it separately.

@tormath1 sorry, the force-push dismissed your approval. Could you take another look? The version framing changed since May.

ervcz:
Rebased on main. Dropped the 4.0.0 section because #1619 (https://github.com/Flatcar/nebraska/issues/1619) already added the real one, so the notes now sit under Unreleased. I also fixed the reason for the PG14 requirement. I wrote that lib/pq caused it, but lib/pq is only a blank import here and never opens a connection. We connect with pgx/v5, and pgx is what needs PG14.
@tormath1 the force push dismissed your approval, sorry. Can you look again?

Side note: lib/pq looks unused and removable. I will do that in a separate PR.

PR #1356, opened 2026-04-09
backend/instances.go: fix instance stats query

t-lo:
The instance stats query contained a bug in which it only returned instances that have _last_ checked in within the given time window instead of all _active_ instances during the window.

As a result, instance count was skewed and much lower than the actual amount of active instances.

This change updates the query from:
```sql
  last_check_for_updates > (ts - window_size)
  AND last_check_for_updates < ts
```
to
```sql
  last_check_for_updates > (ts - window_size)
  AND instance_created_ts < ts
```

The change also updates the default window size from 2 hours to one day to better match our legacy instance count metrics.

### DIFF
--- a/backend/pkg/api/instances.go
+++ b/backend/pkg/api/instances.go
@@ -48,8 +48,8 @@ const (
 )
 
 const (
-	validityInterval postgresDuration = "1 days"
-	defaultInterval  time.Duration    = 2 * time.Hour
+	validityInterval     postgresDuration = "1 days"
+	defaultStatsInterval time.Duration    = 24 * time.Hour
 )
 
 // Instance represents an instance running one or more applications for which
@@ -646,11 +646,17 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
 
 // GetDefaultInterval returns the default interval used for instance stats queries.
 func (api *API) GetDefaultInterval() time.Duration {
-	return defaultInterval
+	return defaultStatsInterval
 }
 
-// instanceStatsQuery returns a SelectDataset prepared to return all instances
-// that have been checked in during a given duration from a given time.
+// instanceStatsQuery returns a SelectDataset to estimate the active fleet size at a given point in time.
+// It answers "how many instances were part of the active fleet on day X",
+// not "how many instances specifically checked in on day X".
+//
+// Since last_check_for_updates gets overwritten on every check-in, we cannot determine
+// whether an instance was active on a specific past day. Instead, we count instances that:
+//  1. existed at the time (created_ts <= timestamp)
+//  2. are still alive (last_check_for_updates > timestamp - duration)
 func (api *API) instanceStatsQuery(t *time.Time, duration *time.Duration) *goqu.SelectDataset {
 	if t == nil {
 		now := time.Now().UTC()
@@ -658,7 +664,7 @@ func (api *API) instanceStatsQuery(t *time.Time, duration *time.Duration) *goqu.
 	}
 
 	if duration == nil {
-		d := defaultInterval
+		d := defaultStatsInterval
 		duration = &d
 	}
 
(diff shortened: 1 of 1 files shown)


backend/pkg/api/instances.go
```diff
@@ -646,19 +646,24 @@ func (api *API) instanceStatusHistoryQuery(instanceID, appID, groupID string, li
...
 
-// instanceStatsQuery returns a SelectDataset prepared to return all instances
-// that have been checked in during a given duration from a given time.
+// instanceStatsQuery returns a SelectDataset prepared to return all active instances within a time period.
+//
+//	This is somewhat complicated by the fact that an instance created _before_ or _during_ our window
+//	can have checked in _after_ the window we're supposed to check. last_check_for_updates gets updated every
+//	time an instance checks for an update. Hence we filter for instances that:
+//	 1. have been created before timestamp 't'
+//	 2. have checked in _during_ or _after_ the time period
```
ervcz:
```suggestion
// instanceStatsQuery returns a SelectDataset to estimate the active fleet size at a given point in time.
// It answers "how many instances were part of the active fleet on day X",
// not "how many instances specifically checked in on day X".
//
// Since last_check_for_updates gets overwritten on every check-in, we cannot determine
// whether an instance was active on a specific past day. Instead, we count instances that:
//   1. existed at the time (created_ts <= timestamp)
//   2. are still alive (last_check_for_updates > timestamp - duration)
```

backend/pkg/api/instances.go
```diff
@@ -710,12 +715,13 @@ func (api *API) instanceStatsQuery(t *time.Time, duration *time.Duration) *goqu.
 				Else("").
 				As("arch"),
 			goqu.C("version").As("version"),
-			goqu.COUNT("*").As("instances")).
+			goqu.COUNT("*").As("instances")).Distinct().
```
ervcz, line 719: The distinct is on the column selection set. Should it be on the instance_id? I think it is not doing anything currently. The `group by` at the end should ensure each row has a unique combination of channel.name, channel.arch and instance_application.version. Which means each row should be unique unless there is something that I don't see... e.g. an instance assigned to multiple apps or something like that?

backend/pkg/api/instances.go
```diff
@@ -710,12 +715,13 @@ func (api *API) instanceStatsQuery(t *time.Time, duration *time.Duration) *goqu.
...
 		Join(goqu.T("groups"), goqu.On(goqu.C("group_id").Eq(goqu.T("groups").Col("id")))).
 		Join(goqu.T("channel"), goqu.On(goqu.T("groups").Col("channel_id").Eq(goqu.T("channel").Col("id")))).
+		Join(goqu.T("instance"), goqu.On(goqu.T("instance_application").Col("instance_id").Eq(goqu.T("instance").Col("id")))).
 		Where(
 			goqu.C("last_check_for_updates").Gt(timestampMinusDuration),
```
ervcz, line 724: We could consider filtering out "fake" instances here too. Example: 
https://github.com/flatcar/nebraska/blob/main/backend/pkg/api/applications.go#L401

backend/pkg/api/instances.go
t-lo: I don't think Postgres supports `distinct` queries of individual columns; the whole `SELECT` column set must be `distinct` (I might be wrong though). Re `GROUP BY`, you're correct in that it will reduce the lines to a single one; however, if for some weird reason we have two check-ins by the same instance id (programmatic error etc.) then `GROUP BY` will count it twice.

I realise `Distinct()` is protecting against a fringe case though and am willing to drop it if it's problematic.

backend/pkg/api/instances.go
t-lo: Hmmm, good point, we had it in the example queries for testing the stats. Will add.

ervcz (approved):

PR #1370, opened 2026-04-20
Add `--ca-file` flag for custom TLS trust

ervcz:
Nebraska only trusts system CAs. If your OIDC provider or update server uses a certificate from an internal or non-public CA, Nebraska refuses to connect. The only workaround today is modifying the container's system trust store, which is inconvenient.

This PR adds a `--ca-file` flag that points to a PEM file with one or more CA certificates. These CAs are added to the system trust store — nothing is replaced. The custom trust applies to both the OIDC client and the syncer. When the flag is omitted, behaviour is identical to before.

If the file is missing, unreadable, or contains no valid PEM data, Nebraska fails at startup with a clear error message.

The flag is also exposed in the Helm chart as `config.caFile`, and documented in the OIDC migration guide.

### DIFF
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -51,6 +51,7 @@ type Config struct {
 	OidcLogoutURL     string `koanf:"oidc-logout-url"`
 	OidcAudience      string `koanf:"oidc-audience"`
 	OidcUseUserInfo   bool   `koanf:"oidc-use-userinfo"`
+	OidcCAFile        string `koanf:"oidc-ca-file"`
 }
 
 const (
@@ -88,6 +89,11 @@ func (c *Config) Validate() error {
 	case "oidc":
 		if c.OidcClientID == "" || c.OidcIssuerURL == "" || c.OidcAdminRoles == "" || c.OidcViewerRoles == "" {
 			return errors.New("invalid OIDC configuration")
+		}
+		if c.OidcCAFile != "" {
+			if _, err := os.Stat(c.OidcCAFile); err != nil {
+				return fmt.Errorf("invalid oidc-ca-file: %w", err)
+			}
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


ervcz:
Makes [#1239](https://github.com/flatcar/nebraska/pull/1239) deprecated.

ervcz:
To be documented on the website by: https://github.com/flatcar/flatcar-website/pull/562

John15321 (approved):
LGTM, thank you!

backend/pkg/config/config.go
```diff
@@ -127,6 +136,7 @@ func Parse() (*Config, error) {
 	f.String("oidc-logout-url", "", "OIDC logout URL (optional fallback when end_session_endpoint is not available in discovery)")
 	f.String("oidc-audience", "", "OIDC audience parameter for the access token")
 	f.Bool("oidc-use-userinfo", false, "Use OIDC UserInfo endpoint for role extraction (for providers that don't include roles in access token)")
+	f.String("ca-file", "", "path to a PEM-encoded CA certificate file to trust for TLS verification (additive to system CAs, supports multiple certs in one file)")
```
John15321, line 139: Is this supposed to be different than `--oidc-ca-file` and https://lemon-wave-085522403-562.westeurope.1.azurestaticapps.net/docs/latest/nebraska/authorization/#:~:text=Encrypt%20staging)%2C%20pass-,%2D%2Dca%2Dfile,-%3D/path/to/ca?

ervcz: It is a good question @John15321. I renamed it with my second commit from `--oidc-ca-file` to `--ca-file`, because the certificate pool should actually be possible to be used by any HTTP clients and shouldn't be restricted only to the OIDC usecase. E.g. the syncer module also has an http client.

John15321: Nice!
Can you change the PR title as well? I think in the website doc PR its already correct right?

ervcz: done

John15321 (approved):
Amazing, thank you!

Issue #1375, opened 2026-04-29
RFC: Running Nebraska as a distributed service with PSQL logical replication


Moustafa-Moustafa:
## Status

We now have a design doc proposed in PR #1405. It's the most up-to-date source for the design. This issue still has the original proposal below, with a few sections updated to match the current direction.

## Problem

We want to be able to host Nebraska as a managed service and need to run multiple instances across different regions, each backed by its own PostgreSQL database, so that nodes can reach the nearest instance with low latency and we are able to scale.

Nebraska today assumes a single instance with a single database. Even Omaha update checks (POST /v1/update/) write telemetry to the database, so read replicas are not an option. Each region needs its own writable database.

The challenge is keeping publisher metadata (apps, channels, packages, groups) consistent across all regional databases while letting each region write telemetry locally.

## How Nebraska's schema already helps

Nebraska's tables split almost cleanly into two categories:

Admin tables (publisher metadata, must be consistent everywhere): `application`, `package`, `channel`, `groups`, `team`, `users`, `flatcar_action`, `package_channel_blacklist`, `package_file`

Runtime tables (per-region telemetry, stays local): `instance`, `instance_application`, `instance_status_history`, `event`, `activity`, `instance_stats`

Reference / seed tables (identical on every node, populated by migrations, not replicated): `event_type`

All foreign keys point from runtime tables to admin tables, never the other direction. This makes PostgreSQL logical replication a natural fit: replicate admin tables from a primary database to regional subscriber databases, while each region writes runtime data locally.

Two tables mix admin and runtime concerns and need to be split first: `groups` (admin policy fields plus runtime-mutated `rollout_in_progress`) and `activity` (admin-originated rows for channel package updates plus runtime-originated rows for rollout lifecycle and instance update results). See Proposal #1 and #2 below.

## Proposal

Control node: one Nebraska instance that accepts admin API writes and replicates publisher metadata to all edges.

Edge nodes: regional Nebraska instances that serve Omaha update checks and write telemetry locally. Admin API calls return 403. The syncer is disabled.

```
Control                          Edge (per-region)
┌──────────────────┐         ┌──────────────────┐
│ Nebraska Control │         │ Nebraska Edge    │
│ admin.Service    │         │ runtime.Service  │
│ runtime.Service  │         │ (no admin writes)│
└────────┬─────────┘         └────────┬─────────┘
         │                            │
    ┌────┴─────┐                 ┌────┴────┐
[...]

PR #1396, opened 2026-06-11
feat(db): make rollout state runtime-local for distributed Nebraska

Moustafa-Moustafa:
The `rollout_in_progress` indicator for a group is the only field on the admin-managed `groups` table that is written by the update-serving engine, not the admin. This is a step towards the distributed topology described in RFC #1375: admin state needs to flow one-way from the control node to each edge node, while edge state must stay local so each edge node records what it observes from its own clients.

This PR splits out the fields that edge nodes need to write or that make sense to configure independently per node, into an edge-local `group_local` table:

- `rollout_in_progress`: written by the engine, moves entirely off `groups`.
- `policy_updates_enabled`: needs a per-node override so the `safe_mode` auto-pause brake keeps working when edge nodes can no longer write to admin tables.
- the remaining policy fields (`safe_mode`, `office_hours`, `timezone`, `period_interval`, `max_updates_per_period`, `update_timeout`) added as nullable overrides in the same shape, since the cost is one column each and it lets each edge node tune its own behavior (regional office hours, rate limits, etc.) when the operator wants to.

The effective value is `COALESCE(override, default)`. Admin defaults on `groups` remain the source of truth unless a node has explicitly overridden them.

The proposed direction for the follow-up UI/API PR is:

- Default fields on `groups` are editable only on the control node.
- Override fields on `group_local` are editable on every node, but only ever affect the local node.
- The control node's UI gets a two-pane editor: one pane for the admin defaults, one pane for the local overrides.
- Edge nodes get a single pane for their local overrides, with a clear disclaimer that changes only affect this node.

This aligns with the planned package-level split between admin-owned and runtime-owned tables: by the time we restrict edge nodes from writing to admin tables, every field they need to write already lives on `group_local`.

Refs: #1375

# Separate the `groups` table

## How to use

No new behavior is introduced. Nebraska running in single-instance mode (currently the only mode) shouldn't be impacted at all. Admin disable still sets `policy_updates_enabled = false` on `groups` exactly as before. The `safe_mode` brake still trips on repeated failures and is still cleared by an admin re-enable; under the hood that brake state now lives on `group_local.policy_updates_enabled_override` instead of `groups.policy_updates_enabled`, but the JSON contract and effective values are unchanged.

## Testing done
- `cd backend && make all` — clean (codegen, generators, golangci-lint, build)
- `cd backend && make check-backend-with-container` — passes (full suite incl. `backend/test/api` integration and `backend/test/auth/oidc`)
- Smoke test on a local environment with both a single instance and a distributed topology of a control and an edge node.


Moustafa-Moustafa:
@ervcz, an alternative came to mind that might be a better solution here, basically a way to split the `groups` table that keeps safe_mode working in distributed mode too.

First, a small side note: I'm using "control" and "edge" terminology rather than "admin" and "runtime" as I think it captures the topology better, but that's just naming.

The idea: the policy fields on `groups` are things that affect how updates get served, which feels more relevant to edge nodes than control. So it makes sense that each edge should be able to control them independently, with the control node setting a default as the fallback when an edge hasn't overridden them.

Create those policy fields onto `group_state` (or whatever we end up calling that table) as nullable override columns. Control writes the default (replicates everywhere). Each edge can write a local override. The engine reads `COALESCE(override, default)`.

That keeps safe_mode working when the engine on an edge sees too many failed updates, it flips the local `policy_updates_enabled_override`. Recovery is the edge operator clearing the override. It also lets regions tune office hours, rate limits, etc. independently without affecting the rest of the fleet.

Single-instance behavior is unchanged (override columns stay NULL, `COALESCE` returns the default), and the JSON contract stays the same since the API just returns the effective value.

Curious if this direction resonates with you before we go further, or if you'd rather keep this PR as the minimal split and stick with disabling safe_mode in distributed mode.

Moustafa-Moustafa:
> @ervcz, an alternative came to mind that might be a better solution here, basically a way to split the `groups` table that keeps safe_mode working in distributed mode too.
> 
> First, a small side note: I'm using "control" and "edge" terminology rather than "admin" and "runtime" as I think it captures the topology better, but that's just naming.
> 
> The idea: the policy fields on `groups` are things that affect how updates get served, which feels more relevant to edge nodes than control. So it makes sense that each edge should be able to control them independently, with the control node setting a default as the fallback when an edge hasn't overridden them.
> 
> Create those policy fields onto `group_state` (or whatever we end up calling that table) as nullable override columns. Control writes the default (replicates everywhere). Each edge can write a local override. The engine reads `COALESCE(override, default)`.
> 
> That keeps safe_mode working when the engine on an edge sees too many failed updates, it flips the local `policy_updates_enabled_override`. Recovery is the edge operator clearing the override. It also lets regions tune office hours, rate limits, etc. independently without affecting the rest of the fleet.
> 
> Single-instance behavior is unchanged (override columns stay NULL, `COALESCE` returns the default), and the JSON contract stays the same since the API just returns the effective value.
> 
> Curious if this direction resonates with you before we go further, or if you'd rather keep this PR as the minimal split and stick with disabling safe_mode in distributed mode.

Quick follow-up on this, I went ahead and implemented the override columns in this PR and pushed it as a second commit. It turns out it's not really optional. Once we land the read/write split in a later PR, edge nodes won't be allowed to write to admin-owned columns on groups, and policy_updates_enabled is exactly the column safe-mode needs to flip. Without a per-node override column, writing to the field would cause a compilation error after the split. I had this covered in my PoC but the thought slipped my mind on the previous commit.

While I was in there, I went ahead and added override columns for the other group policy fields too since the cost is one nullable column each and it unlocks per-edge tuning (regional office hours, rate limits, etc.) without any extra plumbing. They stay NULL by default, so single-instance behavior is unchanged and COALESCE(override, default) collapses to the admin value.

backend/pkg/api/groups.go
```diff
@@ -518,14 +534,35 @@ func (api *API) setGroupRolloutInProgress(groupID string, inProgress bool) error
...
+			goqu.I("groups.track"),
+			goqu.I("group_local.rollout_in_progress"),
+			goqu.COALESCE(goqu.I("group_local.policy_updates_enabled_override"), goqu.I("groups.policy_updates_enabled")).As("policy_updates_enabled"),
+			goqu.COALESCE(goqu.I("group_local.policy_safe_mode_override"), goqu.I("groups.policy_safe_mode")).As("policy_safe_mode"),
+			goqu.COALESCE(goqu.I("group_local.policy_office_hours_override"), goqu.I("groups.policy_office_hours")).As("policy_office_hours"),
+			goqu.COALESCE(goqu.I("group_local.policy_timezone_override"), goqu.I("groups.policy_timezone")).As("policy_timezone"),
+			goqu.COALESCE(goqu.I("group_local.policy_period_interval_override"), goqu.I("groups.policy_period_interval")).As("policy_period_interval"),
+			goqu.COALESCE(goqu.I("group_local.policy_max_updates_per_period_override"), goqu.I("groups.policy_max_updates_per_period")).As("policy_max_updates_per_period"),
+			goqu.COALESCE(goqu.I("group_local.policy_update_timeout_override"), goqu.I("groups.policy_update_timeout")).As("policy_update_timeout"),
+		).
```
ervcz, line 577: non-blocking, but this closure may minimize the repetition with coalesce:

```go
        eff := func(name string) interface{} {
                return goqu.COALESCE(goqu.I("group_local."+name+"_override"), goqu.I("groups."+name)).As(name)
        }
  ```

  ```go
                Select(
                        goqu.I("groups.id"),
                        goqu.I("groups.name"),
                        ...
                        goqu.I("group_local.rollout_in_progress"),
                        eff("policy_updates_enabled"),
                        eff("policy_safe_mode"),
                        ...
                ).

**Review by ervcz — APPROVED** [review]

File: CHANGELOG.md [hunk 2]
```diff
@@ -21,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
   - Package list UI now updates immediately after blacklist changes
   - Channel edit dialog filters out blacklisted packages from selection
   - Floor package selection prevents choosing blacklisted packages with clear visual feedback
+- **Group rollout state moved to runtime-local table:** The `rollout_in_progress` flag on `groups` is now stored in a separate `group_state` table, in preparation for the distributed Nebraska topology described in [RFC #1375](https://github.com/flatcar/nebraska/issues/1375). The JSON contract is unchanged. ([#1396](https://github.com/flatcar/nebraska/pull/1396))
```
ervcz:
```suggestion
- **Group rollout state moved to runtime-local table:** The `rollout_in_progress` flag on `groups` is now stored in a separate `group_local` table, in preparation for the distributed Nebraska topology described in [RFC #1375](https://github.com/flatcar/nebraska/issues/1375). The JSON contract is unchanged. ([#1396](https://github.com/flatcar/nebraska/pull/1396))
```

backend/pkg/api/db/migrations/0023_add_group_local_table.sql
```diff
@@ -0,0 +1,52 @@
...
+    for each row execute function create_group_local_for_group();
+
+alter table groups enable always trigger groups_create_group_local;
+
+-- +migrate Down
+
+alter table groups add column rollout_in_progress boolean not null default false;
+update groups set rollout_in_progress = gl.rollout_in_progress
+    from group_local gl where groups.id = gl.group_id;
+
```
ervcz, line 60: Do we need to think about to preserve a tripped safe-mode brake (or any local updates-enabled override) so rollback doesn't silently re-enable updates? How about the other fields once the UI can write them?
 ```sql
update groups set policy_updates_enabled = gl.policy_updates_enabled_override
     from group_local gl
     where groups.id = gl.group_id
       and gl.policy_updates_enabled_override is not null;```

backend/pkg/api/db/migrations/0023_add_group_local_table.sql
ervcz, line 1: the index (name of the file) of the migration should be increased to 0023

ervcz (approved):

ervcz:
@Moustafa-Moustafa , can you please close/resolve the copilot-made comments? once they are addressed/closed, I will hit the merge button

backend/pkg/api/groups.go
```diff
@@ -518,14 +534,38 @@ func (api *API) setGroupRolloutInProgress(groupID string, inProgress bool) error
...
+// the node-local group_local sidecar to expose rollout_in_progress and the
+// effective policy values, where effective means COALESCE(override, default).
+// The INNER JOIN is safe because the AFTER INSERT trigger on groups guarantees
+// a matching group_local row.
 func (api *API) groupsQuery() *goqu.SelectDataset {
-	query := goqu.From("groups").Order(goqu.I("created_ts").Desc())
-
-	return query
+	eff := func(name string) interface{} {
+		return goqu.COALESCE(goqu.I("group_local."+name+"_override"), goqu.I("groups."+name)).As(name)
```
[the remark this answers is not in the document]
Moustafa-Moustafa: `groups` table has a 1:1 relation to the `group_local` table. We can't just hide a missing record in the `group_local` table by using a left join with default values.

ervcz (approved):

PR #1398, opened 2026-06-15
feat(db): split activity into admin/runtime tables for distributed Nebraska

Moustafa-Moustafa:
The activity table holds two kinds of rows: admin-originated events (channel package updates) and runtime-originated events (rollout lifecycle, instance update results). This is a step towards the distributed topology described in RFC #1375: admin state needs to flow one-way from the admin node to each runtime node, while runtime state must stay local so each runtime node records what it observed from its own clients.

This change splits the table into `admin_activity` (admin events) and `activity` (runtime events), and exposes `all_activity` as a `UNION ALL` view so existing readers are unchanged.

`activity.id` is converted from `serial` to `uuid` so the two tables can share the view without id collisions, and so each node in a future distributed deployment can mint ids independently without coordination. The conversion rewrites the table; existing numeric ids are replaced with fresh uuids. The Go field was already declared `string`, so the API contract is unchanged.

Refs: #1375

# Split the `activity` table

Each `activity` row is *either* admin-originated (channel package updated) *or* runtime-originated (rollout lifecycle, instance update result). With the table split, admin rows can be carried by the admin publication while runtime rows stay on each runtime node's local DB. Reads continue to go through one place (the new `all_activity` view), so the activity feed behaves identically.

## How to use

No new behavior is introduced. Nebraska running in single-instance mode (currently the only mode) is not impacted at all. The view returns the same set of rows the existing `activity` table would have returned before this change.

In a future distributed deployment, `admin_activity` is replicated as part of the admin publication while `activity` stays local to each node. The activity feed on the admin node shows only the admin events it generated; the feed on each runtime node shows the admin events replicated in plus its own local runtime events. No node sees runtime events from other nodes.

## Testing done
- `cd backend && make all`
- `cd backend && make check-backend-with-container` passes
- New `TestActivityRouting` asserts each writer lands in the correct underlying table, that the `all_activity` view returns both, and that `hasRecentActivity` works against the view with the uuid PK
- Smoke test on a local environment, including a two-Postgres-instance setup with a publication on `admin_activity` only, confirming that channel-package-update rows replicate to the subscriber and Omaha-driven rollout events stay local on each side


backend/pkg/api/db/migrations/0022_split_activity_table.sql
```diff
@@ -0,0 +1,64 @@
...
+	group_id uuid references groups (id) on delete cascade,
+	channel_id uuid references channel (id) on delete cascade
+);
+
+create index on admin_activity (application_id);
+create index on admin_activity (group_id);
+create index on admin_activity (channel_id);
+
+-- Move existing admin-originated (channel package updated) rows to the new
+-- table; the original uuids are preserved so external references remain valid.
```
ervcz: shouldn't this say original integer IDs?

backend/pkg/api/db/migrations/0022_split_activity_table.sql
```diff
@@ -0,0 +1,64 @@
...
+drop sequence if exists activity_id_seq;
+
+create table admin_activity (
+	id uuid primary key default uuid_generate_v4(),
+	created_ts timestamptz default current_timestamp not null,
+	class integer not null,
+	severity integer not null,
+	version varchar(255) not null check (version <> ''),
+	application_id uuid not null references application (id) on delete cascade,
+	group_id uuid references groups (id) on delete cascade,
```
ervcz, line 21: is group_id used by admin activities?

backend/pkg/api/db/migrations/0022_split_activity_table.sql
```diff
@@ -0,0 +1,64 @@
...
+                      application_id, group_id, channel_id)
+select created_ts, class, severity, version,
+       application_id, group_id, channel_id
+from admin_activity;
+
+drop table if exists admin_activity;
+
+alter table activity drop constraint activity_pkey;
+alter table activity drop column id;
+alter table activity add column id serial primary key;
```
ervcz, line 72: is channel_id used by the new activity table? can't we drop it?

backend/pkg/api/db/migrations/0022_split_activity_table.sql
Moustafa-Moustafa: > is group_id used by admin activities?

Yes, group_id is unused today. I considered keeping it because admin operations on groups are conceptually possible and may come back (e.g. a future control-side group-related activity). But I also understand your position that we should keep the schema minimal until justified. Adding it later is a simple migration but would require coordination, especially in the distributed deployments.

I see pros and cons for each approach, but both is fine with me, let me know what your think.

backend/pkg/api/db/migrations/0022_split_activity_table.sql
```diff
@@ -0,0 +1,72 @@
...
+	severity integer not null,
+	version varchar(255) not null check (version <> ''),
+	application_id uuid not null references application (id) on delete cascade,
+	group_id uuid references groups (id) on delete cascade,
+	channel_id uuid references channel (id) on delete cascade
+);
+
+create index on admin_activity (application_id);
+create index on admin_activity (group_id);
+create index on admin_activity (channel_id);
```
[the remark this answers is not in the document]
Moustafa-Moustafa: `hasRecentActivity` is now `hasRecentRuntimeActivity` and it's only using the `activity` table which already has the index. The other places where the view is used doesn't need or use the index and as such it's not needed for the `admin_activity` table.

backend/pkg/api/activity_test.go
```diff
@@ -89,3 +89,37 @@ func TestGetActivity(t *testing.T) {
...
+	tGroup, _ := a.AddGroup(&Group{Name: "group_routing", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: true, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+
+	_ = a.newGroupActivityEntry(activityRolloutStarted, activitySuccess, tVersion, tApp.ID, tGroup.ID)
+	_ = a.newChannelActivityEntry(activityChannelPackageUpdated, activityInfo, tVersion, tApp.ID, tChannel.ID)
+
+	var runtimeCount, adminCount, runtimeAdminLeak, adminRuntimeLeak int
+	_ = a.db.QueryRow("select count(*) from activity where application_id = $1", tApp.ID).Scan(&runtimeCount)
+	_ = a.db.QueryRow("select count(*) from admin_activity where application_id = $1", tApp.ID).Scan(&adminCount)
+	_ = a.db.QueryRow("select count(*) from activity where class = 6 and application_id = $1", tApp.ID).Scan(&runtimeAdminLeak)
+	_ = a.db.QueryRow("select count(*) from admin_activity where class <> 6 and application_id = $1", tApp.ID).Scan(&adminRuntimeLeak)
```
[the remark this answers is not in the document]
Moustafa-Moustafa: I'd rather keep the tests concise and consistent with other tests on the same file.

backend/pkg/api/activity_test.go
```diff
@@ -89,3 +89,37 @@ func TestGetActivity(t *testing.T) {
...
+
+	tVersion := "12.1.0"
+	tTeam, _ := a.AddTeam(&Team{Name: "test_team_routing"})
+	tApp, _ := a.AddApp(&Application{Name: "test_app_routing", TeamID: tTeam.ID})
+	tPkg, _ := a.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: tVersion, ApplicationID: tApp.ID})
+	tChannel, _ := a.AddChannel(&Channel{Name: "test_channel_routing", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID)})
+	tGroup, _ := a.AddGroup(&Group{Name: "group_routing", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: true, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+
+	_ = a.newGroupActivityEntry(activityRolloutStarted, activitySuccess, tVersion, tApp.ID, tGroup.ID)
+	_ = a.newChannelActivityEntry(activityChannelPackageUpdated, activityInfo, tVersion, tApp.ID, tChannel.ID)
```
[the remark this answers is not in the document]
Moustafa-Moustafa: I'd rather keep the tests concise and consistent with other tests on the same file.

backend/pkg/api/activity_test.go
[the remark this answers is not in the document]
Moustafa-Moustafa: I'd rather keep the tests concise and consistent with other tests on the same file.

backend/pkg/api/db/migrations/0022_split_activity_table.sql
ervcz: let's keep it in that case :)

ervcz (approved):

PR #1405, opened 2026-06-23
docs: proposed distributed Nebraska mode design doc

Moustafa-Moustafa:
Adds docs/distributed-mode.md, the upstream proposed design doc for running Nebraska as one control node + N edge nodes with PostgreSQL logical replication of admin tables.

Refs: #1375

# Proposed distributed Nebraska mode design doc

Adds `docs/distributed-mode.md`, the design doc consolidating the decisions from RFC #1375 for running Nebraska as one control node + N edge nodes with PostgreSQL logical replication of the admin tables. This is meant to be iterated on and updated based on the discussions and feedback.

## How to use

Doce only, review the design and add any comments.,

## Testing done

Not applicable.

### DIFF
--- a/docs/distributed-mode.md
+++ b/docs/distributed-mode.md
@@ -1,0 +1,212 @@
+# WIP: Distributed Nebraska — Design
+
+**Status:** Proposed (RFC [#1375](https://github.com/flatcar/nebraska/issues/1375), implementation in flight)
+**Audience:** Nebraska maintainers and contributors
+
+## 1. Problem
+
+Nebraska today assumes a single instance with a single Postgres database. Hosting it as a managed service - where instances are deployed across multiple regions so each fleet of Flatcar nodes can reach the nearest one with low latency - hits two walls:
+
+1. **Every node writes runtime data.** Omaha update checks insert into `instance`, `event`, `activity`, etc. Read replicas are not an option.
+2. **A single database is a single point of failure** for the entire fleet across all regions, and adds cross-region latency for every Omaha poll.
+
+The goal is to let Nebraska run as multiple regional nodes, each backed by its own writable Postgres, with publisher metadata (apps, channels, packages, groups) staying consistent across all of them.
+
+## 2. Requirements
+
+- Single-instance behaviour is preserved. An operator who does not opt in sees no change.
+- One control node and N edge nodes can be configured.
+- Admin metadata (apps, channels, packages, groups) is consistent across all edge nodes.
+- Each edge node serves Omaha update checks and records its own runtime data locally.
+- Admin API writes are accepted on exactly one node (the control node) and blocked on the rest.
+
+## 3. Non-goals
+
+- **Control-plane HA / failover.** Single-instance Nebraska already has the DB as a SPOF; HA-Postgres addresses it identically in both worlds. Not regressed, not solved here.
+- **Multi-primary writes.** Admin writes have a single logical target.
+- **Cross-region telemetry aggregation.** Each region's runtime data stays local. Fleet-wide views are an operator concern, not a Nebraska feature.
+- **Auto-pause-on-failure (`safe_mode`) coordinated across nodes.** Each node owns its own brake; see §5 for how this is expressed.
+
+## 4. Design
+
+### 4.1 Topology
+
+```
+        Publisher (admin API)
+                │
+         ┌──────▼──────┐
+         │ Control     │  admin writes accepted
+         │ Nebraska    │  ──┐ logical replication
+         │  + own PG   │    │ (admin tables only)
+         └─────────────┘    │
+                            ▼
+            ┌──────────────────────────┐
+            │ Edge Nebraska × N        │  admin writes return 403
+            │  + own regional PG each  │  Omaha checks + telemetry local
+            └──────────────────────────┘
+```
+
+One control node, N edge nodes. Each runs the same Nebraska binary, each owns its own Postgres. Publisher metadata flows control → edge via **PostgreSQL logical replication** of the admin tables. Runtime tables stay local per edge.
+
+### 4.2 Table model
+
+Nebraska's schema already splits cleanly into admin and runtime tables, with FKs pointing only from runtime to admin. Two tables today mix the two concerns and need to be split before replication is safe:
+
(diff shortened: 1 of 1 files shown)


docs/distributed-mode.md
```diff
@@ -0,0 +1,212 @@
+# Distributed Nebraska — Design
```
ervcz:
```suggestion
# WIP: Distributed Nebraska — Design
```

docs/distributed-mode.md
```diff
@@ -0,0 +1,212 @@
...
+
+- **`groups`** mixes admin-owned policy fields with the runtime-mutated `rollout_in_progress` flag, and `policy_*` fields the runtime needs to be able to override locally (safe-mode brake). → Splits into `groups` (admin defaults, replicated) + node-local `group_local` sidecar. `group_local` holds `rollout_in_progress` plus one nullable `policy_*_override` column for each admin policy field. Reads return `COALESCE(group_local.policy_X_override, groups.policy_X)` for policy fields, and `group_local.rollout_in_progress` directly. An `AFTER INSERT` trigger on `groups` auto-creates the matching `group_local` row on every node, including during a subscriber's initial COPY sync, so the sidecar is always present. [PR #1396](https://github.com/flatcar/nebraska/pull/1396).
+- **`activity`** mixes admin-originated rows (channel-package-updated) and runtime-originated rows (rollout lifecycle, instance update results). → Splits into `admin_activity` (replicated) + `activity` (local) + `all_activity` view. [PR #1398](https://github.com/flatcar/nebraska/pull/1398).
```
ervcz, line 56: I'm in support of splitting this table too, but I just wanted to sanity check with you that the reasons for this migration is whether to replicate the admin originated activities too or not, right? So, in an edge we can still see what admin actions were performed from the central node, right? Am I right that if this requirement is not needed, than we can consider the activity table fully as a runtime table on the edge, since  admin actions cannot be performed there anyways?

ervcz (approved):
Very nice write-up!

docs/distributed-mode.md
ervcz: Ah, answered below. NVM, unless there are other reasons too.

docs/distributed-mode.md
```diff
@@ -0,0 +1,212 @@
...
+Each PR is independently mergeable. Single-instance behaviour is preserved throughout.
+
+1. `group_local` sidecar split. [#1396](https://github.com/flatcar/nebraska/pull/1396)
+2. `activity` row-level split + `all_activity` view. [#1398](https://github.com/flatcar/nebraska/pull/1398)
+3. Extract `pkg/api/dbreads/`.
+4. Split writes into `pkg/api/admin/` and `pkg/api/runtime/`.
+5. Conditional role grants migration + two-DSN migration/serving split (`NEBRASKA_MIGRATIONS_DB_URL`).
+6. Instance mode: `NEBRASKA_INSTANCE_MODE` + `requirePrimary` middleware + validated allowlist.
+7. Per-edge override management endpoint.
+8. Operator documentation.
```
ervcz, line 203: Perhaps these roll-out steps could be tracked in the original issue with sub-tasks if you want to track them. And this list could be reduced to the link to the gh issue.

docs/distributed-mode.md
Moustafa-Moustafa: Replicating the admin activities is one of the reasons as I try to keep the experience consistent across the fleet. An operator on an edge should still see what control did, so each edge feels like a Nebraska with shared control rather than a disjoint instance.
Another reason is to keep the table classification consistent. Basically, with the split a table is either a runtime or an admin table. If we don't split the activity table, it wouldn't be consistent. We can't consider it a purely runtime table, since admin code still writes to it on the control node. Also, this helps with the DB layer enforcement (the third enforcement layer as discussed in the doc). Since edge nodes would be running using a DB role that can write to the `activity` table but not the `admin_activity` table.

docs/distributed-mode.md
Moustafa-Moustafa: Are you referring to GH "sub-issues"? If so, I don't see the option to create sub-issues for the RFC issue. It could be that I don't have access or it's not enabled for the Nebraska repo.
Anyways, I updated the RFC to be more aligned with the current state of this design doc.

docs/distributed-mode.md
```diff
@@ -0,0 +1,212 @@
...
+
+A single env var, `NEBRASKA_INSTANCE_MODE`, with a small validated allowlist:
+
+| Value | Meaning |
+|---|---|
+| unset or `single` | Single-instance (today's behaviour) |
+| `control` | Control node |
+| `edge` | Edge node |
+
+Any other value fails the process at startup with a clear error so we are sure the node starts with an explicit setting for the node type.
```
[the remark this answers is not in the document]
Moustafa-Moustafa: I updated the RFC to be more aligned with the design doc.

docs/distributed-mode.md
```diff
@@ -0,0 +1,212 @@
...
+
+Edge nodes:
+- Return HTTP 403 on admin endpoints via a `requirePrimary` middleware.
+- Do not start the syncer.
+
+### 4.5 Database authentication
+
+Nebraska does not create, drop, or rotate Postgres roles. The operator provisions all roles ahead of deployment. Nebraska reads connection strings from environment variables.
+
+**Two-layer role model.** Migrations attach grants to a fixed set of `NOLOGIN` logical roles whose names Nebraska guarantees. The operator creates their own `LOGIN` roles with whatever names they like and adds them `IN ROLE` the logical ones; PG role inheritance carries the grants through. Nebraska only requires that the logical roles exist and that whatever login role the DSN connects as inherits from them.
```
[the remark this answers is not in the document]
Moustafa-Moustafa: This is already captured under the EFC's issue comments.

docs/distributed-mode.md
ervcz: Yes, I was, but that checklist is more than enough. Appreciate it, thanks! :)

ervcz:
@Moustafa-Moustafa 
The document says replication is safe because foreign keys only point from runtime tables to admin tables. I think that is true for inserts and updates, but I am not sure about deletes: the apply worker runs with session_replication_role = replica, so triggers (including ON DELETE CASCADE) do not fire on the edge. So deleting an app or group on the control node could leave the edge’s local runtime rows (for example instance_application, activity, or the group_local sidecar) pointing at a row that no longer exists.

ervcz:
One thing I did not see in the document: the in-memory caches. GetAppID and GetGroupID are only cleared by in-process add/update/delete, so on an edge (where admin changes arrive through replication) they may keep serving old data after a rename or delete on the control node. How are you thinking about cache invalidation on edges? What was the proposed solution to this  - I don't reember? E.g. use Postgres LISTEN/NOTIFY from an ENABLE ALWAYS trigger, or just skip the cache in edge mode, or add a shorter TTL for the cache?

PR #1412, opened 2026-07-02
refactor(api): extract domain types to internal/types

Moustafa-Moustafa:
# refactor(api): extract domain types to `pkg/api/internal/types`

Groundwork for the sub-package split described in [RFC #1375](https://github.com/flatcar/nebraska/issues/1375) (Rollout PR 3: extract `pkg/api/dbreads/`). See also the design doc in [#1405](https://github.com/flatcar/nebraska/pull/1405).

Move pkg/api data types and their methods into a new internal package pkg/api/internal/types. pkg/api keeps re-exports (type X = types.X, const/var re-exports) so callers stay unchanged. The re-exports let us minimize the size of the change and avoid updating any tests in this commit. ~~Eventually, after splitting into the 3 packages (dbreads, admin, runtime), we will drop the re-exports in the api package and callers can reference the types package directly.~~

Prerequisite for extracting the read layer into pkg/api/internal/dbreads, which is itself a step towards later splitting the admin and runtime writer packages. dbreads cannot import pkg/api without an import cycle, so shared types have to live in a neutral internal package first.

This commit is purely refactoring and shouldn't change any behavior.

### After the follow-up dbreads-extraction PR

```mermaid
flowchart TD
    handler[pkg/handler]
    omaha[pkg/omaha]
    syncer[pkg/syncer]
    api[pkg/api<br/>writers + shims]
    dbreads[pkg/api/internal/dbreads<br/>read queries]
    types[pkg/api/internal/types<br/>domain types]

    handler --> api
    omaha --> api
    syncer --> api
    api --> dbreads
    api --> types
    dbreads --> types
```

## How to use

- This PR is intended to be purely refactoring. To validate, we should ensure that there's no behavior change at all.

## Testing done

```
$ cd backend
$ make code-checks
go build ./...
./tools/check_pkg_test.sh
NEBRASKA_SKIP_TESTS=1 go test ./... >/dev/null
./tools/golangci-lint run --fix
0 issues.
go mod tidy

$ make check-backend-with-container
# ... spins up postgres via docker-compose.test.yaml, runs full backend suite
# with NEBRASKA_RUN_SERVER_TESTS=1, then tears down.
ok  github.com/flatcar/nebraska/backend/pkg/api                44.746s
ok  github.com/flatcar/nebraska/backend/pkg/auth               0.006s
ok  github.com/flatcar/nebraska/backend/pkg/middleware         0.010s
ok  github.com/flatcar/nebraska/backend/pkg/omaha              6.745s
ok  github.com/flatcar/nebraska/backend/pkg/random             0.005s
ok  github.com/flatcar/nebraska/backend/pkg/sessions           0.007s
ok  github.com/flatcar/nebraska/backend/pkg/sessions/memcache  0.017s
ok  github.com/flatcar/nebraska/backend/pkg/sessions/memcache/gob  0.007s
ok  github.com/flatcar/nebraska/backend/pkg/syncer             8.253s
ok  github.com/flatcar/nebraska/backend/test/api               33.606s
ok  github.com/flatcar/nebraska/backend/test/auth/oidc         6.755s
```
Additionally, a manual end-to-end scenario suite was run against the local container.

### DIFF
--- a/backend/pkg/api/actions.go
+++ b/backend/pkg/api/actions.go
@@ -1,26 +1,12 @@
 package api
 
 import (
-	"time"
-
 	"github.com/doug-martin/goqu/v9"
+
+	"github.com/flatcar/nebraska/backend/pkg/api/internal/types"
 )
 
-// FlatcarAction represents an Omaha action with some Flatcar specific fields.
-type FlatcarAction struct {
-	ID                    string    `db:"id" json:"id"`
-	Event                 string    `db:"event" json:"event"`
-	ChromeOSVersion       string    `db:"chromeos_version" json:"chromeos_version"`
-	Sha256                string    `db:"sha256" json:"sha256"`
-	NeedsAdmin            bool      `db:"needs_admin" json:"needs_admin"`
-	IsDelta               bool      `db:"is_delta" json:"is_delta"`
-	DisablePayloadBackoff bool      `db:"disable_payload_backoff" json:"disable_payload_backoff"`
-	MetadataSignatureRsa  string    `db:"metadata_signature_rsa" json:"metadata_signature_rsa"`
-	MetadataSize          string    `db:"metadata_size" json:"metadata_size"`
-	Deadline              string    `db:"deadline" json:"deadline"`
-	CreatedTs             time.Time `db:"created_ts" json:"created_ts"`
-	PackageID             string    `db:"package_id" json:"-"`
-}
+type FlatcarAction = types.FlatcarAction
 
 // AddFlatcarAction registers the provided Omaha Flatcar action.
 func (api *API) AddFlatcarAction(action *FlatcarAction) (*FlatcarAction, error) {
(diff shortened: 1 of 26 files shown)


ervcz (approved):
Clean extraction!
The note about callers eventually "referencing the types package directly" won't hold for `pkg/handler`/`omaha`/`syncer` while it's under `pkg/api/internal/`, Go's internal rule blocks sibling imports. Your diagram keeps `pkg/api` as the facade though, so this may just be a wording inconsistnecy

Moustafa-Moustafa:
> Clean extraction! The note about callers eventually "referencing the types package directly" won't hold for `pkg/handler`/`omaha`/`syncer` while it's under `pkg/api/internal/`, Go's internal rule blocks sibling imports. Your diagram keeps `pkg/api` as the facade though, so this may just be a wording inconsistnecy

Thanks @ervcz for the review.
Right, the diagram is the intended shape, and the sentence in the commit body was not accurate. dbreads/admin/runtime never carry aliases to begin with, and they'll import internal/types directly. The re-exports live only in pkg/api for external callers. I added them mainly to keep this PR small and reviewable, so nothing outside pkg/api had to change.

If we later want to eliminate them, a mechanical follow-up change can move internal/types to pkg/api/types and update handler/omaha/syncer/tests to reference types directly.

PR #1413, opened 2026-07-06
build(deps): upgrade pgx v4 to v5.9.2 in backend

ervcz:
## What

Upgrade the PostgreSQL driver from `github.com/jackc/pgx/v4 v4.18.3` to `github.com/jackc/pgx/v5 v5.9.2`, removing the flagged `jackc/pgx/v4` and `jackc/pgproto3/v2` (plus `pgconn`/`pgtype`) dependencies.

## Why

Clears the open alerts on `backend/go.mod`:
- **CVE-2026-41889** — pgx simple-protocol SQL sanitizer
- **CVE-2026-32286** — pgproto3 `DataRow` decode panic / DoS

Nebraska was not actually exploitable by either: the simple protocol is never enabled (queries run via `database/sql` on the extended protocol and are parameterized), and the pgproto3 panic is only reachable from a malicious/compromised PostgreSQL server. This bump resolves the alerts and keeps us on a supported release.

## Upgrade risk (separate from the CVEs)

Low. Nebraska uses no pgx-specific API — the driver is accessed only through `database/sql`/`sqlx`/`goqu` — so this major-version bump has no direct API surface to break. The only pgx reference in our code is the driver-registration import in `backend/pkg/api/api.go`.

## Testing

- `go build ./...` and `go vet ./...` pass
- Full `go test -p 1 ./...` suite passes against PostgreSQL 17

## Follow-up

`updater/go.mod` carries the same two indirect deps via a pinned `backend` snapshot; it will be updated in a separate PR once this lands.


krnowak (approved):
Looks good, but there seem to be a conflict to be resolved first.

PR #1435, opened 2026-07-28
Fix delete failures updating UI

### LINKED ISSUE #1433: Delete requests with error responses are treated as successful in the ui
When a user tries to delete an application or group and the server rejects the request (for example with 403 because they only have viewer access), the UI removes the item anyway. There is no error message. After a page refresh the item comes back because nothing was actually deleted on the server.

The backend behaves correctly. The problem is entirely on the frontend.

## What is going wrong

**1. `API.doRequest` does not check `response.ok` for DELETE**

In `frontend/src/api/API.ts`, GET/POST/PUT go through a handler that throws when `response.ok` is false. DELETE takes a separate path and returns the raw `fetch()` promise with no status check.

`fetch()` only rejects on network failures. A 403 or 500 still resolves, so callers think the request succeeded.

**2. Store methods update local state on any resolved promise**

In `frontend/src/stores/ApplicationsStore.ts`, `deleteApplication` and `deleteGroup` call the API and remove the item from local state in `.then()` with no `.catch()` and no error handling.

The same DELETE path is also used by `deleteChannel`, `deletePackage`, and `deleteChannelFloor`. Those methods use `await`, but since DELETE never throws on HTTP errors they still refresh local state as if the delete worked.

**3. The confirmation dialog does not handle failures**

`ConfirmationContent` calls the store delete methods when the user clicks Yes and does nothing if the request fails.

## Steps to reproduce

1. Log in as a user with viewer-only access (or any role that cannot delete).
2. Open an application or group menu and choose Delete.
3. Confirm with Yes.

The item disappears from the list immediately. No toast or error dialog appears. Refresh the page and the item is still there.

## Expected behavior

If the server returns an error status, the item should stay in the UI and the user should see clear feedback that the delete failed.

## Verification

I traced the code and wrote targeted unit tests with a mocked `fetch` returning `{ ok: false, status: 403 }`.

| Test | Result |
|------|--------|
| `API.deleteApplication` on 403 | Resolves instead of throwing |
| `ApplicationsStore.deleteApplication` on 403 | Removes app from cache and calls `emitChange()` |
| `ApplicationsStore.deleteGroup` on 403 | Removes group from cache and calls `emitChange()` |

All three confirm the broken behavior. The backend side is covered elsewhere — for example `e2e-oidc/authorization-viewer.spec.ts` asserts that DELETE as a viewer returns 403 at the API level.

## Suggested fix

1. Check `response.ok` for DELETE in `API.doRequest`, same as other methods.
2. Add error handling in the store delete methods so local state is not updated on failure.
3. Show user-visible error feedback from the confirmation flow (toast or inline error).

## Environment

- Component: Nebraska frontend
- Affected operations: delete application, group, channel, package, channel floor

false200:
Fixes: #1433

Deleting an app or group looked like it worked even when the server said no. The UI removed the item, showed nothing wrong, and a refresh brought it back. That happened because DELETE requests skipped the same error check as every other API call, and the store always cleared local state when the promise resolved.

This PR checks `response.ok` for DELETE in `API.doRequest` and adds error handling in `deleteApplication` and `deleteGroup` so the list only updates after a successful response. If the server returns 403 or 500 the item stays put and the user gets a generic error alert.

## How to use

Pull the branch and run the frontend against a backend where you can trigger a failed delete. Easiest option is to open DevTools, paste a fetch override that returns 403 for DELETE requests, then delete an app or group from the menu. The item should stay in the list and you should see the error dialog. Refresh and confirm it is still there.

You can also test with a viewer account on an OIDC setup if you have one. DELETE should be denied by the API and the UI should behave the same way.

Run the unit tests with:

```
cd frontend
npx cross-env TEST_TZ=true npx vitest run src/__tests__/Api/api.spec.js src/__tests__/stores/ApplicationsStore.delete.spec.ts
```

All 18 tests should pass.

## Testing done

Ran the vitest command above locally. Output:

```
Test Files  2 passed (2)
     Tests  18 passed (18)
```

The new API tests confirm DELETE throws on 403 and resolves on success. The store tests confirm the app and group are not removed from cache when delete fails and that the error alert is shown.

### DIFF
--- a/frontend/src/__tests__/Api/api.spec.js
+++ b/frontend/src/__tests__/Api/api.spec.js
@@ -163,4 +163,28 @@ describe('API methods using URLSearchParams', () => {
       }
     );
   });
+
+  describe('deleteApplication', () => {
+    let deleteFetchMock;
+
+    beforeEach(() => {
+      deleteFetchMock = vi.spyOn(global, 'fetch');
+    });
+
+    it('should throw when the response is not ok', async () => {
+      deleteFetchMock.mockResolvedValueOnce({
+        ok: false,
+        status: 403,
+      });
+
+      await expect(API.deleteApplication(applicationID)).rejects.toMatchObject({ status: 403 });
+    });
+
+    it('should resolve when the response is ok', async () => {
+      const response = { ok: true, status: 204 };
+      deleteFetchMock.mockResolvedValueOnce(response);
+
+      await expect(API.deleteApplication(applicationID)).resolves.toBe(response);
+    });
+  });
 });
--- a/frontend/src/__tests__/stores/ApplicationsStore.delete.spec.ts
+++ b/frontend/src/__tests__/stores/ApplicationsStore.delete.spec.ts
@@ -1,0 +1,48 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import API from '../../api/API';
+import ApplicationsStore from '../../stores/ApplicationsStore';
+
+describe('ApplicationsStore delete handling', () => {
+  beforeEach(() => {
+    vi.spyOn(window, 'alert').mockImplementation(() => {});
+  });
+
+  it('does not remove an application when delete fails', async () => {
+    vi.spyOn(API, 'deleteApplication').mockRejectedValue({ status: 403 });
+
+    const store = new ApplicationsStore();
+    store.applications = [{ id: 'app-1', name: 'Test' }];
+    const emitSpy = vi.spyOn(store, 'emitChange');
+
+    store.deleteApplication('app-1');
+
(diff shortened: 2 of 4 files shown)


false200:
@ervcz PTAL this, when you have time, thanks.

LexiNadolski:
@false200 Please refrain from tagging for reviews. We are working to get to everything but we have had an influx of new contributors. Please abide by these standards: https://docs.linuxfoundation.org/lfx/mentorship/standards-of-excellence.

false200:
Oh, sure! I wasn't aware of that. Sorry about it and thanks for letting me know.

ervcz:

Thanks for your patience.

Issue #1166, opened 2025-08-13
Implement Multi-Step Update Support in Nebraska


ervcz:
## Overview
Implement support for complex updates that require intermediate steps before reaching the target version. This addresses scenarios where direct updates would fail due to system incompatibilities.

## Background
Nebraska currently only supports single-step updates. However, complex system changes (partition migrations, configuration format changes) sometimes require intermediate updates to prepare the system.

## Core Requirements
- Extend packages table with prerequisite_id field
- Use package IDs for prerequisite relationships
- Implement "greater or equal" prerequisite rule
- Extended Omaha format for downstream Nebraska syncers
- No changes required for Flatcar clients

## Success Criteria
- [ ] Backend can orchestrate multi-step updates
- [ ] Syncers can reconstruct prerequisite relationships after sync
- [ ] Frontend provides visibility into update paths
- [ ] All edge cases handled gracefully

## Related Discussion
- https://github.com/flatcar/Flatcar/discussions/1831

aeron-gh:
@ervcz PTAL #1436 
whenever you have some time
thank you!

Issue #1131, opened 2025-07-05
Better dependency tracking between updater <- backend


tylerauerbeck:
Currently we've got two modules in this repository: `backend` and `updater`. As neither of these are versioned, this requires us to go grab the latest from `backend` each time we make an update so that `updater` is working from the latest changes that we've made.

What this results in is two changes for each one that we make since we first need to merge the changes into backend so they can then be picked up by updater.

Is this the best setup to accomplish what we're wanting to do? Are two separate modules providing the value that is intended?

aeron-gh:
/assign
Just a heads-up I've already been working on this issue, and the PR is nearly ready for submission.

aeron-gh:
@ervcz  PTAL

Had a look at this. The two modules don't actually build together right now.`go-omaha` moved `UpdateResponse.Manifest` to `Manifests []*Manifest` for multi-step updates and the backend followed, but the updater didn't:

    $ go work init ./backend ./updater && go vet ./updater/...
    updater/updateinfo.go:37:21: app.UpdateCheck.Manifest undefined

Nothing catches it because the updater workflow only triggers on `updater/**`, and the `updaters/**` filter in `backend.yml` is a typo, so it has never matched anything. Even with that fixed the tests would still compile against the pinned
backend rather than the checkout.

Opened #1436. Used a `go.work` in CI rather than a committed `replace`, for the reason you gave, and refreshed the pin. The tagged releases part needs amaintainer to push tags so I left that out.

PR #1436, opened 2026-07-28
build(updater): track backend dependency via Go workspace

aeron-gh:
Refs: #1131

## What

Build the updater against the backend in the checkout instead of the published pseudo-version pinned in `updater/go.mod`, and fix the workflow path filters so backend changes reach the updater job at all.

- `updater/updater_test.go`: adapt to the `pkg/api` admin and runtime split from #1434 and #1498, and realign the backend pin past it, since the current pin predates the split and the two no longer build together
- `.github/workflows/updater.yml`: trigger on `backend/**`, and run `go work init ./backend ./updater` before the tests
- `.github/workflows/backend.yml`: drop the `updaters/**` filter. The directory is `updater/`, so it never matched, and nothing under `backend/` imports the updater, so backend CI has no reason to run on updater changes. Repairing the typo would only have added CI runs that cannot fail for a related reason
- `go.work` and `go.work.sum` are gitignored, with the command in the ignore comment so it is discoverable

## Why

`updater/updater_test.go` is the only thing in the updater that depends on the backend, and it compiles against the pin rather than the working tree. A backend change therefore cannot break the updater's tests in the PR that causes it, which is the "two changes for each one that we make" in #1131.

@ervcz set out the shape of the fix on the issue: use a `replace` for local development, keep it out of the tree, and refresh the outdated pin. This PR follows that, with a Go workspace as the mechanism so CI can use it too. `go.work` is not part of the published module, so it cannot reach consumers. A committed `replace` would not reach them either, since Go ignores `replace` outside the main module, but it would then sit in the published `go.mod` doing nothing while hiding the pin from every local build. The workspace keeps that separation explicit.

The drift this exposes is recurring. #1414 refreshed the pin to the pgx v5 merge and ported `updateinfo.go` to the new `go-omaha` `Manifests` field, but since that refresh the `pkg/api` split has drifted the two modules apart again. On `main` today:

```
$ go work init ./backend ./updater && go vet ./updater/...
updater\updater_test.go:31:29: cannot use api (variable of type *api.API) as *"github.com/flatcar/nebraska/backend/pkg/api/runtime".Service value in argument to omaha.NewHandler
updater\updater_test.go:121:29: apiInstance.AddPackage undefined (type *api.API has no field or method AddPackage)
updater\updater_test.go:124:20: apiInstance.UpdateChannel undefined (type *api.API has no field or method UpdateChannel)
updater\updater_test.go:173:24: cnf.api.AddTeam undefined (type *api.API has no field or method AddTeam)
updater\updater_test.go:175:23: cnf.api.AddApp undefined (type *api.API has no field or method AddApp)
updater\updater_test.go:177:23: cnf.api.AddPackage undefined (type *api.API has no field or method AddPackage)
updater\updater_test.go:179:27: cnf.api.AddChannel undefined (type *api.API has no field or method AddChannel)
updater\updater_test.go:181:25: cnf.api.AddGroup undefined (type *api.API has no field or method AddGroup)
```

None of that is covered by CI today. The first commit fixes the current break, the second makes it CI's job to catch the next one in the PR that causes it.

## Trade-offs

These are the costs of the approach, worth weighing before it lands:

- CI now exercises the working tree rather than the pinned pseudo-version, so the pin could go stale without anything failing. The impact is limited because the backend is a test-only dependency of the updater and consumers of the library never build it, but the pin does need its own upkeep.
- A workspace applies MVS across both modules, so the updater resolves to the backend's dependency versions. That is real coupling rather than a free win: the updater picks up backend dependency bumps whether or not it needs them.
- I have not verified how the existing `gomod` Dependabot entry for `/updater` behaves with a workspace in CI. It is untouched here and should be unaffected, since `go.work` is generated at job time and never committed, but I have not tested that.

## Rebase note

Rebased onto current `main`. #1414 landed the pin refresh to pgx v5, the `go` 1.25 raise and the `updateinfo.go` port to `Manifests`, so this PR no longer carries any of that and `updateinfo.go` is untouched here. The manifest selection question from the previous revision, first entry versus last, was settled by #1414 on `Manifests[0]`, so this PR takes no position on it and any revisit belongs with the multi step update work in #1166.

Rebased again after #1552 made the domain types public and #1556 reworked the shared connection, which broke the updater tests against the in-tree backend a third time. The first commit absorbs that adaptation and moves the pin to current `main`.

## Testing

With the workspace in place, against the current tree, `go vet ./updater/...` and `go build ./updater/...` pass:

```
$ go work init ./backend ./updater
$ go vet ./updater/...
$ go build ./updater/...
```

And the published pin still builds on its own, without a workspace:

```
$ GOWORK=off go vet ./...   # in updater/
```

The full updater suite against PostgreSQL 17 last ran on the previous revision of this branch. The only test change since is the adaptation to the split, and the workspace CI arrangement this PR adds runs exactly that suite.

## Follow-up

I originally documented the workspace in `CONTRIBUTING.md`, but dropped it when rebasing onto #1440, which moved Nebraska development guidance to the [development guide](https://www.flatcar.org/docs/latest/updates-releases/nebraska/development/). Happy to send it there instead if you think it is worth documenting.

Tagging backend releases (`backend/v0.1.0`) so Dependabot can bump the updater automatically needs a maintainer to push tags, so it is not something this PR can do. With the workspace in place the pin no longer blocks development, so it only affects keeping the published module graph current.


aeron-gh:
@ervcz Ptal 
When u get chance 
Thank you!

ervcz:

Thanks for your patience.

aeron-gh:
@ervcz Rebased onto current main, conflicts are gone. #1414 already landed the updateinfo port, so this PR now carries only the workspace CI and the test adaptation to the pkg/api split. Details in the rebase note in the description.

PR #1442, opened 2026-07-29
Fix instance rename cancel triggering unnecessary refetch.

### LINKED ISSUE #1441: Canceling instance rename still triggers a refetch
## What's going wrong

On the instance details page, if you open the rename dialog and hit Cancel, the page still refetches the instance data. Nothing was saved, so that extra request shouldn't happen.

The problem is in `onEditHide` inside `frontend/src/components/Instances/Details.tsx`. Cancel calls `onHide()` with no argument, which means `newInstance` is `undefined`. The check uses `newInstance !== null`, and `undefined !== null` is true, so `onInstanceUpdated()` runs anyway.

```tsx
function onEditHide(newInstance?: Instance) {
  setShowEdit(false);
  if (newInstance !== null) {
    onInstanceUpdated();
  }
}
```

Save works fine because a successful update passes the updated instance into `onHide(updatedInstance)`.

## Expected behavior

- Cancel closes the dialog and does not refetch.
- Save closes the dialog and refetches so the page shows the new alias.

## Suggested fix

Change the condition to only refetch when an instance was actually returned.

```tsx
if (newInstance !== undefined) {
  onInstanceUpdated();
}
```

One line change in `Details.tsx`.

## Scope

Small frontend bug fix. One line in `Details.tsx`, test already written.

false200:
Fixes #1441

On the instance details page, opening the rename dialog and hitting Cancel still triggered a refetch even though nothing was saved. Cancel calls `onHide()` with no argument, so `newInstance` comes through as `undefined`. The check was `newInstance !== null`, and since `undefined !== null` is true, `onInstanceUpdated()` ran anyway.

The fix changes that check to `newInstance !== undefined`. Cancel closes the dialog and stops there. Save still passes the updated instance back and refetches as before.

A test was added in `frontend/src/__tests__/Instances/Details.spec.tsx` covering both paths — cancel should not refetch, save should.

## How to use

Pull the branch and open any instance details page. Click the menu, choose Rename, then Cancel without changing anything. The dialog should close and you should not see a fresh `GET` for that instance in the network tab.

To confirm save still works, open Rename again, change the alias, hit Save, and check that the page updates with the new name.

Run the unit tests with:

```
cd frontend
npx cross-env TEST_TZ=true npx vitest run src/__tests__/Instances/Details.spec.tsx
```

Both tests should pass.

## Testing done

Ran the vitest command above locally. Output:

```
Test Files  1 passed (1)
     Tests  2 passed (2)
```

The cancel test confirms `onInstanceUpdated` is not called when the dialog is dismissed. The save test confirms it is called once after a successful update.

### DIFF
--- a/frontend/src/__tests__/Instances/Details.spec.tsx
+++ b/frontend/src/__tests__/Instances/Details.spec.tsx
@@ -1,0 +1,122 @@
+import '../../i18n/config.ts';
+
+import { ThemeProvider } from '@mui/material/styles';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { BrowserRouter } from 'react-router';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import API from '../../api/API';
+import { Application, Group, Instance } from '../../api/apiDataTypes';
+import DetailsView from '../../components/Instances/Details';
+import themes from '../../lib/themes';
+
+function renderDetails(onInstanceUpdated = vi.fn()) {
+  const application = {
+    id: 'app-1',
+    name: 'Test App',
+  } as Application;
+
+  const group = {
+    id: 'group-1',
+    name: 'Test Group',
+    channel: null,
+  } as unknown as Group;
+
+  const instance = {
+    id: 'instance-1',
+    alias: '',
+    ip: '10.0.0.1',
+    application: {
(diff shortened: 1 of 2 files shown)


ervcz:

Thanks for your patience.

ervcz (approved):
Thanks for your contribution!

Issue #1172, opened 2025-08-13
[Multi-Step Update] Testing


ervcz:
## Description

Comprehensive test suite for multi-step update functionality.

## Tasks

- [ ] Unit tests for prerequisite evaluation logic
- [ ] Integration tests for complete update flows
- [ ] Syncer tests for relationship reconstruction
- [ ] Edge case tests (circular deps, missing packages, etc.)
- [ ] End-to-end tests with multiple Nebraska instances

PR #1443, opened 2026-07-29
test(api): add comprehensive tests for floor package edge cases

kartikangiras:
# Add edge case tests for multi step update floor packages

This PR adds comprehensive edge-case test coverage for the multi-step update (floor packages) feature. While the feature was implemented and tested for the happy path, several edge cases lacked test coverage: packages marked as floors being deleted, orphaned target packages, out-of-range floors, and syncer behavior when floor count exceeds the response limit.

Relates to #1172 
Five new tests ensure that:

Floor deletion is protected: packages marked as floors cannot be deleted until unmarked
Orphaned targets are handled gracefully: deleting a channel's target package (which is allowed) doesn't cause crashes; update requests fail with ErrNoPackageFound
Out-of-range floors are ignored: a floor mistakenly marked with a version higher than the target is excluded from all update paths, preventing instances from getting wedged on an unreachable prerequisite
Sequential progression works with per-query limits: regular clients step through all floors one by one, even when the floor count exceeds NEBRASKA_MAX_FLOORS_PER_RESPONSE
Syncer floor truncation is documented: a real gap where syncers receiving truncated floor lists are still granted to the full target is pinned down with a test that explains the limitation and will fail once it's fixed

## How to use

- Read the test cases in backend/pkg/api/floors_edgecases_test.go to understand the edge cases covered
- Run the new tests locally: go test ./pkg/api -run  'TestDelete|TestOrphaned|TestFloorAbove|TestManyFloors|TestSyncerFloorLimit' -v
- Verify all 5 new tests pass alongside the existing floor/update test suite

## Testing done

```yaml
psql -h 127.0.0.1 -p 5432 -U kartik -d postgres -c "CREATE ROLE postgres WITH LOGIN SUPERUSER PASSWORD 'nebraska';"
psql -h 127.0.0.1 -p 5432 -U kartik -d postgres -c "CREATE DATABASE nebraska_tests OWNER postgres;"
```
```yml
cd backend
go test ./pkg/api/ -run 'TestDeleteFloorPackageBlocked|TestOrphanedFloorAfterTargetPackageDeleted|TestFloorAboveTargetIgnored|TestManyFloorsExceedingResponseLimitProgression|TestSyncerFloorLimitTruncationKnownLimitation' -v
`
=== RUN   TestDeleteFloorPackageBlocked
--- PASS: TestDeleteFloorPackageBlocked (0.11s)
=== RUN   TestOrphanedFloorAfterTargetPackageDeleted
--- PASS: TestOrphanedFloorAfterTargetPackageDeleted (0.11s)
=== RUN   TestFloorAboveTargetIgnored
--- PASS: TestFloorAboveTargetIgnored (0.12s)
=== RUN   TestManyFloorsExceedingResponseLimitProgression
--- PASS: TestManyFloorsExceedingResponseLimitProgression (0.14s)
=== RUN   TestSyncerFloorLimitTruncationKnownLimitation
--- PASS: TestSyncerFloorLimitTruncationKnownLimitation (0.12s)
PASS
ok  	github.com/flatcar/nebraska/backend/pkg/api	1.614s
```

### DIFF
--- a/backend/pkg/api/floors_edgecases_test.go
+++ b/backend/pkg/api/floors_edgecases_test.go
@@ -1,0 +1,189 @@
+package api
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/guregu/null.v4"
+)
+
+// TestDeleteFloorPackageBlocked verifies that a package marked as a floor for
+// any channel cannot be deleted until it's unmarked first. Without this guard,
+// deleting a floor package would silently remove the channel_package_floors
+// row (ON DELETE CASCADE) and erase the mandatory-upgrade-path guarantee for
+// any instance still below that version.
+func TestDeleteFloorPackageBlocked(t *testing.T) {
+	a := newForTest(t)
+	defer a.Close()
+
+	setup := setupFloors(t, a, "delete-floor-blocked", []string{"2000.0.0"}, "3000.0.0")
+	floor := setup.Floors[0]
+
+	err := a.DeletePackage(floor.ID)
+	assert.Equal(t, ErrPackageIsFloor, err, "deleting a package marked as floor must be rejected")
+
+	// The package must still exist and still be a floor after the rejected delete.
+	pkgs, err := a.GetChannelFloorPackages(setup.Channel.ID)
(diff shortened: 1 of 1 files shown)


ervcz:

Thanks for your patience.

Issue #1008, opened 2025-04-15
Return proper status code when attempting to upload a duplicate package


ErvinRacz:
## Current situation

When testing the package metadata upload via the API, We receive 500 Internal Server Error. Initially it was  assumed this was due to an invalid token, but eventually discovered the error was caused by attempting to upload a package that already exists.

Currently, this results in a 500 response due to a database constraint violation:

```
ERROR: duplicate key value violates unique constraint "package_appid_version_arch_unique" (SQLSTATE 23505)
```

## Impact

The current behavior is misleading. A 500 error suggests a server-side issue, but in reality, it's a client-side mistake (duplicate package upload). This can lead to confusion, unnecessary debugging, and delays in troubleshooting.

## Ideal future situation

Instead of returning a 500, the API should return a more appropriate status code, such as:

- `409 Conflict` — to indicate that the request could not be completed due to a conflict with the current state of the resource.
- Or a custom error message that clearly states the cause (e.g., "Package with the same AppID, version, and arch already exists").

This would make it easier for users to understand what went wrong and take appropriate action.

## Implementation options

- Catch the specific database error for duplicate key violation and return a `409 Conflict` response with a meaningful error message.
- Add a check before attempting to insert the package, to validate if it already exists, and return an appropriate response.

## Additional information

Sample log entry:

```
{"level":"error","username":"tormath1","error":"ERROR: duplicate key value violates unique constraint \"package_appid_version_arch_unique\" (SQLSTATE 23505)","time":"2025-04-15T08:26:42Z","context":"nebraska","message":"addPackage - adding package {e96281a6-d1af-4bde-9a0a-97b76e56dc57 1 [] Flatcar Linux 9.9.9 0xc0074e57a0 flatcar_production_update.gz 0xc007d17ed0 2jmj7l5rSw0yVb/vlWAYkK/YBwk= 0 1 https://update.release.flatcar-linux.net/amd64-usr/9.9.9/ 9.9.9}"}
```

aeron-gh:
Had a look. The 500 comes from the blanket error return in `CreatePackage` (`pkg/handler/packages.go:66`). Nothing in `pkg/` classifies the unique violation today, so the handler can't tell a duplicate from a genuine failure.

Plan: add an `ErrDuplicatePackage` sentinel in `pkg/api` next to the existing ones, detect the constraint violation in `AddPackage`, and map it to 409 in the handler. Happy to pick this up if nobody else is on it.

[...]

PR #1445, opened 2026-07-30
fix(api): return 409 for duplicate package uploads

aeron-gh:
Refs: #1008

## What

Return 409 instead of 500 when creating a package that already exists, and make the syncer treat that same condition as benign.

- `pkg/api/types/errors.go`: a new `ErrDuplicatePackage` sentinel, defined next to the existing ones
- `pkg/api/admin/packages.go`: detect the unique constraint violation on the package insert and return the sentinel
- `pkg/handler/packages.go`: map it to 409, in the same `switch err` shape already used for the other client errors in this file
- `api/spec.yaml` and the regenerated `spec.gen.go`: document the 409 on `createPackage`
- `pkg/syncer/syncer.go`: on a duplicate, adopt the existing package instead of failing the channel sync

## Why

Uploading a package that already exists returned a 500, because `CreatePackage` returned a blanket `StatusInternalServerError` for anything `AddPackage` failed on and nothing in `pkg/` classified the Postgres unique violation. A client mistake looked like a server fault, which is what #1008 describes.

Detection sits next to the insert in the admin service rather than in the handler, so `pkg/handler` does not need to know Postgres error codes. That matches how `ErrInvalidPackage` and `ErrArchMismatch` are already surfaced. It is scoped to the `package` insert specifically, since `addPackage` also writes `package_channel_blacklist`, whose primary key raises the same SQLSTATE, and a conflict there is not a duplicate package.

The handler returns a fixed message rather than the underlying error. The sentinel text and the SQLSTATE are useful in logs, not in a response body.

## The syncer

@ervcz pointed out on the issue that the syncer reaches the same insert through `AddPackageWithMetadata`, and asked what happens to already downloaded files when that insert fails.

`getOrCreatePackage` already checks `GetPackageByVersionAndArch` for the same app, version and arch triple the constraint covers, so the syncer only reaches the insert when the package is genuinely absent. A duplicate therefore means either a race with another writer, or a transient failure on that read, since any error on that read falls through to `createPackage`.

That second question mattered more than I expected. With `hostPackages` enabled the payload filename is derived from arch and version, and the extra file names from arch, version and the original name, so a syncer-created row for that version already points at exactly the paths just written. The existing cleanup on insert failure removes them, so a duplicate deleted the payloads of the package that already existed while its row kept advertising them. That is the behaviour on main today, independent of this change.

The duplicate branch therefore skips the cleanup, logs at info since the syncer is an unattended background process, and returns the existing package so the channel sync carries on. If that lookup itself fails it logs a warning and returns the original error rather than a misleading one.

## Rebase note

Rebased over the `pkg/api` write surface split (#1434, #1498). The detection moved with `addPackage` into `pkg/api/admin`. The unit tests moved to the new `pkg/api/admin/packages_test.go`.

Rebased again over #1552 and #1556. The domain types are public now, so the sentinel lives in `pkg/api/types/errors.go` and the handler and syncer reference `types.ErrDuplicatePackage` directly, matching how `types.ErrArchMismatch` is handled since that refactor. No re-export in `pkg/api` anymore, upstream dropped that pattern.

## Testing

Added:

- `TestAddPackageDuplicate`, covering `AddPackage`, `AddPackageWithMetadata`, and that a different arch or version is not a duplicate
- `TestAddPackageBlacklistConflictIsNotDuplicate`, which pins the scoping down: a `package_channel_blacklist` primary key conflict raises the same SQLSTATE and must not surface as `ErrDuplicatePackage`
- `TestIsUniqueViolation`, a database free unit test of the classifier across bare, wrapped, foreign key, plain and nil errors
- `TestSyncer_CreatePackageDuplicate`, asserting the syncer returns the existing package
- `TestSyncer_CreatePackageDuplicateKeepsHostedFiles`, which serves a real payload over `httptest`, triggers the duplicate, and asserts the main payload and both extra files are still on disk with the right contents
- a `duplicate_conflict` case in `test/api/package_test.go`, asserting the endpoint returns 409 and that the body leaks neither the sentinel text nor the SQLSTATE

After the rebase, `go build ./...` and `go vet ./...` (which also typechecks the test files) pass, and rerunning `make codegen` with the pinned oapi-codegen v2.3.0 produces no diff.

The results below are from the previous revision of this branch, before the rebase onto the `pkg/api` split. The ported tests carry the same assertions in the new location, and CI runs the full suite:

```
ok  github.com/flatcar/nebraska/backend/pkg/api      131.778s
ok  github.com/flatcar/nebraska/backend/pkg/omaha     11.636s
ok  github.com/flatcar/nebraska/backend/pkg/syncer    16.195s
ok  github.com/flatcar/nebraska/backend/test/api       5.890s

--- PASS: TestAddPackageDuplicate
--- PASS: TestAddPackageBlacklistConflictIsNotDuplicate
--- PASS: TestIsUniqueViolation
--- PASS: TestSyncer_CreatePackageDuplicate
--- PASS: TestSyncer_CreatePackageDuplicateKeepsHostedFiles
--- PASS: TestCreatePackage/duplicate_conflict
```

`TestSyncer_CreatePackageDuplicateKeepsHostedFiles` is the one worth pointing at: it serves a real payload over `httptest`, triggers the duplicate, and asserts the main payload and both extra files are still on disk with the right contents.

## Note

Happy to split the syncer commit out if you would rather review the two separately, it is self contained.


ervcz:

Thanks for your patience.

aeron-gh:
@ervcz Rebased onto current main over the pkg/api split, conflicts are gone. Details in the rebase note in the description.

PR #1449, opened 2026-07-30
fix(api): serialize empty read results as [] instead of null

### LINKED ISSUE #588: backend API null returned for empty []
## Description

I noticed whilst testing some components in the [storybook PR](https://github.com/kinvolk/nebraska/pull/509) that the backend API returns null for empty [], where the spec and the frontend were not expecting this. This is a WATy quirk for golang json serialization.

## Impact

Not sure yet... Maybe some components not rendering correctly. Either breaking views, showing loading states wrong, or showing misleading information.

YO-WHATS-UP2:
# fix(api): serialize empty read results as [] instead of null

Fixes #588

`dbreads` declared its result slices with `var x []T`, so a query returning zero rows
yielded a nil slice — which `encoding/json` marshals as `null`. `spec.yaml` declares
these fields `type: array` and marks none of them `nullable`, so the API contradicted
its own contract, and callers compensated with `|| []`.

Initialise the slices as empty at the sites whose values reach a JSON response. Two of
them were explicit `ExtraFiles = nil` assignments, so `"extra_files": null` was emitted
for every package without extra files — the common case.

Also drops the workaround in `GetGroupVersionBreakdown`, which returned `[]string{}`
where the schema expects `versionBreakdownEntry` objects, and only worked because empty
arrays serialize identically regardless of element type.

## Testing done

The new test asserts the marshalled bytes equal `[]` rather than using `assert.Empty`,
which passes for a nil slice too. Confirmed failing on the previous code.

### DIFF
--- a/backend/pkg/api/empty_slice_json_test.go
+++ b/backend/pkg/api/empty_slice_json_test.go
@@ -1,0 +1,82 @@
+package api
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/guregu/null.v4"
+
+	"github.com/flatcar/nebraska/backend/pkg/api/types"
+)
+
+func assertMarshalsToEmptyArray(t *testing.T, name string, v interface{}) {
+	t.Helper()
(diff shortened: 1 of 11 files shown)


ervcz:

Thanks for your patience.

PR #1452, opened 2026-07-31
fix(api): base version breakdown percentage on filtered instances

### LINKED ISSUE #1451: Version breakdown percentages do not add up to 100% when a group has instances with bracketed IDs
## What I found

`GetGroupVersionBreakdown` skips instances with bracketed IDs in its `WHERE`, but the `total` it divides by comes from a separate subquery that does not have that filter:

```sql
SELECT version, count(*) as instances, (count(*) * 100.0 / total) as percentage
FROM instance_application, (
    SELECT count(*) as total
    FROM instance_application
    WHERE group_id=$1 AND last_check_for_updates > now() at time zone 'utc' - interval '1 days'
    ) totals                                                            -- no filter here
WHERE group_id=$1 AND last_check_for_updates > now() at time zone 'utc' - interval '1 days'
      AND (instance_id IS NULL OR instance_id NOT LIKE '{________-____-____-____-____________}')
GROUP BY version, total
```

(`backend/pkg/api/internal/dbreads/groups.go`)

So the numerator counts only real instances while the denominator counts real and bracketed ones together. Every percentage comes out lower than it should be, and a group's percentages add up to 100% only when it has no bracketed instances at all.

The `instances` counts in the same response are correct. Only `percentage` is affected.

## What it looks like

I set up a group with 15 normal instances (9 on `4152.2.3`, 6 on `4230.2.1`) and 5 instances with bracketed IDs. `GET /api/apps/<app>/groups/<group>/version_breakdown` gives:

```json
[{"version":"4230.2.1","instances":6,"percentage":30},
 {"version":"4152.2.3","instances":9,"percentage":45}]
```

45 + 30 = 75, so a quarter of the group is unaccounted for. The correct split for the 15 real instances is 60 / 40.

The same numbers reach the group card, where the tooltip shows 45.0% and 30.0%:

version breakdown tooltip showing 45.0% and 30.0%

With the denominator filtered the same group reads 60.0% and 40.0%:

version breakdown tooltip showing 60.0% and 40.0%

One thing worth mentioning, because it makes this easy to miss: the bar itself still looks fine. `VersionBreakdownBar` uses `<XAxis hide type="number" />` without a domain, so recharts scales the axis to whatever the values add up to and the bar always fills its width. The proportions between segments are unaffected too, since every value is scaled by the same factor. What is wrong is the numbers.

## The "Other" bucket

There is a second effect that is easier to overlook. `VersionBreakdownBar` moves every version under 10% into the "Other" bucket:

https://github.com/flatcar/nebraska/blob/main/frontend/src/components/common/VersionBreakdownBar/VersionBreakdownBar.tsx#L101

Because all percentages are scaled down, a version that is really above 10% can drop below the threshold and disappear. With a group of 180 normal instances (100 on `4152.2.3`, 61 on `4230.2.1`, 19 on `4081.2.0`) and 20 bracketed ones, which is a 10% bracketed ratio:

| version | instances | reported now | should be |
| --- | --- | --- | --- |
| 4152.2.3 | 100 | 50.0% | 55.6% |
| 4230.2.1 | 61 | 30.5% | 33.9% |
| 4081.2.0 | 19 | 9.5% | 10.6% |
| | | **90.0%** | **100.0%** |

`4081.2.0` holds 10.6% of the real fleet but is reported as 9.5%, so it is folded into "Other" and stops being visible on the card:

tooltip showing 30.5%, 50.0% and Other 9.5%

With the denominator filtered it comes back as its own version:

tooltip showing 33.9%, 4081.2.0 at 10.6% and 55.6%

## On a bigger group

PostgreSQL 17, one group with 200k instances across six versions, 2% of them bracketed:

| | percentages add up to | execution time (median of 5) |
| --- | --- | --- |
| current | 98.0000% | 142.8 ms |
| with the denominator filtered | 100.0000% | 55.8 ms |

The per version `instances` values were identical in both runs, only the percentages changed. The time difference comes from the extra scan the subquery does.

## Where it comes from

The filter has been one sided since b6bb9b5c ("backend: Ignore instances with bracketed IDs in more places"), whose commit message says it is there "to avoid skews in statistics or metrics resulting from taking fake instances into account". That commit added the condition to several queries, and in this one it only reached the outer query. The query has been moved and reworked a few times since then, but the subquery was never revisited.

I also checked the other callers of `ignoreFakeInstanceCondition`. They all apply it in a single `WHERE`, so this is the only query where the numerator and the denominator can disagree.

## Suggested fix

Rather than repeating the condition in the subquery, the denominator can come from the grouped rows themselves:

```sql
SELECT version, count(*) as instances, (count(*) * 100.0 / sum(count(*)) over ()) as percentage
FROM instance_application
WHERE group_id=$1 AND last_check_for_updates > now() at time zone 'utc' - interval '1 days' AND <filter>
GROUP BY version
```

Both sides then read the same rows by construction, so they cannot drift apart again if the filter changes later, and `instance_application` is scanned once instead of twice.

PR with tests: #1452

shubham76-dev:
Fixes #1451

## Why

`GetGroupVersionBreakdown` filters out instances with bracketed IDs, but the `total` it divided by came from a subquery without that filter, so the numerator and the denominator were counting different sets of rows. Every percentage came out lower than it should be, and a group's percentages only added up to 100% if it had no bracketed instances. The `instances` counts in the same response were always correct, only `percentage` was off.

The visible effect is in the group card. The tooltip shows these percentages directly, and since `VersionBreakdownBar` folds anything under 10% into "Other", a version that is really above 10% could get pushed into "Other" and disappear from the card. The bar width itself looked fine, because the chart has no axis domain set and scales to whatever the values add up to, which is probably why this went unnoticed for so long.

## What changed

Instead of repeating the same condition inside the subquery, the denominator now comes from the grouped rows through `sum(count(*)) over ()`. Both sides read the same rows by construction, so they cannot drift apart again if someone changes the filter later, and `instance_application` is scanned once instead of twice.

```sql
SELECT version, count(*) as instances, (count(*) * 100.0 / sum(count(*)) over ()) as percentage
FROM instance_application
WHERE group_id=$1 AND last_check_for_updates > ... AND <bracketed id filter>
GROUP BY version
```

## Testing

`TestGetGroupsFiltered` already sets up one real and two bracketed instances, so it only needed the percentage assertion. It reported 33.33% before this change instead of 100%.

I also added `TestVersionBreakdownPercentages`, which uses two versions plus a bracketed instance and checks both the split and the total. Without the fix it reports 20 / 60 adding up to 80, with the fix 25 / 75 adding up to 100.

Full backend suite passes against PostgreSQL 17.

## Numbers

One group, 200k instances, six versions, 2% of them bracketed:

| | percentages add up to | execution time (median of 5) |
| --- | --- | --- |
| before | 98.0000% | 142.8 ms |
| after | 100.0000% | 55.8 ms |

`instances` per version was identical in both runs.

## Note

I kept this to the one query. The other callers of `ignoreFakeInstanceCondition` all apply it in a single `WHERE`, so this was the only place where the numerator and denominator could disagree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bugfixes
 
 - Fixed package blacklist changes not appearing in UI immediately after save
+- Fixed group version breakdown percentages not adding up to 100% for groups containing instances with bracketed IDs
 
 ## [3.0.0] - 28/11/2025
 
--- a/backend/pkg/api/groups_test.go
+++ b/backend/pkg/api/groups_test.go
@@ -191,7 +191,45 @@ func TestGetGroupsFiltered(t *testing.T) {
 			vb := versionBreakdown[0]
 			assert.Equal(t, "1.0.0", vb.Version)
 			assert.Equal(t, 1, vb.Instances)
+			// The only instance that counts is the real one, so it accounts
+			// for the whole group and not just for a share of it.
+			assert.Equal(t, 100.0, vb.Percentage)
 		}
+	}
+}
+
+func TestVersionBreakdownPercentages(t *testing.T) {
+	a := newForTest(t)
+	defer a.Close()
+	as := adminSvc(a)
+
+	tTeam, _ := as.AddTeam(&Team{Name: "test_team"})
+	tApp, _ := as.AddApp(&Application{Name: "test_app", TeamID: tTeam.ID})
+	tPkg, _ := as.AddPackage(&Package{Type: PkgTypeOther, URL: "http://sample.url/pkg", Version: "12.1.0", ApplicationID: tApp.ID})
+	tChannel, _ := as.AddChannel(&Channel{Name: "test_channel", Color: "blue", ApplicationID: tApp.ID, PackageID: null.StringFrom(tPkg.ID)})
+	tGroup, _ := as.AddGroup(&Group{Name: "test_group1", ApplicationID: tApp.ID, ChannelID: null.StringFrom(tChannel.ID), PolicyUpdatesEnabled: true, PolicySafeMode: true, PolicyPeriodInterval: "15 minutes", PolicyMaxUpdatesPerPeriod: 2, PolicyUpdateTimeout: "60 minutes"})
+
+	for _, version := range []string{"1.0.0", "1.0.0", "1.0.0", "2.0.0"} {
+		_, _ = a.RegisterInstance(Instance{ID: uuid.New().String(), IP: "10.0.0.1"}, NewInstanceApplication(tApp.ID, tGroup.ID, version))
+	}
+	_, _ = a.RegisterInstance(Instance{ID: "{" + uuid.New().String() + "}", IP: "10.0.0.2"}, NewInstanceApplication(tApp.ID, tGroup.ID, "3.0.0"))
+
+	versionBreakdown, err := a.GetGroupVersionBreakdown(tGroup.ID)
+	assert.NoError(t, err)
+
+	total := 0.0
+	for _, entry := range versionBreakdown {
+		total += entry.Percentage
+	}
+	assert.InDelta(t, 100.0, total, 0.01)
+
+	if assert.Len(t, versionBreakdown, 2) {
+		assert.Equal(t, "2.0.0", versionBreakdown[0].Version)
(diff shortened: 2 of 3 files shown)


ervcz:

Thanks for your patience.

ervcz:
Hey @shubham76-dev, thank you very much for this! It solves a real conflict in the sql.

Could you please rebase your work and check if it still runs?

PR #1456, opened 2026-07-31
fix(api): return oem and aleph_version from the instances list endpoint

### LINKED ISSUE #1454: instance oem and aleph_version are served but undeclared, and the list endpoint omits them
While reading the instance code path I noticed that `oem` and `aleph_version` are recorded and partially served, but are not part of the OpenAPI contract, and the single-instance and list endpoints disagree about whether they are returned at all.

## What the data path looks like today

`oem` is ingested from the Omaha check-in and persisted:

- `backend/pkg/omaha/omaha.go:154` sets `OEM: reqApp.OEM` on the `api.Instance`
- `backend/pkg/api/instances.go:96-98` upserts the `oem` column
- `backend/pkg/api/db/migrations/0021_add_instance_oem.sql` adds `oem` and `aleph_version` as `VARCHAR(256) NOT NULL DEFAULT ''`
- `backend/pkg/api/internal/types/instance.go:48-49` carries both with `json:"oem,omitempty"` / `json:"aleph_version,omitempty"`

## The inconsistency

**`GET /api/apps/{app}/groups/{group}/instances/{id}` returns them.** `GetInstance` builds `goqu.From("instance")` with no `.Select(...)` (`dbreads/instances.go:66-68`), so it is `SELECT *` and `StructScan` populates `OEM` and `AlephVersion`. The handler passes the internal `types.Instance` straight to `ctx.JSON` (`handler/instances.go:26`), and `server.go:113` installs only `OapiRequestValidatorWithOptions` — request validation, not response validation. So both fields go out on the wire.

**The list endpoint does not.** `GetInstances` explicitly selects four columns (`dbreads/instances.go:186-187`):

```go
instancesQuery.Select("id", "ip", "created_ts", goqu.Case().
    When(goqu.C("alias").Neq(""), goqu.C("alias")).Else(goqu.C("id")).As("alias"))
```

`oem` and `aleph_version` are left unset, and `omitempty` drops them from the JSON.

**Neither is declared in the spec.** The `instance` schema (`backend/api/spec.yaml:1955`) lists only `id`, `ip`, `createdTs`, `application` and `alias`, so `codegen.Instance` (`backend/pkg/codegen/types.gen.go:220-226`) has no field for either. The generated client and the frontend TypeScript types cannot see data the API is already emitting — `grep -ri oem frontend/src` returns nothing.

So the same logical resource has three different shapes depending on whether you read the spec, the list endpoint, or the single-instance endpoint.

## Why it might matter

`oem` identifies the hardware/cloud platform a machine runs on (`azure`, `gcp`, and similar values appear in `backend/pkg/api/instances_test.go:50-74`). That is fleet composition information Nebraska already stores per instance but never surfaces in the dashboard, and there is no aggregate query or Prometheus metric over it either — `backend/pkg/metrics/metrics.go:19-65` registers only `application_instances_per_channel`, `failed_updates`, and three DB connection gauges.

## Questions

1. Is the spec/handler drift intentional, or should the `instance` schema declare `oem` and `aleph_version` so the contract matches what is served?
2. Should the list endpoint select them too, so the two endpoints agree?
3. If an OEM distribution report were added, would you want it grouped per application/group like `GetGroupVersionBreakdown` (`dbreads/groups.go:301`) — which needs an `instance` ↔ `instance_application` join, since `oem` lives on `instance` and `group_id` on `instance_application` — or fleet-wide?

Happy to send a PR for whichever shape you prefer.

WasThatRudy:
Fixes #1454

## Why

The paginated instances list endpoint silently drops the `oem` and `aleph_version` fields: its CTE selects an explicit column list that predates the OEM attribute capture, and the `omitempty` JSON tags then remove the zero-valued fields from every list response. The single-instance endpoint returns both via `SELECT *`. On top of that, neither field was declared in the OpenAPI `instance` schema, so spec-generated clients could not see them on any endpoint. The upcoming distribution reporting work builds on these fields being visible.

## What

- Add `oem` and `aleph_version` to the instances list query and its row scan. The select list is now a package-level `instanceListColumns` variable shared by `GetInstances` and `GetInstancesCount` so the two queries cannot drift apart again.
- Declare `oem` and `alephVersion` as optional string properties on the `instance` schema, mirroring the existing `createdTs` snake_case tag override, and regenerate the codegen output.
- Regression tests at both layers: a Postgres-backed query test (`TestGetInstancesOEMAndAlephVersion`) that fails without the query fix, and a server test asserting both fields arrive over the wire through the spec-generated client type, including the empty-OEM case where both fields are omitted.

## Test plan

- [x] `go build ./...` and `go vet ./...`
- [x] `golangci-lint run` (0 issues)
- [x] `NEBRASKA_SKIP_TESTS=1 go test ./...` (compile check)
- [x] `go test -p 1` for `pkg/api`, `pkg/api/admin`, `test/api` with `NEBRASKA_RUN_SERVER_TESTS=1` against containerized PostgreSQL 17 and the rebuilt server image
- [x] Verified the new query-layer test fails when the list query change is reverted
- [x] `make codegen` output committed, work tree clean

Manual verification: register instances with and without OEM attributes, then compare `GET /api/apps/{app}/groups/{group}/instances` (both fields now present when set, omitted when empty) against `GET .../instances/{id}`.


ervcz:

Thanks for your patience.

WasThatRudy:
> Sorry for the slow reply, and thanks for the contribution.
> 
> Nebraska is one of the projects proposed for CNCF/LFX mentorship this term, and since that went up we've had a lot of attention here at once. The queue grew much faster than we can review it, so replies are slower than usual. Your PR hasn't been forgotten and it will get a proper look.
> 
> Thanks for your patience.

no issues, you guys must be stacked, take your time :)

backend/pkg/api/internal/dbreads/instances.go
```diff
@@ -134,6 +134,12 @@ func (q *Queries) GetInstanceStatusHistory(instanceID, appID, groupID string, li
 	return instanceStatusHistory, nil
 }
 
+// instanceListColumns holds the instance columns selected by both the
+// instances list query and its count query; the scan in GetInstances relies
+// on this exact order.
+var instanceListColumns = []interface{}{"id", "ip", "created_ts", "oem", "aleph_version", goqu.Case().
```
chengr4, line 140: maybe we can change to `[]any`?

backend/pkg/api/internal/dbreads/instances.go
```diff
@@ -134,6 +134,12 @@ func (q *Queries) GetInstanceStatusHistory(instanceID, appID, groupID string, li
 	return instanceStatusHistory, nil
 }
 
+// instanceListColumns holds the instance columns selected by both the
+// instances list query and its count query; the scan in GetInstances relies
```
chengr4, line 138: Do we forget to modify `GetInstancesCount`?

backend/api/spec.yaml
```diff
@@ -1963,6 +1963,12 @@ components:
           type: string
         ip:
           type: string
+        oem:
+          type: string
+        alephVersion:
```
chengr4, line 1968: Just a small suggestion: since `alephVersion` is derived from the Omaha protocol, could we perhaps add a brief description here? E.g.:

```yaml
description: The OS version the instance was originally installed with.
```

backend/api/spec.yaml
```diff
@@ -1963,6 +1963,12 @@ components:
           type: string
         ip:
           type: string
+        oem:
```
chengr4, line 1966: OEM also added e.g. 🤔

```yaml
description: The hardware platform or cloud provider identifier (e.g. ami, azure, gcp, qemu).
```

backend/pkg/api/instances_test.go
```diff
@@ -191,6 +192,38 @@ func TestGetInstances(t *testing.T) {
 	assert.Error(t, err, "Application id and group id are required and must be valid uuids.")
 }
 
+func TestGetInstancesOEMAndAlephVersion(t *testing.T) {
```
chengr4, line 195: I am not sure if we should keep this as a standalone function or move it under `TestGetInstances` + `t.Run`. 

I personally prefer the latter.

chengr4:
Hi @WasThatRudy,

Thank you for the PR! Since I am **the assignee of this issue**, I am helping review this PR.

Could you please fix the merge conflicts first so that I can review your test cases?

Thank you

--

cc @ervcz @John15321

PR #1458, opened 2026-08-01
frontend: replace array index with unique id for React keys

### LINKED ISSUE #328: key values using array index
**Description**

There's a few places in the code using array index as key values. It's unclear to me if each of the times this happens it is intended and safe.

**Impact**

Perhaps some views will be updated with the incorrect items.

invinciblearyan:
Fixes #328

This PR removes the use of array indices as `React key props` during mapping, replacing them with unique identifiers (like `id` or `name`). This prevents potential reconciliation bugs when the UI updates.

## How to use

Review the code changes to ensure that .map() functions in the frontend now use unique data properties for their key props instead of the array index i.

## Testing done
- Ran `npm run lint` and fixed all warnings (removed unused index variables).

- Ran `npm run format` to ensure styling compliance.

- Ran `tsc -b` to verify no typescript errors.

Built and ran the frontend locally to ensure components render correctly.

### DIFF
--- a/frontend/src/components/Applications/ApplicationItemChannelsList.tsx
+++ b/frontend/src/components/Applications/ApplicationItemChannelsList.tsx
@@ -16,8 +16,8 @@ function ApplicationItemChannelsList(props: { channels?: Channel[] }) {
 
   return (
     <Grid container justifyContent="space-between">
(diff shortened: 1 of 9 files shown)


ervcz:

Thanks for your patience.

PR #1469, opened 2026-08-02
fix(api): return fresh instance after single-table RegisterInstance u…

### LINKED ISSUE #1448: RegisterInstance returns stale instance after single-table updates
**Description**
When RegisterInstance updates only the instance table or only instance_application (not both), it writes the new values to the DB but returns the pre-update object.

The dual-update path correctly re-fetches via GetInstance; the single-table path does not.

**Steps to reproduce**

1. Register an instance.
2. Call RegisterInstance again changing only the IP (same version/group, within the recent-check window).
3. Compare the returned *Instance with GetInstance.

Harsh63870:
## Summary
- Fixes #1448: when `RegisterInstance` updates only `instance` or only `instance_application`, it now re-fetches via `GetInstance` instead of returning the stale pre-update object.
- Adds a regression test covering IP/alias-only and version-only updates.

## Test plan
- [ ] `cd backend && go test ./pkg/api/ -run 'TestRegisterInstance'`
- [ ] Confirm dual-update path (IP + version together) still returns updated fields
- [ ] Confirm IP-only update within the 5-minute window returns the new IP/alias


Harsh63870:
hi @ervcz , friendly ping when you have a moment, small bugfix (#1448), stale instance returned on single-table RegisterInstance updates, with regression test, waiting on code owner review/workflow approval. Thanks!

ervcz:

Thanks for your patience.

Harsh63870:
thanks @ervcz, totally understand, and no rush at all. Appreciate you taking the time to explain.
I'll resolve the `instances.go` conflict with current `main` so it stays easy to review whenever you get to it, happy to adjust anything once you've had a chance to look.

PR #1479, opened 2026-08-03
feat(api): add OEM/platform distribution breakdown per group

sid200727:
Add OEM/platform distribution breakdown endpoint for groups

Surfaces which hardware/cloud platform (OEM, e.g. AWS, Azure, VMware) each instance in a group runs on data Nebraska already records via `instance.oem` but never exposes. Adds `GET /api/apps/{appIDorProductID}/groups/{groupID}/oem_breakdown`, mirroring the existing `version_breakdown` endpoint's shape and query pattern exactly:

-> New OpenAPI path + schema in spec.yaml, codegen regenerated via `make codegen` (oapi-codegen v2.3.0)
-> `types.OEMBreakdownEntry` (mirrors `VersionBreakdownEntry`)
-> `dbreads.GetGroupOEMBreakdown` - groups by `instance.oem`, joined from `instance_application` to `instance` (oem lives on `instance`, not `instance_application`, unlike version)
-> `handler.GetGroupOEMBreakdown` wiring DB to HTTP JSON response
-> `api.OEMBreakdownEntry` type alias, matching the existing pattern for `VersionBreakdownEntry`

## How to use

Hit `GET /api/apps/{appID}/groups/{groupID}/oem_breakdown` against a group with instances - response shape matches `version_breakdown` (array of `{oem, instances, percentage}`), or `curl` it locally after `make run`.

## Testing done

`TestGroupOEMBreakdown` added to `test/api/stats_test.go`, mirroring `TestGroupVersionBreakdown` - compares DB query output directly against the live HTTP endpoint response.
Full suite green, including:

--- PASS: TestGroupOEMBreakdown (0.16s)
--- PASS: TestGroupOEMBreakdown/success (0.16s)
ok github.com/flatcar/nebraska/backend/test/api 1.428s
`golangci-lint run` - 0 issues.


backend/pkg/api/internal/dbreads/groups.go
```diff
@@ -336,6 +336,45 @@ func (q *Queries) GetGroupVersionBreakdown(groupID string) ([]*types.VersionBrea
 	return entryList, nil
 }
 
+// GetGroupOEMBreakdown returns the distribution of hardware/cloud platforms
+// (OEM) currently reported by the instances belonging to a given group.
+// Mirrors GetGroupVersionBreakdown's shape, but groups by instance.oem
+// instead of instance_application.version, so it joins instance_application
+// to instance (oem lives on instance, not instance_application).
```
chengr4, line 343: I think this should be removed

backend/pkg/api/internal/dbreads/groups.go
```diff
@@ -336,6 +336,45 @@ func (q *Queries) GetGroupVersionBreakdown(groupID string) ([]*types.VersionBrea
...
+	SELECT oem, count(*) as instances, (count(*) * 100.0 / total) as percentage
+	FROM instance_application
+	JOIN instance ON instance_application.instance_id = instance.id, (
+		SELECT count(*) as total
+		FROM instance_application
+		WHERE group_id=$1 AND last_check_for_updates > now() at time zone 'utc' - interval '%[1]s'
+		) totals
+	WHERE group_id=$1 AND last_check_for_updates > now() at time zone 'utc' - interval '%[1]s' AND %[2]s
+	GROUP BY oem, total
+	ORDER BY instances DESC
```
chengr4, line 357: Honestly, #1571 might provide a better query. Could you please refer to it and improve the query here?

backend/pkg/handler/groups.go
```diff
@@ -231,6 +231,24 @@ func (h *Handler) GetGroupVersionBreakdown(ctx echo.Context, _ string, groupID s
...
+		}
+		l.Error().Err(err).Str("groupID", groupID).Msg("getOEMBreakdown - getting OEM breakdown")
+		return ctx.NoContent(http.StatusInternalServerError)
+	}
+
+	if len(oemBreakdown) == 0 {
+		// WAT?: because otherwise it serializes to null not []
+		return ctx.JSON(http.StatusOK, []string{})
+	}
+	return ctx.JSON(http.StatusOK, oemBreakdown)
```
chengr4, line 252: I think this should be fixed in `dbreads/group.go`

chengr4 (changes requested):

chengr4:
Hi @sid200727,

Thank you for the PR. Since **I am the assignee** of this issue, I will review this PR for the team.

Can you please take the following steps to improve this PR?

1. Fix merge conflicts.
2. Refer to the backend part of #1571, improve the query you made, and add proper unit tests for `GetGroupOEMBreakdown`.

  Also, give credit to anyone who contributed to the patch by adding this at the end of the commit message:

  ```bash
  Co-authored-by: Harsh63870 <harshvardhanpandey372@gmail.com>
  ```
3. Improve the integration test as well; right now it does not test anything.

**Please reply to this PR within 1 week on whether you would still like to work on this, or we will close this PR.**

If you have any question, feel free to ask.

Thank you so much again!

---
cc @ervcz

sid200727:
Hi @chengr4 ! Looked at #1571, its query is better than mine: COALESCE(NULLIF(...), 'unknown') for missing OEM, and a sum(count(*)) OVER () window instead of my correlated subquery. It also has the spec, a unit test, a CHANGELOG entry and the frontend chart, so at that point it's substantially the same PR.

How would you like these reconciled? 
I'm happy to adopt that query and test shape here, or if #1571 is the one going forward it may make more sense to close this one. Whichever avoids duplicating the work.

chengr4:
> How would you like these reconciled? I'm happy to adopt that query and test shape here, or if #1571 is the one going forward it may make more sense to close this one. Whichever avoids duplicating the work.

Hi @sid200727,

Since your PR came first and I think you have the passion to move forward with it, we will push your endeavor first! You will have a lot of code duplication with #1571 , which is why the 'Co-authored-by' commit is very important.

My plan right now is to have your PR handle the backend while #1571 handles the frontend. So please only refer to #1571's backend changes and **do not touch its frontend**. 

Also, please address the review and fix comments I left here, and we should be good to go! :)

sid200727:
Thank You @chengr4! that split works for me, I'll keep #1479 backend-only and leave #1571's frontend alone, with the Co-authored-by trailer for @Harsh63870 on the borrowed query.

One check before I push: #1590 also looks like an OEM breakdown backend. Does that overlap with what you'd like #1479 to cover, or are they separate pieces?
For context on the codegen question @ervcz raised on #1590: #1479 is rebased onto current main (past the pkg/api/runtime/ and api/types moves) and its generated files come from an actual make codegen run.

chengr4:
> One check before I push: #1590 also looks like an OEM breakdown backend. Does that overlap with what you'd like #1479 to cover, or are they separate pieces?

That one should be closed after you are finished.

PR #1485, opened 2026-08-04
fix(dev): enable development on macOS

### LINKED ISSUE #374: [RFE]: Provide dev contribution guide for Mac OS system
## Current situation
Mac OS runs zsh and the makefile of the project fails to execute certain commands on MAC Os.
Ex: `cd backend && ./tools/check_pkg_test.sh`
One can try to apply some hacks over the file to run the project manually.
But if option comes from upstream it will be something trustworthy.
Makefile of the project acc

## Impact

MAC Os users will be able to contribute.

KartikSuryavanshi:
Closes #374

## Summary

Makes Nebraska's development tooling work on macOS, which ships bash 3.2 and BSD utilities by default.

## Changes

- **`backend/tools/check_pkg_test.sh`** — replaced `mapfile` (bash 4+) with a `while read` loop, which works on the bash 3.2 shipped with macOS. Verified by running the script under `/bin/bash` (3.2.57).
- **`backend/tools/setup_local_db.sh`** and **`backend/tools/run-all.sh`** — replaced the GNU `getopt --longoptions` calls with portable argument parsing that also supports the `--opt=value` form used by `run-all.sh`. Verified under `/bin/bash` (3.2): short, long, and `=` forms parse; invalid and missing arguments produce clean errors.
- **`backend/Makefile`** — only passes `-extldflags "-static"` on non-macOS platforms. On macOS the system linker (`clang`) does not support static linking and the build fails with `ld: library 'crt0.o' not found`. Verified: `make -C backend build` now produces `backend/bin/nebraska` (Mach-O) on macOS; Linux container builds keep the flag.
- **`docs/development-macos.md`** — new macOS development guide, linked from the README's Development section.
- **`CHANGELOG.md`** — changelog entry.

## Verification

- All three scripts pass `bash -n` and run correctly under `/bin/bash` 3.2.
- `make -C backend build` succeeds on macOS (previously failed at link time).
- `NEBRASKA_SKIP_TESTS=1 go test -p 1 ./...` passes.
- Full test suite (DB-backed) requires Docker, so it runs in CI.


KartikSuryavanshi:
Thanks for the review. Both suggestions are addressed in the amended commit:

- **run-all.sh / setup_local_db.sh:** added a `[[ "${#}" -ge 2 ]] || fail "Missing value for ${1}"` guard before reading `${2}`, so `./run-all.sh --port` (or any option without a value in either script) now exits with a controlled error message instead of an unbound-variable error under `set -u`. Also switched the leftover-args message to `${*}` so both scripts are shellcheck-clean.
- **docs/development-macos.md:** the DB-free test command is now `NEBRASKA_SKIP_TESTS=1 make -C backend check-code-coverage`, since that target runs `go test -p 1 ./...` and the DB-backed packages' `TestMain` only skip when `NEBRASKA_SKIP_TESTS` is set.

Verified: both scripts pass `bash -n`, run cleanly under `/bin/bash` 3.2, and produce controlled "Missing value for ..." errors for missing-argument invocations. DCO is green.

KartikSuryavanshi:
@ervcz Can you please review this PR?

ervcz:

Thanks for your patience.

backend/Makefile
```diff
@@ -29,7 +29,15 @@ ifeq ($(VERSION),)
...
 
-LDFLAGS := "-w -X github.com/flatcar/nebraska/backend/pkg/version.Version=$(VERSION) -extldflags \"-static\""
+# The external linker on macOS (clang) does not support static linking
+# (-static), so that flag is only used on other platforms (e.g. Linux,
+# where the container image is built).
+VERSION_LDFLAGS := -w -X github.com/flatcar/nebraska/backend/pkg/version.Version=$(VERSION)
+ifeq ($(shell uname -s),Darwin)
+	LDFLAGS := "$(VERSION_LDFLAGS)"
+else
+	LDFLAGS := "$(VERSION_LDFLAGS) -extldflags \"-static\""
```
chengr4, line 39: I am not quite sure here, but if the backend does not depend on any C libraries,
maybe we should remove `-extldflags \"-static\"` and add `CGO_ENABLED=0` somewhere in Makefile 🤔?

## References

https://github.com/flatcar/nebraska/blob/35f11f46aedc6722bf747cafea63f2954c3ef6ec/Dockerfile#L8

backend/tools/check_pkg_test.sh
```diff
@@ -8,7 +8,10 @@
 set -euo pipefail
 
 output_lines=()
-mapfile -t output_lines < <(go list -f '{{.Dir}} - {{.TestGoFiles}} {{.XTestGoFiles}}' ./...)
+# mapfile is not available on macOS (bash 3.2), so collect the lines in a loop.
```
chengr4, line 11: Since `mapfile` is removed, mentioning it in the comment might confuse future readers 🤔?

backend/Makefile
ervcz: Good catch. I haven't tested any of this myself, but reading around it I wonder if it goes further than the PR assumes.
                                                                                                                                                                                                                                                                                                                                                                                      
`go tool link -help` describes `-extldflags` as "pass flags to external linker". If Go doesn't invoke an external linker when `CGO_ENABLED=0`, would the flag be doing anything for us at all? You already pointed at `Dockerfile#L8`, which sets `CGO_ENABLED=0`, so the container may be getting its static binary from there rather than from this flag.
                                                                                                                                                                                                                                                                                                                                                                                      
Could you try building both ways and see what actually comes out? If `CGO_ENABLED=0` alone gives a static binary, would something like this be simpler than the platform conditional?
                                                                                                                                                                                                                                                                                                                                                                                      
```make
bin/nebraska: run-generators
	CGO_ENABLED=0 go build -a -tags netgo -trimpath -ldflags ${LDFLAGS} -o bin/nebraska ./cmd/nebraska
```
                                                                                                                                                                                                                                                                                                                                                                                      
I'm also unclear why cgo is involved at all given `-tags netgo`. `go doc net` says netgo "disables entirely the use of the native (CGO) resolver", so it sounds like it only covers DNS, and the `os/user` docs mention an `osusergo` tag for the same purpose. Could `os/user` be what pulls cgo in? Comparing `go list -deps` with and without `osusergo` would presumably show it.
                                                                                                                                                                                                                                                                                                                                                                                      
One more, and I may have this wrong: doesn't `uname -s` report the host rather than the build target? If so, would `GOOS=linux make build` on a Mac still take the Darwin branch?

docs/development-macos.md
ervcz, line 1: My hesitation is that once the Makefile question is settled, I'm not sure how much is left that's genuinely macOS specific and worth adding a dedicate doc for Mac

chengr4:
Hi @KartikSuryavanshi,

Thank you for the PR. Since I am the assignee of this issue, I am helping review this PR.

Please reply to this PR within 3 days regarding whether you would still like to work on this; otherwise, we will close this PR.

If you have any question, feel free to ask.

Thank you so much again!

-- 

cc @ervcz

PR #1489, opened 2026-08-04
fix(api): correct regex character class in product ID validation

prayag0one4:
Fixes a regex character class typo in `validateProductID` where subsegments specified `[a-zA-Z0-1\-]` instead of `[a-zA-Z0-9\-]`. 

While this typo was inadvertently masked by the greedy repetition of the outer group and did not functionally reject product IDs in practice (as the trailing `[a-zA-Z0-9]` in the regex matched digits 2-9), the pattern is now corrected to strictly match the documented alphanumeric intent and prevent future confusion.

Additionally, this PR optimizes performance by pre-compiling `productIDRegex` at the package level using `regexp.MustCompile`, which avoids recompiling the expression on every single API validation call.

### DIFF
--- a/backend/pkg/api/admin/applications.go
+++ b/backend/pkg/api/admin/applications.go
@@ -83,6 +83,8 @@ func (s *Service) AddAppCloning(app *types.Application, sourceAppID string) (*ty
 	return app, nil
 }
 
+var productIDRegex = regexp.MustCompile(`^[a-zA-Z]+([a-zA-Z0-9\-]*[a-zA-Z0-9])*(\.[a-zA-Z]+([a-zA-Z0-9\-]*[a-zA-Z0-9])*)+$`)
+
 func validateProductID(productID null.String) error {
 	if productID.Ptr() == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


ervcz:
Hello @prayag0one4, how did you find the problem and have you ever tested/checked your claim about the issue? I cannot reproduce it.

prayag0one4:
> Hello @prayag0one4, how did you find the problem and have you ever tested/checked your claim about the issue? I cannot reproduce it.

Hey @ervcz, thanks for looking into this! You're totally right, and I owe you an apology for the confusing PR description. I spotted the 0-1 typo while reading through the code and jumped the gun, assuming it was rejecting valid IDs without actually running a test to verify it. My bad!

I just dug into why it wasn't failing for you, and it turns out the regex engine is masking the typo. Because of the * repetition and the trailing [a-zA-Z0-9], it parses the string one character at a time. The trailing part just swallows up any digits 2-9, so it never actually fails.

Even though it's not breaking functionality, do you think it's still worth merging? The PR fixes the 0-1 typo so it matches the alphanumeric comment and doesn't confuse the next person reading the code. It also pre-compiles the regex at the package level (regexp.MustCompile) so we aren't rebuilding the regex on every single API validation call.

Let me know what you think! Happy to update the PR description so it's accurate if you're open to re-opening it.

ervcz:
Hello @prayag0one4, the typo seems to be real, can you please correct your commit message although?
"causing valid product IDs containing digits 2-9 to be erroneously rejected." <-- because after my quick test, this seems to be a false claim.

prayag0one4:
Hello @ervcz , i have updated the PR description to removed that claim.

PR #1499, opened 2026-08-05
Fix duplicate-key violations in HA deployments by coordinating syncer leadership across replicas

### LINKED ISSUE #388: Issus when running more than a replica for HA
## Description

Errors running randomly when deployment has more than 1 replica

## Impact

Operations are not performed (example: create group)

1. **Set-up**: 
Deployed via helm chart with replicaCount: 2

## Expected behavior

Operation are correctly applied

## Additional information

```bash
runtime error: invalid memory address or nil pointer dereference
runtime/panic.go:199 (0x4451eb)
runtime/signal_unix.go:394 (0x445028)
github.com/kinvolk/nebraska@/cmd/nebraska/controller.go:339 (0xda8809)
github.com/gin-gonic/gin@v1.6.3/context.go:161 (0x999fba)
github.com/kinvolk/nebraska@/cmd/nebraska/controller.go:167 (0xda66ee)
github.com/gin-gonic/gin@v1.6.3/context.go:161 (0x999fba)
github.com/Depado/ginprom@v1.7.0/prom.go:323 (0xda4f7a)
github.com/gin-gonic/gin@v1.6.3/context.go:161 (0x999fba)
github.com/gin-gonic/gin@v1.6.3/recovery.go:83 (0x9ac873)
github.com/gin-gonic/gin@v1.6.3/context.go:161 (0x999fba)
github.com/gin-gonic/gin@v1.6.3/gin.go:409 (0x9a3efc)
github.com/gin-gonic/gin@v1.6.3/gin.go:367 (0x9a35fd)
net/http/server.go:2831 (0x72dc53)
net/http/server.go:1919 (0x7294f4)
runtime/asm_amd64.s:1357 (0x45d360)


<nil> ERR addGroup - adding group &{e4657731-2362-404e-a45c-ad14e9ee990c flatcar-k8s.test flatcar-k8s.test 0001-01-01 00:00:00 +0000 UTC false e96281a6-d1af-4bde-9a0a-97b76e56dc57 {{e06064ad-4414-4904-9a6e-fd465593d1b2 true}} true false false {{Europe/Madrid true}} 1 hours 999999 60 days <nil> flatcar-k8s.test} error="ERROR: duplicate key value violates unique constraint \"groups_application_id_name_key\" (SQLSTATE 23505)" context=nebraska
```

sid200727:
Fixes #388: every replica previously ran its own independent syncer with no coordination between them. `cmd/nebraska/main.go` starts `syncer.Start()` unconditionally on every process, and the ticker loop calls `checkForUpdates()` on the same interval on each replica independently. In HA deployments (e.g. via the Helm chart's `replicaCount > 1`), this caused every replica to write to the same shared tables concurrently, producing the duplicate-key violations and nil-pointer panics reported in the issue.

Adds a Postgres advisory-lock-based leadership check so only one replica performs the sync per tick; others skip it:

-> `api.AcquireSyncLeadership` - session-scoped `pg_try_advisory_lock` on a dedicated connection. If the leader crashes, Postgres detects the dead connection and releases the lock automatically, so the next tick's winner picks it up cleanly, no manual lease/TTL needed
-> `Syncer.Start()` now calls `runSyncTick()`, which acquires leadership before `checkForUpdates()` and releases it after

There is only one syncer per Nebraska deployment (not per-tenant), so a single fixed lock id is sufficient here.

## How to use

Deploy with `replicaCount > 1` (or run two local instances against the same DB) and confirm only one replica's logs show sync activity per tick; the other should log "sync leadership not acquired, skipping tick".

## Testing done

`TestAcquireSyncLeadership_ConcurrentReplicas` verifies real `pg_try_advisory_lock` contention between two API instances sharing a database - exactly one acquires leadership, and a third attempt succeeds again after release. Passes consistently across repeated runs.
make check-backend-with-container

Full suite green, including:

ok github.com/flatcar/nebraska/backend/pkg/api 13.209s
ok github.com/flatcar/nebraska/backend/pkg/syncer 3.252s

`golangci-lint run` - 0 issues.

### DIFF
--- a/backend/pkg/api/sync_leadership.go
+++ b/backend/pkg/api/sync_leadership.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2025 The Nebraska Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"context"
+	"database/sql"
+)
+
+// syncLeadershipLockID is a fixed, arbitrary Postgres advisory lock id used
+// to coordinate the syncer's periodic sync tick across replicas. There is
+// only ever one syncer per Nebraska deployment (it is not per-tenant), so a
+// single fixed id is sufficient - unlike per-TMS locks, no derivation from
+// configuration is needed here.
+const syncLeadershipLockID = 5548415426157
+
+// SyncLeadership represents held leadership for one sync tick, backed by a
+// session-scoped Postgres advisory lock on a dedicated connection. Close
+// releases the lock and the connection. If the process holding it crashes,
+// Postgres detects the dead connection and releases the lock automatically.
+type SyncLeadership struct {
+	conn *sql.Conn
+}
+
+// Close releases the advisory lock and the underlying connection.
+func (l *SyncLeadership) Close() error {
+	defer func() { _ = l.conn.Close() }()
+
+	_, err := l.conn.ExecContext(context.Background(), "SELECT pg_advisory_unlock($1)", syncLeadershipLockID)
+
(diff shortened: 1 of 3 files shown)


ervcz:

Thanks for your patience.

sid200727:
No worries at all, I completely understand. Thank you for the heads-up! Happy to rebase or adjust anything whenever it gets picked up. I'll also take a look at the changelog entry checkbox in the meantime so it's ready for review.

PR #1500, opened 2026-08-05
omaha: reject registration when machine ID is empty

### LINKED ISSUE #550: No ID for insert
## Description

There's a number of postgresql error log items of an insert with no ID.

```
20NN-0N-0N NN:NN:NN UTC:x.x.x.x(47910):nebraska-xx@nebraska-xx:[3653]:STATEMENT: INSERT INTO "instance" ("id", "ip", "alias") VALUES ('', 'y.y.y.y', '') ON CONFLICT (id) DO UPDATE SET "alias"='',"id"='',"ip"='y.y.y.y'
```

## Impact

Not sure. There shouldn't really be inserts happening with no ID though.

See postgresql error logs.

## Expected behavior

ID should not be empty.

Mounika-max-348:
Postgres error logs showed repeated failed inserts such as:

    INSERT INTO "instance" ("id", "ip", "alias") VALUES ('', 'y.y.y.y', '') ...

This happens because `reqApp.MachineID` from an incoming Omaha request is
used directly to build the `Instance` struct with no validation. If a
client sends a request with an empty or missing machine ID, that empty
string flows straight into `RegisterInstance`, which then tries to insert
it into the `instance` table. The `id` column already has a
`check (id <> '')` constraint, so Postgres correctly rejects the insert —
but the failure was only logged at Debug level and silently swallowed, so
it kept recurring on every request from an affected client with no one
noticing.

This was especially sneaky on ping-only requests (no `updatecheck`): the
error from `RegisterInstance` was discarded and the response still
reported an `ok` ping status back to the client, so there was no feedback
loop telling anyone something was wrong.

This PR adds a check in `buildOmahaResponse` (`backend/pkg/omaha/omaha.go`)
that rejects the app entry early when `reqApp.MachineID` is empty, before
any instance registration or event processing happens. The response now
correctly reports `error-instanceRegistrationFailed` instead of silently
succeeding (or silently failing, in the ping case).

Fixes #550

This is my first contribution here, so I'd really appreciate any feedback
on the approach — happy to adjust if there's a preferred pattern for this
kind of validation in the codebase!

## How to use

Reviewers can validate by sending an Omaha request with an `<app>` element
that has no `machineid` attribute against a local nebraska instance, and
confirming the response now reports `error-instanceRegistrationFailed`
instead of a silent `ok`. The relevant code path is in
`backend/pkg/omaha/omaha.go`, `buildOmahaResponse`.

## Testing done

### DIFF
--- a/backend/pkg/omaha/omaha.go
+++ b/backend/pkg/omaha/omaha.go
@@ -139,6 +139,12 @@ func (h *Handler) buildOmahaResponse(omahaReq *omahaSpec.Request, ip string) (*o
 			respApp.AddUpdateCheck(omahaSpec.UpdateInternalError)
 			return omahaResp, nil
 		}
+		if reqApp.MachineID == "" {
+			l.Warn().Str("app", reqApp.ID).Msgf("buildOmahaResponse - request has no machine ID, refusing to register instance")
+			respApp.Status = h.getStatusMessage(api.ErrRegisterInstanceFailed)
+			respApp.AddUpdateCheck(omahaSpec.UpdateInternalError)
+			continue
+		}
 
 		for _, event := range reqApp.Events {
 			if err := h.processEvent(reqApp.MachineID, appID, group, event); err != nil {


Mounika-max-348:
Hi @ervcz  and @tormath1 , I opened a PR for issue #550 — Omaha requests with an empty machine ID were causing failed inserts and spamming the Postgres logs, sometimes silently on ping-only requests. Fixed by rejecting empty machine IDs early in buildOmahaResponse. Would appreciate a review when you get a chance — happy to adjust anything. Thanks!

ervcz:

Thanks for your patience.

Mounika-max-348:
Thank you for your words!
I will wait until the PR get reviewed properly..

PR #1518, opened 2026-08-05
helm: enable PostgreSQL persistence by default

ojasshelke46:
## Why

The chart shipped `postgresql.primary.persistence.enabled: false`, so the
documented install path put the Nebraska database in the container
filesystem. Nothing warns the operator at install time — the deployment
comes up looking completely healthy. The next restart of the PostgreSQL
pod (node reboot, cluster upgrade, rescheduling, OOM kill) discards every
registered application, update group and its rollout policy, every
channel, every uploaded package definition, and the complete update
history of every machine in the fleet. There is nothing to restore from
because nothing was ever written to disk.

## What changed

- `postgresql.primary.persistence.enabled` now defaults to `true`
  (`charts/nebraska/values.yaml`), so an operator opts into an ephemeral
  database instead of opting out of a durable one.
- `NOTES.txt` prints a warning on every `helm install`/`helm upgrade` when
  persistence is disabled, so the risk stays visible.
- README values table updated to match.
- Chart version bumped 2.0.0 → 2.1.0, CHANGELOG entry added.

## Note for reviewers

Clusters without a default StorageClass will now have the bundled
PostgreSQL PVC stay `Pending` on a fresh install unless
`postgresql.primary.persistence.storageClass` is set. That's the intended
trade-off (loud failure over silent data loss) but flagging it since it's
a behavior change on install.

## Test plan

- [x] `helm lint charts/nebraska`
- [x] `helm template` with default values → PVC/`volumeClaimTemplates` present
- [x] `helm template --set postgresql.primary.persistence.enabled=false` → no PVC
- [x] `helm install --dry-run` with persistence off → warning renders in NOTES
- [x] `helm install --dry-run` with default values → no warning
- [x] `helm install --dry-run --set postgresql.enabled=false` → no warning (guarded)


ojasshelke46:
@ervcz can you review this

ervcz:

Thanks for your patience.

PR #1519, opened 2026-08-05
sessions: mark session cookie Secure and SameSite=Lax

ojasshelke46:
## Why

Session.Save built the session cookie with HttpOnly set but never Secure
or SameSite.

- Without Secure, the cookie is sent over plain HTTP too. Anyone
  observing traffic on shared Wi-Fi, or able to get a browser to load
  any plain `http://` link to the Nebraska host, can capture the cookie
  in readable form and use it to impersonate that user — including an
  admin.
- Without SameSite, another site the user visits can cause their
  browser to send authenticated requests to Nebraska. Modern browsers
  apply a Lax default on their own, but relying on that leaves older
  and differently configured browsers exposed.

## What changed

- `Session.Save` (`backend/pkg/sessions/session.go`) now sets
  `Secure` and `SameSite: http.SameSiteLaxMode` on the cookie.
- `Secure` follows a new `--secure-cookie` flag
  (`config.SecureCookie`, defaults to `true`), since Nebraska can
  legitimately run behind a plain-HTTP reverse proxy in some setups.
  Threaded through `sessions.NewStore` → session builder → `Session`.
- `pkg/server/server.go`'s `setupSessionStore` passes
  `conf.SecureCookie` through — the only real caller, since only the
  `github` auth mode uses cookie sessions today.
- Test harness (`pkg/sessions/harness.go`) updated to pass
  `secure: true` so its own coverage is unchanged.
- Added `TestSessionSaveCookieAttributes`, which exercises
  `Session.Save` with the store configured both ways and asserts
  `HttpOnly`, `Secure`, and `SameSite` on the resulting cookie.

## Note for reviewers

`harness.go` also builds a raw `http.Cookie` in `newRequestWithCookie`,
used to simulate an *incoming* request cookie. I left that alone —
it's only ever passed to `http.Request.AddCookie`, which only reads
`Name`/`Value`, so `Secure`/`SameSite`/`HttpOnly` have no effect on it.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run ./pkg/sessions/... ./pkg/config/... ./pkg/server/...` → 0 issues
- [x] `go test ./pkg/sessions/...` — all pass, including new cookie-attribute test

### DIFF
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	Debug               bool   `koanf:"debug"`
 	ServerPort          uint   `koanf:"port"`
 	RollbackDBTo        string `koanf:"rollback-db-to"`
+	SecureCookie        bool   `koanf:"secure-cookie"`
 
 	GhClientID        string `koanf:"gh-client-id"`
 	GhClientSecret    string `koanf:"gh-client-secret"`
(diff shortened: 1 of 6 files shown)


ojasshelke46:
@ervcz can you review this

ervcz:

Thanks for your patience.

PR #1520, opened 2026-08-05
fix(db): remove event_type lookup table and flatten event schema

### LINKED ISSUE #454: Don't keep a list of allowed event types and results in the DB
## Description

Currently Nebraska keeps a list of allowed event types and results in the DB. This is coming from the original Coreroller project and it's likely due to restricting what applications do.

## Impact

We cannot send events like "download started.error" because only the update complete is set with an error result in that event table.

So we should remove the events table and migrate the instances status that point to hardcoded values if needed.

yogita-mehta:
Resolves #454
- Creates migration 0024 to drop event_type table
- Adds event_type and event_result columns directly to event table
- Updates Go models and SQL queries to bypass lookup table
- Fixes test seed data in sample_data.sql

# Title: Refactor event schema to remove event_type lookup table

This pull request resolves issue #454 by flattening the `event` table schema. Previously, the database relied on an `event_type` lookup table, which imposed strict constraints and made it difficult to introduce new Omaha protocol event types or gracefully handle undefined ones. 

By dropping the `event_type_id` foreign key and replacing it with `event_type` and `event_result` as direct integer columns on the `event` table, we greatly simplify the schema and increase backend flexibility. The Go models, `goqu` SQL insert logic, metric queries, and the test suite's `sample_data.sql` have all been updated to align with this new structure.

## How to use

Reviewers can validate this PR by running `make check-backend-with-container` to spin up a local test instance. The database should initialize with the new migration (`0024_remove_event_type_restrictions.sql`), successfully seed the data, and pass all API tests without hitting foreign key constraint errors.

## Testing done

I ran the full backend test suite locally against the updated `sample_data.sql` seed to ensure all models, database reads, and endpoints function correctly without the lookup table.

Command:
`make check-backend-with-container`

Output:
```text
ok  [github.com/flatcar/nebraska/backend/pkg/api](https://github.com/flatcar/nebraska/backend/pkg/api) (cached)
ok  [github.com/flatcar/nebraska/backend/test/api](https://github.com/flatcar/nebraska/backend/test/api) 13.205s
ok  [github.com/flatcar/nebraska/backend/test/auth/oidc](https://github.com/flatcar/nebraska/backend/test/auth/oidc) 22.550s


ervcz:

Thanks for your patience.

yogita-mehta:
Hi @ervcz, no worries at all! I completely understand how busy the LFX influx has made the queue. I just went ahead and resolved the merge conflicts that popped up with `main `and fixed my DCO sign-off. Looking forward to the review whenever the team has bandwidth! I have also officially submitted my LFX application for this project.

PR #1546, opened 2026-08-09
fix(omaha): continue processing remaining apps on unknown app ID or group

prayag0one4:
## Summary

When an Omaha request contains multiple `<app>` elements and the first app has an unknown ID or invalid track/group, `buildOmahaResponse` was returning early with only the failed app's response. This silently dropped all subsequent apps from the response.

## Testing

Added regression test `TestMultiAppRequestWithInvalidApp` that sends a multi-app request with:
1. First app: invalid UUID → expects `AppUnknownID` + `UpdateInternalError`
2. Second app: valid app ID → expects `AppOK` + `NoUpdate`

Both apps now appear in the response with correct status codes.

### DIFF
--- a/backend/pkg/omaha/omaha.go
+++ b/backend/pkg/omaha/omaha.go
@@ -119,7 +119,7 @@ func (h *Handler) buildOmahaResponse(omahaReq *omahaSpec.Request, ip string) (*o
 			respApp.Status = h.getStatusMessage(err)
 			respApp.AddUpdateCheck(omahaSpec.UpdateInternalError)
 
-			return omahaResp, nil
+			continue
 		}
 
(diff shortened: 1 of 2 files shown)


ervcz:
Sorry for the slow reply. We've had a big influx of PRs over the last two weeks and I'm behind on reviews. Yours is in the queue and I'll get to it — thanks for your patience, and for the contribution.

prayag0one4:
> Sorry for the slow reply. We've had a big influx of PRs over the last two weeks and I'm behind on reviews. Yours is in the queue and I'll get to it — thanks for your patience, and for the contribution.

No worries at all! Thanks for update and for taking the time to review it.

PR #1555, opened 2026-08-10
fix(dbreads): use timezone-aware now() for stats windows

### LINKED ISSUE #1541: Instance stats and rollout policy queries break when the Postgres session timezone is not UTC
## Description

A bunch of queries in `backend/pkg/api/internal/dbreads` compare `timestamptz` columns against `now() at time zone 'utc'`. That expression does not mean "now, in UTC". It returns a plain `timestamp` with no zone attached. When Postgres compares it back against a `timestamptz` column, it has to convert, and it converts using the **session** timezone, not UTC.

So the whole comparison gets shifted by the session's UTC offset. If the database session runs in UTC everything works. If it runs in anything else, every one of these time windows is off by the offset.

The columns involved are `last_check_for_updates` and `last_update_granted_ts`, both declared `timestamptz` in `0001_initial.sql`.

Worth noting that the timeline queries in the same package (`GetGroupVersionCountTimeline`, `GetGroupStatusCountTimeline`) already use plain `now()`, which is correct. So right now two reporting queries reading the same column disagree about where the window starts.

## How to reproduce

Plain psql, no Nebraska needed:

```sql
CREATE TABLE ia(instance_id text, last_check_for_updates timestamptz);
INSERT INTO ia VALUES
  ('recent-2h',  now() - interval '2 hours'),
  ('recent-10h', now() - interval '10 hours'),
  ('old-22h',    now() - interval '22 hours'),
  ('old-27h',    now() - interval '27 hours'),
  ('old-30h',    now() - interval '30 hours');
```

Three instances checked in within the last 24 hours, so the answer should always be 3.

```sql
SET TIME ZONE 'UTC';
SELECT count(*) FROM ia WHERE last_check_for_updates > now() at time zone 'utc' - interval '1 days';
-- 3

SET TIME ZONE 'America/New_York';
SELECT count(*) FROM ia WHERE last_check_for_updates > now() at time zone 'utc' - interval '1 days';
-- 2   <- the 22h old instance is gone, window shrank to 20h

SET TIME ZONE 'Asia/Kolkata';
SELECT count(*) FROM ia WHERE last_check_for_updates > now() at time zone 'utc' - interval '1 days';
-- 4   <- the 27h old instance is counted, window grew to 29.5h
```

With plain `now()` it returns 3 in all three sessions.

The interval arithmetic is off the same way, which is the part that worries me more:

```sql
SET TIME ZONE 'America/New_York';
SELECT (now() at time zone 'utc') - (now() - interval '1 hour') AS buggy_age,
        now()                     - (now() - interval '1 hour') AS correct_age;
--  buggy_age | correct_age
--  05:00:00  | 01:00:00
```

An update granted one hour ago looks five hours old.

Tested on `postgres:16.3-alpine`.

## What this affects

Dashboard numbers:

- `appInstancesCountQuery` (`applications.go:221`) - the instance count shown for every app and group
- `instanceAppQuery` (`instances.go:258`)
- `getFilterInstancesQuery` (`instances.go:283`)
- `GetGroupVersionBreakdown` (`groups.go:314`, `groups.go:316`)
- `GetGroupInstancesStats` (`groups.go:376`)

Rollout behaviour, which is the bigger problem. `GetGroupUpdatesStats` (`groups.go:250-253`) builds `updates_granted_in_last_period`, `updates_in_progress` and `updates_timed_out`, and those feed `enforceRolloutPolicy` in `pkg/api/updates.go:307` and the safe mode handling in `pkg/api/events.go:176` and `:217`.

If the session timezone is west of UTC, granted updates look older than they are. They drop out of `updates_in_progress` and land in `updates_timed_out` early. With `PolicySafeMode` on, that path calls `disableUpdates(group.ID)` and turns off updates for the entire group.

If it is east of UTC, they look younger. They stay counted in `updates_in_progress` well past `policy_update_timeout`, so the max concurrent updates limit stays full and new instances get parked in `InstanceStatusOnHold`.

The `policy_period_interval` rate limit is shifted by the same offset either way.

## Why the session timezone can be non-UTC

Nebraska never pins it. The DSN in `pkg/api/api.go` has no `TimeZone` or `options` parameter, and nothing runs `SET TIME ZONE` after connecting. The session just inherits the server's `timezone` setting. The official Postgres image defaults to UTC, but a managed instance or a distro package often does not, and Nebraska supports external databases (#673).

`charts/nebraska/values.yaml` sets `TZ: UTC` in `extraEnvVars`, but that is the Nebraska pod's own environment. It does not change the Postgres session timezone.

## Suggested fix

Drop the `at time zone 'utc'` and use plain `now()` at all of the sites listed above. `now()` is already timezone aware, so the comparisons stay correct no matter what the session is set to. It also makes these queries consistent with the timeline queries that already do this.

On top of that it might be worth adding `options=-c timezone=UTC` to the default DSN, or running `SET TIME ZONE 'UTC'` once per connection, so anything added later is safe by default.

I'd be happy to open a PR for this, along with a test that runs the stats queries under a non-UTC session timezone. Wanted to check first whether you'd prefer just the `now()` change or the `now()` change plus pinning the connection timezone.

Harsh63870:
## Summary
Fixes #1541.

Several reporting and rollout-policy queries compared `timestamptz` columns (`last_check_for_updates`, `last_update_granted_ts`) to `now() at time zone 'utc'`. That expression returns a **timestamp without time zone**. When Postgres compares it back to `timestamptz`, it converts using the **session** timezone, so the window shifts by the UTC offset whenever the DB session is not UTC.

That skews:
- Dashboard counts (`GetGroupInstancesStats`, `GetGroupVersionBreakdown`, app instance counts, instance list filters)
- Rollout enforcement (`GetGroupUpdatesStats` → `updates_in_progress` / `updates_timed_out` / period rate limits), which can pause updates early (safe mode) or hold instances too long

Timeline queries in the same package already used plain `now()`, so two reporting paths reading the same columns disagreed.

## Changes
1. **Query fix:** replace `now() at time zone 'utc'` with timezone-aware `now()` in:
   - `backend/pkg/api/internal/dbreads/applications.go`
   - `backend/pkg/api/internal/dbreads/instances.go`
   - `backend/pkg/api/internal/dbreads/groups.go`
2. **Defense in depth:** append `TimeZone=UTC` to the DB URL on connect when the operator has not already set a timezone (`withUTCSessionTimezone` in `api.go`). Explicit `TimeZone` / `timezone` / `options=-c TimeZone=...` are left alone.
3. **Tests:**
   - `TestGroupInstanceStatsStableAcrossSessionTimezones` — same 1-day window returns 3 under UTC, America/New_York, and Asia/Kolkata
   - `TestWithUTCSessionTimezone` — DSN helper behavior
4. **CHANGELOG** Unreleased bugfix entry

## Test plan
- [ ] `cd backend && go test ./pkg/api/ -run 'TestWithUTCSessionTimezone|TestGroupInstanceStatsStableAcrossSessionTimezones'`
- [ ] `cd backend && go test ./pkg/api/ -run 'TestGetGroup|TestRegisterInstance|TestGetInstances'`
- [ ] Optional: on a Postgres with `SET TIME ZONE 'Asia/Kolkata'`, confirm group instance totals match UTC before/after

## Notes
Sample fixture SQL (`sample_data.sql`) still uses `now() at time zone 'utc'` for inserts only; those are not the comparison bug described in #1541 and were left unchanged to keep this PR focused.


Harsh63870:
@ervcz , resolved the merge conflict in backend/pkg/api/api.go against upstream/main, upstream had relocated nowUTC to pkg/api/runtime, so this branch now keeps only withUTCSessionTimezone and updates the timezone regression test to use runtimeSvc / runtime.NewInstanceApplication, no changes to the original fix logic, just adapted to the new location...

PR #1582, opened 2026-08-13
fix(helm): default PostgreSQL to the official maintained image

### LINKED ISSUE #1574: PostgreSQL default image (bitnamilegacy/postgresql) will never receive security patches
## Problem

The chart's default PostgreSQL image is `bitnamilegacy/postgresql` — the
frozen archive Bitnami moved its free catalog into when it retired free
image maintenance. Those images stay downloadable but are never
rebuilt, so every PostgreSQL and OS-package CVE published from now on
goes unpatched in every Nebraska install using chart defaults.

I know this was a deliberate call, not an oversight — #1227 moved here
specifically to stop installs breaking when Bitnami retired the free
catalog, and it was the right call at the time. Raising this as the
intended follow-up, since the archive was always meant to be a stopgap
and the gap grows every month it stays default.

## What I'd propose

Point the default at `docker.io/postgres` instead of a Bitnami-family
image. A few things worth deciding together before any PR, since they
affect every existing deployment:

- **Migration is not in-place for persistence-enabled installs.** The
  official image has no `/bitnami` tree, keeps `postgresql.conf`/
  `pg_hba.conf` outside the data directory (Bitnami doesn't), and its
  own images are built against a different glibc than Bitnami's. Reusing
  an existing PVC as-is doesn't work — it needs a dump and restore.
  That's a real migration burden for anyone running this in production
  today, and worth deciding how much notice/documentation that needs
  before it ships.
- **Version pinning policy** — happy to track whatever convention the
  chart already follows elsewhere, wasn't sure if there's a preference
  beyond "not `latest`."
- **Any preferred alternate/mirror** — official Docker Hub `postgres`
  vs. a different maintained rebuild, in case there's a registry or
  supply-chain preference I don't know about.

## Status

I've already got this working on a branch — image swap, the PGDATA
path change the official image requires (`/var/lib/postgresql/pgdata`,
since `/var/lib/postgresql` and its data subdir ship `1777`, which
Postgres rejects), README migration notes, and verified on a local kind
cluster: fresh installs with and without persistence, data surviving a
pod restart, Nebraska's migrations applying cleanly, and a real Omaha
update check returning a package.

Happy to open it as a PR once there's agreement on direction — didn't
want to land a breaking change to every existing deployment without
checking in first, given the history here.

ojasshelke46:
Fixes #1574

## Summary

The chart pulled PostgreSQL from bitnamilegacy/postgresql, the frozen
archive Bitnami moved its free catalog into when it retired free image
maintenance. Never rebuilt, so every PostgreSQL and OS-package CVE
published from now on stays unpatched in it. #1227 moved here as a
stopgap when the free catalog was retired; this is the follow-up
discussed in #1574.

## What changed

- Default image is now `docker.io/postgres:17.10-trixie`
- Data directory moves to `/var/lib/postgresql/pgdata` — the official
  image has no `/bitnami` tree, and `PGDATA` has to be a subdirectory
  the container user creates itself, since `/var/lib/postgresql` and
  its data subdirectory ship `1777`, which Postgres rejects. Volume
  mount path moves accordingly.
- No template changes needed — the subchart drives the image through
  `POSTGRES_USER`/`POSTGRES_PASSWORD`/`POSTGRES_DB`/`PGDATA`, all
  honored by `docker.io/postgres`.
- README documents the migration path for existing deployments with
  persistence enabled — in-place PVC reuse does not work, since Bitnami
  keeps `postgresql.conf`/`pg_hba.conf` outside the data directory and
  its image was built against a different glibc, so a dump and restore
  is required.
- CHANGELOG entry under Security.

## Test plan

Verified on a local kind cluster (deleted after testing):

- [x] `helm lint` clean
- [x] Fresh install, persistence disabled (default) — pods Ready
- [x] Fresh install, persistence enabled — data survives pod delete/restart
- [x] Migrations apply (18 tables, seeded "Flatcar Container Linux" app)
- [x] `/health` returns 200
- [x] Omaha `/v1/update` returns a real package manifest
- [x] Non-default `postgresql.auth.username` variant works

### DIFF
--- a/charts/nebraska/README.md
+++ b/charts/nebraska/README.md
@@ -123,6 +123,13 @@ procedure above, with two additions:
   the official image, and `postgresql.auth.postgresPassword` is then unused. Set
   the password of the user Nebraska connects with through `postgresql.auth.password`.
 
+If your values file already overrides `postgresql.image.registry`,
+`postgresql.image.repository`, `postgresql.image.tag`,
+`postgresql.postgresqlDataDir`, or `postgresql.primary.persistence.mountPath`
+from a previous install, update those overrides to the values above. An old
+override will keep pointing at the bitnami image and paths, and the steps in
+this section will not apply.
+
 ## Parameters
 
 ### Global parameters
--- a/charts/nebraska/README.md
+++ b/charts/nebraska/README.md
@@ -88,8 +88,8 @@ statefulset.apps/nebraska-postgresql scaled
 
 3. Upgrade PostgreSQL version, e.g:
 ```diff
--          image: docker.io/bitnami/postgresql:13.8.0-debian-11-r18
-+          image: docker.io/bitnamilegacy/postgresql:17.5.0
+-          image: docker.io/postgres:17.10-trixie
++          image: docker.io/postgres:18.2-trixie
 ```
 
 5. Apply the changes and scale up Nebraska statefulset to its original value
@@ -100,6 +100,28 @@ $ kubectl exec -ti pod/nebraska-postgresql-0 -- psql < backup.sql
 ```
 
 7. Scale up Nebraska deployment and assert that everything is back to normal
+
+### Migrating from the Bitnami PostgreSQL image
+
+Charts up to 2.0.0 defaulted to `bitnamilegacy/postgresql`, an archive that is no
+longer rebuilt and therefore never receives security fixes. The default is now the
+official `docker.io/postgres` image.
+
+If persistence is disabled (the default) there is nothing to do: the database is
+recreated on upgrade as before.
+
+If persistence is enabled, the existing volume cannot be reused in place, because
+the Bitnami image keeps `postgresql.conf` and `pg_hba.conf` outside the data
+directory and runs `initdb` against a different glibc, which the official image
+reports as a collation version mismatch. Migrate with a dump and restore using the
+procedure above, with two additions:
+
+- The data directory moves from `/bitnami/postgresql/data` to
+  `/var/lib/postgresql/pgdata`. The old directory is left untouched on the volume;
+  remove it once the restore is verified.
+- `postgresql.auth.username` values other than `postgres` become the superuser of
+  the official image, and `postgresql.auth.postgresPassword` is then unused. Set
+  the password of the user Nebraska connects with through `postgresql.auth.password`.
 
 ## Parameters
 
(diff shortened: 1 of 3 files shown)


charts/nebraska/values.yaml
```diff
@@ -207,13 +207,26 @@ postgresql:
     username: postgres
     database: nebraska
     postgresPassword: changeIt
+  # The official PostgreSQL image is used instead of a Bitnami one: the
```
tylerauerbeck: This might just be my opinion here, but I think the bitnami notes are fine in the README, but probably unnecessary in the values file.

charts/nebraska/values.yaml
```diff
@@ -207,13 +207,26 @@ postgresql:
...
+  # PGDATA, which docker.io/postgres honours as well.
+  # Note: `auth.username` values other than "postgres" become the superuser of
+  # the official image, and `auth.postgresPassword` is then unused.
   image:
     registry: docker.io
-    repository: bitnamilegacy/postgresql
-    tag: 17.5.0
+    repository: postgres
+    tag: 17.10-trixie
+  # The official image has no /bitnami tree, so the data directory moves under
```
tylerauerbeck: Same note as above.

charts/nebraska/README.md
```diff
@@ -101,6 +101,28 @@ $ kubectl exec -ti pod/nebraska-postgresql-0 -- psql < backup.sql
 
 7. Scale up Nebraska deployment and assert that everything is back to normal
 
+### Migrating from the Bitnami PostgreSQL image
```
tylerauerbeck, line 104: It might be worth mentioning that users should update their values files with the new appropriate values in case they already have an override for them

charts/nebraska/values.yaml
ojasshelke46: Good call, trimmed it to one line. Full context stays in the README migration section

charts/nebraska/values.yaml
ojasshelke46: Same, trimmed.

charts/nebraska/README.md
ojasshelke46: Added a note covering values files that already override postgresql.image.*, postgresql.postgresqlDataDir, or postgresql.primary.persistence.mountPath - those need to be updated to the new defaults or the migration steps above won't apply.

PR #1619, opened 2026-09-02
fix(auth): validate OIDC access token audience

ervcz:
Forward ports the two 4.0.0 security fixes onto `main`. Both are already released and described in GHSA-8535-v9q5-f7qp, which is public.

OIDC access tokens were verified for signature, issuer and expiry but not audience, so a token the provider issued for any other application in the same realm was accepted. `--oidc-audience` is now required in OIDC mode and checked against the `aud` claim, and tokens that identify themselves as ID tokens are rejected. This is a breaking change for OIDC deployments, see the migration guide. The test realm fixture gains an audience mapper, which the end to end tests need in order to mint tokens carrying `aud: nebraska-api`.

The `groupID` path parameter of the group version timeline endpoint was interpolated into three SQL statements. All three now bind it as a query parameter, and the endpoint declares `groupID` as a UUID in the OpenAPI spec so malformed values are rejected before reaching the database.

The changelog commits move the 4.0.0 entries out of `[Unreleased]` into a released 4.0.0 section, so `main`'s record of 4.0.0 matches the release branch.

Two unrelated tidy-ups ride along. `config.caFile` returns to the chart: it was dropped so chart 3.0.0, which declares appVersion 4.0.0, would not offer a value the 4.0.0 image rejects, and that chart is now published. The `--oidc-roles-path` help said the roles array is in the id token when `Authorize` reads it from the access token.

No UserInfo commit here, `main` already has that feature.


backend/pkg/api/groups_test.go
```diff
@@ -271,3 +273,63 @@ func TestGroupTrackName(t *testing.T) {
...
+// runs, not merely that it returns nothing. The payload writes if the groupID
+// reaches the SQL text, so an unchanged canary is evidence of non execution.
+func TestGetVersionCountTimelineRejectsSQLInjection(t *testing.T) {
+	a := newForTest(t)
+	defer a.Close()
+
+	before, err := a.GetUser("admin")
+	require.NoError(t, err)
+
+	payload := `00000000-0000-0000-0000-000000000000') ish ON true), ` +
```
krnowak, line 323: Is that `ish` a typo?

krnowak (approved):

backend/pkg/api/groups_test.go
ervcz: No, it is not a typo. ish is the alias of the instance_status_history subquery in the real query. The payload uses the same alias

charts/nebraska/templates/deployment.yaml
```diff
@@ -66,6 +66,9 @@ spec:
             {{- if .Values.config.app.httpStaticDir }}
             - "-http-static-dir={{ .Values.config.app.httpStaticDir }}"
             {{- end }}
+            {{- with .Values.config.caFile }}
+            - "-ca-file={{ . }}"
+            {{- end }}
```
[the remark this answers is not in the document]
ervcz: Will be fixed in a follow-up PR.

PR #1621, opened 2026-09-07
fix(api): validate event previous version

### LINKED ISSUE #1553: api: RegisterEvent() stores previousVersion in DB without semver validation
## Problem

`RegisterInstance()` in `backend/pkg/api/instances.go` validates the version string before storing it:

if !dbreads.IsValidSemver(instApp.Version) {
    return nil, ErrInvalidSemver
}

But `RegisterEvent()` in `backend/pkg/api/events.go` stores `previousVersion` directly into the database without calling `IsValidSemver()`:

Vals(goqu.Vals{eventTypeID, instanceID, appID, previousVersion, errorCode})

The only guard in `RegisterEvent()` is:

if previousVersion == "" || previousVersion == "0.0.0.0" {
    // Flatcar specific case
}

This means versions like "v1.2.3", "not-a-version", "1.2" all pass the check and get stored in the DB unvalidated.

## Impact

- Malformed `previousVersion` values are stored in the event table
- When `triggerEventConsequences()` reads `lastUpdateVersion` back and calls `semver.Make()`, the error is silently discarded and the version becomes 0.0.0
- Inconsistent validation — same codebase validates in one place but not another

## Before vs After

Before — no validation on previousVersion:

// RegisterEvent() stores previousVersion with no validation
Vals(goqu.Vals{eventTypeID, instanceID, appID, previousVersion, errorCode})

After — consistent with RegisterInstance():

// Allow empty and Flatcar-specific "0.0.0.0"
if previousVersion != "" && previousVersion != "0.0.0.0" {
    if !dbreads.IsValidSemver(previousVersion) {
        return ErrInvalidSemver
    }
}
Vals(goqu.Vals{eventTypeID, instanceID, appID, previousVersion, errorCode})

## Proposed Fix

Add `IsValidSemver()` validation to `RegisterEvent()` for `previousVersion`, consistent with how `RegisterInstance()` validates versions. Allow empty string and "0.0.0.0" as special Flatcar cases.

pinankshah8-Aa:
Fixes #1553

## Why

`RegisterEvent` could store a malformed `previousVersion` even though instance registration already validates version strings. I reproduced this on current `main` with Nebraska's real API test harness and PostgreSQL: after putting an instance into an update-in-progress state, `RegisterEvent(..., "not-a-version", ...)` returned `nil`.

## What changed

- Validate non-empty, non-`0.0.0.0` `previousVersion` values with the existing `dbreads.IsValidSemver()` helper.
- Return the existing `types.ErrInvalidSemver` before inserting the event when validation fails.
- Preserve the existing Flatcar-specific empty / `0.0.0.0` behavior.
- Add a regression case to `TestRegisterEvent_InvalidParams` that exercises `RegisterEvent` itself.

## Validation

- `go test -p 1 ./pkg/api/... -count=1` against PostgreSQL 17 — passed.
- `go build ./...` — passed.
- `./tools/check_pkg_test.sh` — passed.
- `NEBRASKA_SKIP_TESTS=1 go test ./...` — passed.
- `golangci-lint v2.12.2 run` — 0 issues.
- `git diff --check` — passed.
- `gofmt -d` on both changed Go files — no diff.

## AI assistance

I used ChatGPT to assist with codebase exploration, test planning, and review. I personally reproduced the bug, reviewed and understand every change, and ran the validation above myself.

### DIFF
--- a/backend/pkg/api/events_test.go
+++ b/backend/pkg/api/events_test.go
@@ -40,6 +40,9 @@ func TestRegisterEvent_InvalidParams(t *testing.T) {
 
 	_, _ = rs.GetUpdatePackage(types.Instance{ID: tInstance.ID, IP: "10.0.0.1"}, runtime.NewInstanceApplication(tApp.ID, tGroup.ID, "1.0.0"))
 
+	err = rs.RegisterEvent(tInstance.ID, tApp.ID, tGroup.ID, types.EventUpdateComplete, types.ResultSuccessReboot, "not-a-version", "")
+	assert.Equal(t, types.ErrInvalidSemver, err)
+
 	err = rs.RegisterEvent(tInstance.ID, tApp.ID, tGroup.ID, 1000, types.ResultSuccess, "", "")
 	assert.Equal(t, types.ErrInvalidEventTypeOrResult, err)
 
--- a/backend/pkg/api/runtime/events.go
+++ b/backend/pkg/api/runtime/events.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"github.com/doug-martin/goqu/v9"
 
+	"github.com/flatcar/nebraska/backend/pkg/api/internal/dbreads"
 	"github.com/flatcar/nebraska/backend/pkg/api/types"
 )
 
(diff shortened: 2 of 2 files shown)


pinankshah8-Aa:
Hi! This PR has been ready for review since September 13. The change remains narrowly scoped to validating `previousVersion` in `RegisterEvent` and includes a focused regression test. The branch is mergeable and DCO is passing.

Could a maintainer please take a look when available? I’m happy to address any feedback or run any additional validation you would like.

